PR #7, opened 2019-12-04
Vision

normanrink:
Modify the README.md in two ways:
1) Be more specific about the purpose of YUL IR and the role that it could play in the future.
2) Make a more balanced statement about tools for working with (formal) specifications by listing additional examples of such tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,8 +124,10 @@ We wish to provide a complete language specification with the following parts:
   their relationship, and the context in which they operate.
 
 A secondary goal to providing a specification document could be to also provide
-an executable specification via tools such as the [K
-framework](http://www.kframework.org/index.php/Main_Page).  Providing such an
+an executable specification via tools such as, for example, the
+[K framework](http://www.kframework.org/index.php/Main_Page),
+[PLT Redex](https://redex.racket-lang.org/), or
+[Ott](https://www.cl.cam.ac.uk/~pes20/ott/). Providing such an
 executable specification may become a primary goal if it can be achieved
 without too much effort and without hampering the achievement of other goals.
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,15 @@ We wish to promote the following development practices:
 We wish to leverage the YUL IR format as a means of targeting both the eWASM
 and EVM platforms.  YUL is actively developed by the Solidity community and we
 see a great opportunity for synergy with the Solidity project through shared
-use of YUL.
+use of YUL.  In particular, we advocate that YUL become an independently specified
+and implemented IR through which implementations of high-level smart contract
+languages can be factored.
+
+If, as is our intention, this Rust Vyper implementation is indeed to rely on the
+YUL IR, then the quality of this implementation will crucially depend on a
+high-quality YUL IR.  We therefore strongly suggest that the same development,
+testing, and specification pratices that Rust Vyper is subject to will be applied
+to the YUL IR.
 
 ### Use of Rust as an implementation language
 


README.md
```diff
@@ -84,7 +84,15 @@ We wish to promote the following development practices:
...
-use of YUL.
+use of YUL.  In particular, we advocate that YUL become an independently specified
+and implemented IR through which implementations of high-level smart contract
+languages can be factored.
+
+If, as is our intention, this Rust Vyper implementation is indeed to rely on the
+YUL IR, then the quality of this implementation will crucially depend on a
+high-quality YUL IR.  We therefore strongly suggest that the same development,
+testing, and specification pratices that Rust Vyper is subject to will be applied
+to the YUL IR.
```
davesque: I wonder if some slightly different wording might work better here.  For example, something like this:

"Since Rust Vyper will crucially depend on YUL, we foresee an opportunity to promote YUL as an independently maintained and specified IR suitable for use by other high-level smart contract languages."

I prefer this for a couple of reasons:

1. I haven't spoken much with the Solidity guys about the idea of factoring YUL out of the Solidity project.  I don't want them to be caught off guard if any of them happen to notice this verbiage.  So I'd rather that the wording be a bit more "invitational."
2. I do feel that I haven't encountered a lot of thorough specification material on YUL, but I also don't want to imply that they're not already using adequate development and testing practices.  As far as I know, the Solidity team are a solid group of engineers.  So I'd rather leave the sentence about "strongly suggesting" the same development practices etc. out for now.

axic: It was noticed anyhow :wink:

Can you explain the reason why it would be good to move it outside the "solidity tree"? It is a separate library and namespace, and can be independently used through the "standard json" interface.

With 0.6.0 we're working hard to make the C API much better and as a result with 0.6.0 [solc-rust](https://github.com/axic/solc-rust) will work with complete functionality.

One implementation of Yul which is available in Rust is [yultsur](https://github.com/axic/yultsur), which we plan to update, will support parsing, printing, an AST and some transformation, but has no plans to include a the Yul optimiser or Yul backend (e.g. compilation to targets).

I suggest to consider working together on yultsur and using solc-rust :wave:

davesque: @axic And hah, yeah there you are :).  Honestly, at this point that whole idea is not really a priority.  It just came up in some initial discussions I had with Norman and so I was thinking it might be okay to think of it as a sort of long term goal.  I guess, in theory, having YUL as a separately maintained project could reduce the likelihood that it somehow gets tightly coupled with Solidity.  But let's just not worry about it for now.  I'm sure we'll get a lot of mileage out of using yultsur and solc-rust in the near term.

davesque: @axic I guess you've been keeping an eye on us, huh?

normanrink: I would just like to stress the fact that we are not suggesting to make YUL or a YUL back-end an independent project. The current implementation of YUL as part of the Solidity compiler can stay within the Solidity project.

That said, if the YUL IR were to be elevated to a language/IR in its own right (i.e. with its own and fairly stable specification), this would enable clean and efficient compiler development on top of YUL. (Similar to how LLVM enables many compilers for C-like/systems languages, including Rust.) This can be achieved with the current YUL implementation as part of Solidity provided it exposes a clean (and stable) API that other compiler front-ends can call into to generate and manipulate YUL IR.

axic: > That said, if the YUL IR were to be elevated to a language/IR in its own right (i.e. with its own and fairly stable specification), this would enable clean and efficient compiler development on top of YUL. 

This is the way it is developed. The [documentation](https://solidity.readthedocs.io/en/v0.5.13/yul.html) includes a grammar, a formal specification (!) and some examples. The suite contains a specific Yul directory ([here](https://github.com/ethereum/solidity/tree/develop/test/libyul)) to test the language thoroughly.

There is an external API for Yul using the ["standard json i/o"](https://solidity.readthedocs.io/en/v0.5.13/using-the-compiler.html#compiler-input-and-output-json-description) (hint: use `language: "yul"`, but I admit this is not well advertised).

Please let me know any suggestions how to improve this. I'd actually encourage opening issues on the Solidity repo with some suggestions. Don't be afraid, every suggestion is considered.

axic: Probably the most confusing aspect in the ["Solidity grammar"](https://github.com/ethereum/solidity/blob/develop/docs/grammar.txt) is that we have `InlineAssembly` there, which is a duplication of the "official" Yul grammar as linked in the documentation above.

There's plenty of improvements to make in terms of clarity, but I can assure you that Yul is not treated as a second class citizen behind Solidity, just is still in relatively early stages of development. I'd love to collaborate on Yul.

davesque: Hey I think we're all on the same page here.  And I'm sure there are many things about the YUL project and its tooling/tests/documentation that we're not aware of.  Our lack of knowledge about those things may have led us to believe there's more work to be done there than is actually necessary.  So I say let's not worry too much about this conversation for now and just know that our real intentions currently are to take advantage of YUL and also contribute to it in the process.

axic: I wasn't worried, just had a bit of time on my hands while waiting at an airport :wink:

I really wanted to emphasize though that feel free to reach out, even via public issues on the Solidity repo, as we'd be more than happy to hear how to improve visibility of Yul.

davesque: > I really wanted to emphasize though that feel free to reach out

Awesome!  Glad that you're available!

Issue #10, opened 2020-01-24
Compile guest book contract to Yul


g-r-a-n-t:
It should be possible to compile the following Rust Vyper code into Yul and bytecode:

```
/* guest_book.vy */
type BookMsg = bytes[100]

contract GuestBook:
    pub guest_book: map<address, BookMsg>

    event Signed:
        idx book_msg: BookMsg

    pub def sign(book_msg: BookMsg):
        self.guest_book[msg.sender] = book_msg

        emit Signed(book_msg=book_msg)

    pub def get_msg(addr: address) -> BookMsg:
        return self.guest_book[addr]
```

**TODO list:**
- Function-only ABI builder. See: #15
- Basic Yul compiler. See: #16
- Yul -> bytecode compiler and EVM tests. See: #17 (includes: #15 and #16)
- Support for: addresses, bytes, storage maps, and arrays. See: #18 
- Event definitions and logging. See: #18 also

PR #15, opened 2020-02-14
Generating ABIs from Vyper source code

g-r-a-n-t:
Part of #10

This is a function-only ABI builder with support for u256 values. Other types and events will be added in future PRs.


compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// Builds a contract ABI.
+pub fn contract_def<'a>(
+    type_defs: &'a TypeDefs<'a>,
+    stmt: &'a vyp::ModuleStmt<'a>,
+) -> Result<Contract, CompileError> {
+    if let vyp::ModuleStmt::ContractDef { name: _, body } = stmt {
+        return Ok(Contract {
+            functions: body
+                .iter()
```
ralexstokes, line 101: seems like we can `filter_map` for `vyp::ContractStmt::FuncDef` after `.iter` here and just yield the `stmt.node` -- that way we can avoid the enum check inside `func_def`....

can likely use a similar pattern to avoid the same thing in this function (with `if ContractDef ... else Err`)

compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+                .map(|stmt| func_def(type_defs, &stmt.node))
+                .collect::<Result<Vec<Option<Function>>, CompileError>>()?
+                .into_iter()
+                .filter_map(|function| function)
+                .collect(),
```
ralexstokes, line 106: esp. if we do the above thing, then it seems like we can skip the double iteration and clunky `filter_map(|x| x)` w/ `filter_map` ahead of time; something like:

```rust
body
  .iter()
  .filter(|stmt| is_func_def(stmt))
  .filter_map(|func_def_node| func_def(type_defs, &func_def_node)
  .collect::<_>()?,
```

i can tighten this up in a playground if its unclear

compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+    Err(CompileError::static_str(
+        "Contract definition translation requires ContractDef parameter.",
+    ))
+}
+
+/// Builds an ABI function element, returning None if the function is private.
+pub fn func_def<'a>(
+    type_defs: &'a TypeDefs<'a>,
+    stmt: &'a vyp::ContractStmt<'a>,
+) -> Result<Option<Function>, CompileError> {
```
ralexstokes, line 119: another way to keep the `Option` out of the type signature (see above for how to take the `Result` out...) is to just return the `Function` and let the caller `filter(|f| f.is_public()`

compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/// Maps the type description to an ABI-friendly type and handles custom types.
+pub fn type_desc<'a>(
+    type_defs: &'a TypeDefs<'a>,
+    typ: &'a vyp::TypeDesc<'a>,
+) -> Result<VariableType, CompileError> {
+    if let vyp::TypeDesc::Base { base } = typ {
+        if let Some(custom_type) = type_defs.get(base) {
+            return type_desc(&HashMap::new(), custom_type);
```
ralexstokes, line 175: why pass a new `type_defs` here (via `HashMap::new()`) and not supply the function's `type_defs`... seems like we are losing the capacity for embedded custom types this way...

compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut type_defs = HashMap::new();
+        type_defs.insert("CustomType", &vyp::TypeDesc::Base { base: "u256" });
+
+        let variable_type = type_desc(&type_defs, &vyp_type_desc).unwrap();
+        assert_eq!(variable_type, VariableType::Uint256, "Incorrect type");
+    }
+
+    #[test]
+    fn test_func_def() {
+        let toks = vyper_parser::get_parse_tokens("pub def foo(x: u256) -> u256:\n   return x").unwrap();
```
ralexstokes, line 205: would be nice to test `pub` vs `private` fns (likely in another unit test here...)

can skip if you feel this is covered in a broader test elsewhere....

compiler/src/abi/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+/// source file. See: https://github.com/ethereum/rust-vyper/issues/12
+/// TODO: Improve the JSON ABI builder API.
+pub fn build(src: &str) -> Result<String, CompileError> {
+    let tokens = parser::get_parse_tokens(src).unwrap();
+    let vyp_module = parser::parsers::file_input(&tokens[..]).unwrap().1.node;
+    let contracts = json_builder::module(&vyp_module)?;
+
+    if let Some(contract) = contracts.get(0) {
+        return Ok(serde_json::to_string(&contract.functions)?);
+    }
```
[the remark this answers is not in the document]
g-r-a-n-t: No reason other than just wanting a simple API to create a minimal compiler binary with. I've since removed the binary component, and now think it would be better to just return a map `contract name -> JSON ABI`.

I think it might be good to add an additional function like this:

``` rust
pub fn build_ethpm(src: &str) -> Result<EthPMDesc, CompileError> { ... }
```

in a separate PR addressing the [issue](https://github.com/ethereum/rust-vyper/issues/12) mentioned above.

compiler/src/abi/json_builder.rs
g-r-a-n-t: Ah, this is much cleaner.

compiler/src/abi/json_builder.rs
g-r-a-n-t: Yeah, not sure what the thinking was here.

g-r-a-n-t:
Thanks for the review @ralexstokes! I'll cherry-pick the latest version of this module from #18 and then make changes based on your comments. That way there will be way less noise in the other PRs.

compiler/src/abi/json_builder.rs
g-r-a-n-t: Definitely more readable and efficient.

compiler/src/abi/json_builder.rs
```diff
@@ -0,0 +1,240 @@
...
+        .collect();
+
+    module
+        .body
+        .iter()
+        .filter_map(|stmt| match stmt.node {
+            vyp::ModuleStmt::ContractDef { .. } => Some(contract_def(&type_defs, &stmt.node)),
+            _ => None,
+        })
+        .collect::<Result<Vec<Contract>, CompileError>>()
```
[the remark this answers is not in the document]
g-r-a-n-t: Yeah, it might be better to have that ~~`ModuleDef`~~ ~~`Module`~~ `ModuleABI` struct and then make the public API:

``` rust
#[derive(Serialize, ...)]
pub struct ModuleABI {
  contracts: HashMap<String, ContractABI>,
}

pub fn build(src: &str) -> Result<ModuleABI, CompileError> { ... }

...

let abi = build(src)?;

// It's still easy to create a json list for a given contract.
let contract_json = abi.get("MyContract").unwrap().serialize();
// or serialize the whole module into a json object
let module_json = abi.serialize();

// or just make use of the module pragmatically 
```

compiler/src/abi/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+use vyper_parser as parser;
+
+mod json_builder;
+
+/// Builds a JSON ABI from the source file.
+///
+/// This API should be rethought as it only provides the ABI for the first contract in the
+/// source file. See: https://github.com/ethereum/rust-vyper/issues/12
+/// TODO: Improve the JSON ABI builder API.
+pub fn build(src: &str) -> Result<String, CompileError> {
```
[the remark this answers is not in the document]
njgheorghita: Hope I'm understanding this correctly, but I'm not sure if this is the right place for ethpm. In the solidity compiler - they are adopting the [ethpm v3 spec](http://ethpm.github.io/ethpm-spec/v3-package-spec.html) for their [`"metadata"` output](https://solidity.readthedocs.io/en/v0.6.2/metadata.html) rather than anything to do with the ABI structure, which is what it appears to me to be happening here. But, let me know if I'm misunderstanding or can help with any other questions.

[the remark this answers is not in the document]
pipermerriam: Nick, it might be interesting to start a "EthPM integration guide for compilers" to try and capture some of the ideas and concepts that might be useful for language designers to know but might not be immediately obvious.

The ideas that come to mind are:

- what the process would be for using an EthPM package as the input for compilation
- if the compiler generates an EthPM package, they should make sure they do....
    - what does compiler metadata look like and where should it be injected.
    - make sure you populate (sources, filesystem structure, contract types and their mappings back to their source files, etc?)
- suggested conventions for any areas where they may need to make a decision (such as any areas where unique identifiers need to be generated, etc).

Just a few passing thoughts.  Use them if they sound useful.

compiler/src/abi/builder.rs
```diff
@@ -0,0 +1,281 @@
...
+        .collect::<Result<TypeDefs<'a>, _>>()?;
+
+    Ok(ModuleABIs(
+        module
+            .body
+            .iter()
+            .filter(|s| is_contract_def(&s.node))
+            .map(|s| contract_def(&type_defs, &s.node))
+            .collect::<Result<HashMap<_, _>, _>>()?,
+    ))
```
ralexstokes, line 28: we don't need to do this here but i'll call your attention to `scan`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan

you could collect the type defs into the iterator state which is supplied to every call to `contract_def` and then we can skip the double iteration over the module.... would require some special handling for forward references but i don't know if we are treating those in a special way yet or not

PR #16, opened 2020-02-17
Basic Yul compiler

g-r-a-n-t:
Part of #10.

This includes just enough functionality to compile a contract like this:

```
contract Foo:
    pub def bar(x: u256) -> u256:
        return x
```

To keep it simple, types are basically ignored (everyting is `u256`) and all variables are stored on the stack. There are some `Scope` data structures in here as well, which aren't necessary for it to work, but would be worth reviewing as I intend to build more complexity on top of them.

The Yul code generated by the compiler would not actually be runnable, since it doesn't have the ABI builder from #15 to generate the selector switch from. In #17, I add that to this module and demonstrate that the contract can be deployed and ran in a test.

### DIFF
--- a/compiler/Cargo.toml
+++ b/compiler/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "vyper-compiler"
+version = "0.1.0"
+authors = ["Grant Wuerker <gwuerker@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+vyper-parser = {path = "../parser", version = "0.1.0"}
+hex = "0.4"
+yultsur = { git = "https://github.com/g-r-a-n-t/yultsur" }
+ethabi = "11.0"
--- a/compiler/Cargo.toml
+++ b/compiler/Cargo.toml
@@ -11,3 +11,5 @@ vyper-parser = {path = "../parser", version = "0.1.0"}
 hex = "0.4"
 yultsur = { git = "https://github.com/g-r-a-n-t/yultsur" }
 ethabi = "11.0"
+tiny-keccak = { version = "2.0", features = ["keccak"] }
+stringreader = "0.1"
--- a/compiler/src/yul/ast_builder.rs
+++ b/compiler/src/yul/ast_builder.rs
@@ -174,7 +174,7 @@ pub fn func_def<'a>(
             .collect::<Result<Vec<yul::Identifier>, CompileError>>()?;
 
         let returns = if return_type.is_some() {
-            vec![base::untyped_identifier("return_value")]
+            vec![base::untyped_identifier("return_val")]
         } else {
             Vec::new()
         };
@@ -250,21 +250,22 @@ fn expr<'a>(
 
 #[cfg(test)]
 mod tests {
+    use crate::yul::ast_builder::{contract_def, func_def, type_def, ContractScope, ModuleScope};
+    use std::rc::Rc;
+    use vyper_parser::ast::TypeDesc;
+    use vyper_parser::parsers;
 
-    /*
     #[test]
     fn test_compile_type_def() {
-        let toks = vyper_parser::get_parse_tokens("type Num = u256").unwrap();
+        let src = "type Foo = u256";
+        let toks = vyper_parser::get_parse_tokens(src).unwrap();
         let stmt = parsers::type_def(&toks[..]).unwrap().1.node;
+        let scope = ModuleScope::new();
 
-        let mut scope = ModuleScope::new();
-
-        // Expecting the type definition to exist within new_scope.
-        // There is no resulting statement from compiling a TypeDef.
-        type_def(Rc::clone(&scope), &stmt);
+        type_def(Rc::clone(&scope), &stmt).expect("Unable to handle type def");
 
         assert_eq!(
-            scope.borrow().type_defs["Num"],
+            scope.borrow().type_defs["Foo"],
             &TypeDesc::Base { base: "u256" },
             "Compilation of type definition failed."
         );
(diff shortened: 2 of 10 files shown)


compiler/src/yul/selectors.rs
```diff
@@ -0,0 +1,175 @@
...
+pub fn case(function: &ethabi::Function) -> Result<yul::Case, CompileError> {
+    let selector = Some(selector_literal(function.signature()));
+
+    if function.outputs.is_empty() {
+        return Ok(yul::Case {
+            literal: selector,
+            block: yul::Block {
+                statements: vec![yul::Statement::Expression(yul::Expression::FunctionCall(
+                    yul::FunctionCall {
+                        identifier: base::untyped_identifier("foo"),
```
ralexstokes, line 50: might not need to update here if this is temporary but i expect we can grab some `identifier` from the `ethabi::Function`

compiler/src/yul/selectors.rs
```diff
@@ -0,0 +1,175 @@
...
+    Ok(yul::Case {
+        literal: selector,
+        block: yul::Block {
+            statements: vec![
+                yul::Statement::Expression(yul::Expression::FunctionCall(yul::FunctionCall {
+                    identifier: base::untyped_identifier("mstore"),
+                    arguments: vec![
+                        base::untyped_literal_expr("0"),
+                        yul::Expression::FunctionCall(yul::FunctionCall {
+                            identifier: base::untyped_identifier(&function.name),
```
ralexstokes, line 67: see above about use of hardcoded literal "foo"

compiler/src/yul/base.rs
```diff
@@ -0,0 +1,23 @@
+use yultsur::yul;
```
ralexstokes, line 1: what do you think about calling this module `base_types` instead of just `base`?

compiler/src/yul/constructor.rs
```diff
@@ -0,0 +1,48 @@
...
+                    identifier: base::untyped_identifier("return"),
+                    arguments: vec![
+                        base::untyped_literal_expr("0"),
+                        base::untyped_identifier_expr("size"),
+                    ],
+                })),
+            ],
+        },
+    }
+}
```
ralexstokes, line 39: i'll assume the specifics here are yul conventions for the constructor

compiler/src/yul/ast_builder.rs
```diff
@@ -0,0 +1,307 @@
...
+        .collect::<Result<Vec<Option<yul::Statement>>, CompileError>>()?
+        .into_iter()
+        .filter_map(|statement| {
+            if let Some(yul::Statement::Object(object)) = statement {
+                return Some(object);
+            }
+
+            None
+        })
+        .collect::<Vec<yul::Object>>())
```
ralexstokes, line 70: i would expect this can be simplified like over here: https://github.com/ethereum/rust-vyper/pull/15#discussion_r421920833

compiler/src/yul/ast_builder.rs
```diff
@@ -0,0 +1,307 @@
...
+            type_def(scope, stmt)?;
+            Ok(None)
+        }
+        vyp::ModuleStmt::ContractDef { .. } => {
+            contract_def(scope, stmt).map(|object| Some(yul::Statement::Object(object)))
+        }
+        _ => Err(CompileError::static_str(
+            "Unable to translate module statement.",
+        )),
+    }
```
ralexstokes, line 89: seems like the uses of `Option` here can just be cases of `CompileError` and then it simplifies much of the control flow...

compiler/src/yul/ast_builder.rs
```diff
@@ -0,0 +1,307 @@
...
+    let new_scope = ContractScope::new(scope);
+
+    if let vyp::ModuleStmt::ContractDef { name: _, body } = stmt {
+        let mut statements = body
+            .iter()
+            .map(|stmt| contract_stmt(Rc::clone(&new_scope), &stmt.node))
+            .collect::<Result<Vec<yul::Statement>, CompileError>>()?;
+
+        // TODO: Use functions from actual contract ABI once the builder is merged.
+        statements.push(yul::Statement::Switch(selectors::switch(vec![])?));
```
ralexstokes, line 106: see comment under `selectors` module about this

```suggestion
        statements.push(yul::Statement::Switch(selectors::dispatch_switch(vec![])?));
```

compiler/src/yul/selectors.rs
```diff
@@ -0,0 +1,175 @@
+use crate::errors::CompileError;
```
ralexstokes, line 1: something that came across my mind when reviewing this was the genericity of the function names `switch`, `expression` and `case`. you could make the argument that we rebind the names on import elsewhere and they make sense in the context of this module, but it is something to consider.

i left a comment at a usage of `switch` to point out what I mean

compiler/src/yul/selectors.rs
```diff
@@ -0,0 +1,175 @@
...
+        .collect::<Result<Vec<yul::Case>, CompileError>>()?;
+
+    Ok(yul::Switch {
+        expression: expression(),
+        cases,
+    })
+}
+
+/// Builds an expression that loads the first 4 bytes of the calldata.
+pub fn expression() -> yul::Expression {
```
ralexstokes, line 22: do we need anything `pub` in this module except `switch`?

compiler/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+authors = ["Grant Wuerker <gwuerker@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+vyper-parser = {path = "../parser", version = "0.1.0"}
+hex = "0.4"
+yultsur = { git = "https://github.com/g-r-a-n-t/yultsur" }
```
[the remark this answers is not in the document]
g-r-a-n-t: I needed to add a few [structs](https://github.com/axic/yultsur/pull/8/files#diff-a81ba8e5fef7f52c355be380438e0635) that exist in the Yul spec, but not in `yultsur`.

compiler/src/yul/selectors.rs
g-r-a-n-t: Nope :)

https://github.com/g-r-a-n-t/rust-vyper/blob/types/compiler/src/yul/runtime/abi.rs

compiler/src/yul/selectors.rs
g-r-a-n-t: I agree. There must be applicable CS nomenclature for this.

compiler/src/yul/selectors.rs
g-r-a-n-t: Oh, woops. This got fixed [later on](https://github.com/g-r-a-n-t/rust-vyper/blob/types/compiler/src/yul/runtime/abi.rs#L25), probably after trying to run a function not named "foo".

compiler/src/yul/selectors.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    let mut keccak = Keccak::v256();
+    let mut selector = [0u8; 4];
+
+    if let Some(first_half) = sig_halves.next() {
+        keccak.update(first_half.as_bytes());
+        keccak.finalize(&mut selector);
+    }
+
+    base::untyped_literal(&format!("0x{}", hex::encode(selector)))
```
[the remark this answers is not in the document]
g-r-a-n-t: This is what it looks like now:

`literal! { (format!("0x{}", hex::encode(selector))) }`

See the literal [macro](https://github.com/axic/yultsur/pull/8/files#diff-5d646c29d2485962115a1de5deb88893R20). (I should add better docs to that file, as the notation is somewhat confusing.)

Currently, solc's yul compiler doesn't even handle explicit types. And even if it did, it probably wouldn't be very useful, so I just removed the prefix all together.

compiler/src/yul/ast_builder.rs
g-r-a-n-t: Yeah, I'll remove this pattern wherever I see it.

compiler/src/yul/constructor.rs
g-r-a-n-t: Basically [this](https://solidity.readthedocs.io/en/v0.6.7/yul.html#specification-of-yul-object)

compiler/src/yul/selectors.rs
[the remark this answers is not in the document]
ralexstokes: cool cool

compiler/src/yul/constructor.rs
ralexstokes: thanks for the link -- i did end up there last night and saw where you got it from -- nice work on the constructor basics here!

compiler/src/yul/selectors.rs
ralexstokes: i'd just go with what sounds natural and readable. and ideally unique across as many contexts as possible. of course, we quickly end up optimizing a many-dimensional space and can't satisfy all of our constraints at once :)

off the top of my head, i would call the thing `expression` as `predicate` as the `selector::predicate`, while still jargon-y is clearer to me which part of the expression it is referring to, rather than just `selector::expression`.

and i'd prolly add some more semantic content to the symbol for `case` -- something like `switch_case` or `arm` or something.... this being said if these functions don't leak outside of this module there is less need to do this given the context gained as a local reader of this module

ralexstokes:
thanks for the feedback! looks good to me :)

PR #17, opened 2020-02-19
Bytecode compiler and EVM tests

g-r-a-n-t:
Part of #10.

*It's probably best to review this after #15 and #16 get merged, since those are included.*

This includes an `evm` module which uses the solc Rust bindings to compile Yul into bytecode. It also has a test that deploys a simple contract onto the EVM and interacts with it using the ABI produced by the `abi` module.

### DIFF
--- a/compiler/Cargo.toml
+++ b/compiler/Cargo.toml
@@ -16,3 +16,8 @@ solc = { git = "https://github.com/g-r-a-n-t/solc-rust" }
 ethabi = "11.0"
 tiny-keccak = { version = "2.0", features = ["keccak"] }
 stringreader = "0.1"
+
+[dev-dependencies]
+evm = "0.14"
+primitive-types = "0.6"
+evm-runtime = "0.14"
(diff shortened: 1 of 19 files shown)


compiler/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+vyper-parser = {path = "../parser", version = "0.1.0"}
+serde_json = "1.0"
+serde = "1.0"
+hex = "0.4"
+yultsur = { git = "https://github.com/g-r-a-n-t/yultsur" }
+solc = { git = "https://github.com/g-r-a-n-t/solc-rust" }
```
ralexstokes, line 15: anything specific in this fork?

g-r-a-n-t: I needed to update the version of solc that it binds to to get Yul compilation.

https://github.com/axic/solc-rust/pull/1

ralexstokes: cool, looks good; we will want to switch to the upstream when we can but that can happen in a later PR

PR #18, opened 2020-03-01
Misc features for guestbook

g-r-a-n-t:
As stated in #10, this PR includes support for Vyper -> Bytecode compilation of the following contract:

```vyper
/* guest_book.vy */
type BookMsg = bytes[100]

contract GuestBook:
    pub guest_book: map<address, BookMsg>

    event Signed:
        idx book_msg: BookMsg

    pub def sign(book_msg: BookMsg):
        self.guest_book[msg.sender] = book_msg

        emit Signed(book_msg=book_msg)

    pub def get_msg(addr: address) -> BookMsg:
        return self.guest_book[addr]
```

It also includes support for a number of other contracts with different features. Those contracts can be found [here](https://github.com/g-r-a-n-t/rust-vyper/tree/types/compiler/tests/fixtures).

The contracts mentioned above are automatically tested [here](https://github.com/g-r-a-n-t/rust-vyper/blob/types/compiler/tests/evm_contracts.rs) using [rust-evm](https://github.com/sorpaas/rust-evm) and [ethabi](https://github.com/openethereum/ethabi). This ensures that the JSON ABIs provided for each contract allow for proper encoding and decoding, and that the contracts can be deployed onto the EVM and executed as expected.

This includes: #15, #16, and #17.


g-r-a-n-t:
Hey @pipermerriam, this is feature complete and ready for review.

parser/src/grammar/Python.asdl
```diff
@@ -9,7 +9,6 @@ module Python
     mod = Module(stmt* body)
         | Interactive(stmt* body)
         | Expression(expr body)
-
```
ralexstokes, line 12: why remove the whitespace here? just a formatting change?

g-r-a-n-t: Not sure, it wasn't intentional.

ralexstokes: if it is extra then we can remove it -- if it was there to provide some visual segregation then may be worth keeping in!

g-r-a-n-t: Yeah, I'll undo this change and make adjustments based on your other comments.

Also, I'll add some better documentation to this library.

compiler/src/abi/elements.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+/// An event interface.
+#[derive(Serialize, Debug, PartialEq, Clone)]
+pub struct Event {
+    /// The event's name.
+    pub name: String,
+    // FIXME: This attribute should be added when serialized instead of being defined in the struct.
+    /// The type of an event (Always "event").
```
ralexstokes, line 63: ignore prior comment, you have it here :)

compiler/src/abi/elements.rs
```diff
@@ -0,0 +1,272 @@
...
+#[derive(Serialize, Debug, PartialEq, Clone)]
+pub struct FuncOutput {
+    /// The output's name.
+    pub name: String,
+    /// The output's type.
+    #[serde(rename = "type")]
+    pub typ: VariableType,
+}
+
+// FIXME: This should be removed in favor of separate structs for each function type.
```
ralexstokes, line 119: curious why?

compiler/tests/evm_contracts.rs
```diff
@@ -0,0 +1,321 @@
...
+        );
+
+        harness.caller = sender.clone().to_address().unwrap();
+
+        harness.test_function(&mut executor, "sign", vec![bytes.clone()], None);
+
+        harness.test_function(&mut executor, "get_msg", vec![sender], Some(bytes.clone()));
+
+        // Destruct the executor
+        harness.event_emitted(executor, "Signed", vec![bytes]);
```
ralexstokes, line 300: what exactly does `Destruct` mean here? i'm assuming just bc we are leaving the `with_executor` block but this comment reads like it is attached to the event check

compiler/tests/evm_contracts.rs
```diff
@@ -0,0 +1,321 @@
...
+        let mut harness = deploy_contract(&mut executor, "return_sender.vy", "Foo");
+
+        let sender = address_token("1234000000000000000000000000000000005678");
+
+        harness.caller = sender.clone().to_address().unwrap();
+
+        harness.test_function(
+            &mut executor,
+            "bar",
+            // FIXME: There's an issue parsing functions with 0 params. This is here to mitigate that failure.
```
ralexstokes, line 316: general comment on `FIXME` -- if they are going to remain in the codebase for some indefinite period of time, we should move them to issues before we merge in this PR

compiler/src/yul/runtime/functions.rs
```diff
@@ -0,0 +1,161 @@
+use yultsur::*;
```
ralexstokes, line 1: interesting -- is this runtime not defined elsewhere -- like in `yultsur`? 

it may not be but it would be nice to have a single `yul` runtime available to rust programs that we can just import as a crate

compiler/src/yul/runtime/abi.rs
```diff
@@ -0,0 +1,105 @@
+use crate::errors::CompileError;
```
ralexstokes, line 1: there may not be a nice way to do this but i'll just point out it would be ideal to avoid duplicating the logic of making the selector in the ABI and also here...

compiler/src/yul/mappers/module.rs
```diff
@@ -0,0 +1,55 @@
...
+        .collect::<Result<Vec<Option<yul::Statement>>, CompileError>>()?
+        .into_iter()
+        .filter_map(|statement| {
+            if let Some(yul::Statement::Object(object)) = statement {
+                return Some(object);
+            }
+
+            None
+        })
+        .collect::<Vec<yul::Object>>())
```
ralexstokes, line 26: consider if we can simplify this by having `module_stmt` returning an `Option<yul::Statement>` as in other places

compiler/src/abi/builder.rs
```diff
@@ -0,0 +1,248 @@
...
+type TypeDefs<'a> = HashMap<&'a str, &'a vyp::TypeDesc<'a>>;
+type TypeDef<'a> = (&'a str, &'a vyp::TypeDesc<'a>);
+
+/// Parse a map of contract ABIs from the input `module`.
+pub fn module<'a>(module: &'a vyp::Module<'a>) -> Result<ModuleABIs, CompileError> {
+    let type_defs = module
+        .body
+        .iter()
+        .filter(|s| is_type_def(&s.node))
+        .map(|s| type_def(&s.node))
```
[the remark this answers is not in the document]
g-r-a-n-t: In places where an `Err` is returned for receiving an incorrect enum variant, it would probably be better to panic (ideally this would be [dealt](https://github.com/rust-lang/rfcs/pull/2593) with at compile time). Result errors should be reserved for things like duplicate type definitions. If an error is returned for something like a duplicate type definition, `abi::build` should return an `Err` to notify the user of a duplicate type definition, so I'm a bit hesitant to filter that out. The other option is to destruct the enum and pass all of its values into the `type_def` function, same goes for all of the other `_def` functions, this way there is no need for a panic. This is the [pattern](https://github.com/g-r-a-n-t/rust-vyper/blob/types/compiler/src/yul/mappers/contracts.rs#L49) I ended up going with in the compiler. It's a bit clunky up-front, but keeps things organized overall. It would also be easy to refactor this into something ideal if enum variant types are introduced to the language.

You mentioned [Scan](https://doc.rust-lang.org/std/iter/struct.Scan.html) before, I'll look into that and see if I can tighten this up a bit. It might be best to just go with the same pattern found in the compiler where state is provided by scan instead of scope.

compiler/tests/evm_contracts.rs
g-r-a-n-t: I guess [deconstruct](https://docs.rs/evm/0.16.1/evm/executor/struct.StackExecutor.html#method.deconstruct) might be a better term...

The only way you can get events from the executer is by passing self by value into deconstruct, which isn't ideal. The comment isn't very useful tbh.

compiler/tests/evm_contracts.rs
g-r-a-n-t: Alright, I'll create some issues for these comment.

compiler/src/abi/builder.rs
```diff
@@ -0,0 +1,248 @@
...
+    Err(CompileError::static_str("Not a TypeDef"))
+}
+
+fn type_desc<'a>(
+    type_defs: &'a TypeDefs<'a>,
+    typ: &'a vyp::TypeDesc<'a>,
+) -> Result<VariableType, CompileError> {
+    if let vyp::TypeDesc::Base { base } = typ {
+        if let Some(custom_type) = type_defs.get(base) {
+            return type_desc(&HashMap::new(), custom_type);
```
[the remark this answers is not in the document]
g-r-a-n-t: Huh, thought I changed this

compiler/src/abi/elements.rs
g-r-a-n-t: If the ABIs built for contracts are used internally by the compiler for things like module imports, it would be best to have data structures that are easy to use as opposed to 1-1 reflections of the serialized JSON format.

ralexstokes: ah gotcha. perfect!

compiler/src/yul/mappers/module.rs
g-r-a-n-t: Maybe... `Vec<yul::Object>` is wrong and should at least be replaced with `HashMap<String, yul::Object>`, `yul::Statement`, or `yul::Object`. At the moment, I think `yul::Statement` is best. In the case of a contract-less module, we would just return the Yul statement `object "MyModule" { }` and with multiple contracts, we would return:

```
object "MyModule" {
  object "MyContract1" { ... }
  object "MyContract2" { ... }
}
```

I'll need to dig a bit deeper into Yul to figure out what's most appropriate. With Solc, I'm pretty sure it's possible to get the bytecode for nested Yul objects after compiling.

compiler/src/yul/runtime/functions.rs
g-r-a-n-t: Not that I'm aware of. It would definitely be useful to have those somewhere tho and `yultsur` seems like the right place.

Any input on this @axic? :)

compiler/src/yul/mappers/module.rs
g-r-a-n-t: I'll need to look into this further. Being able to return a single statement depends on whether or not we can retrieve the bytecode for each individual contract contained within that statement after compiling...

If this is not possible, it's probably better to return `HashMap<String, yul::Object>`, where `String` is a contract name.

compiler/src/abi/builder.rs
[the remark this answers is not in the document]
g-r-a-n-t: Having thought about this a bit more, it's probably pretty important that we pass the whole `Spanned` object into each mapper function.

e.g. `fn contract_def(scope: Shared<ModuleScope>,  def: Spanned<vyp::ModuleStmt>) -> Result<_, CompilerError>`

This way we can return line numbers and other useful contextual details back to the user if an error occurs. I think this gets back to an earlier point you made about *not throwing away useful information*.

Maybe the statement iterators could look something like this then:

```rust
module
  .body
  .iter()
  .try_fold(HashMap::new(), |mut contracts, stmt| {
    match &stmt.node {
      vyp::ModuleStmt::TypeDef { .. } => type_def(scope, stmt)?,
      vyp::ModuleStmt::SimpleImport { .. } => simple_import(scope, stmt)?,
      vyp::ModuleStmt::FromImport { .. } => from_import(scope, stmt)?,
      vyp::ModuleStmt::ContractDef { .. } => contracts.insert(contract_def(scope, stmt)?),
    };

    Ok(contracts)
  })
  .collect::<Result<HashMap<String, yul::Object>, CompileError>()?
```

It seems pretty clean to me and eliminates the double iteration.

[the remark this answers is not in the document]
ralexstokes: if this compiles it looks a lot simpler to me :)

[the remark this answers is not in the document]
g-r-a-n-t: Cool, updated this with [code](https://github.com/ethereum/rust-vyper/pull/18/commits/29289b723a104c7bdae6ec739591dbed2d7acb39#diff-0d15d418dfe7b76c76b4203800ee1d02R20) that actually compiles :)

It would probably be a good idea to have some sort of `CompileError` method that formats an error based on a given `Span`, which is something to consider along with this [comment](https://github.com/ethereum/rust-vyper/pull/15#discussion_r421826210). I think it makes sense to create another issue for this.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -0,0 +1,278 @@
...
+            Ok((identifier_expression! {(name)}, FixedSize::Base(base)))
+        }
+        Some(FunctionDef::Array(array)) => {
+            Ok((identifier_expression! {(name)}, FixedSize::Array(array)))
+        }
+        None => Err(CompileError::static_str("No definition found")),
+    }
+}
+
+// FIXME: The left side should be evaluated and then indexed.
```
g-r-a-n-t, line 70: Let me expand on this a bit.

Say we have the expression `self.messages[5]`, where `self.messages` is a `map<u256, bytes[100]>` declared in the contract scope. Currently, the `expr_subscript` mapper function takes the entire expression and filters it through a series of nested `expr_subscript_` calls that resolve to a `(yul::Expression, FixedSize)` tuple. It works for guest_book.vy, but is extremely limited in that you have to hard-code every possible left side (`self.<contract def>` in this case) to implement the language's desired behavior. I believe I just copied this pattern from the assignments module where it works better.

The `FIXME` comment vaguely implies that we should instead use the normal `expr` mapper function to map the left side and then combine it with the separately mapped slices to form a full expression, instead of hard-coding a bunch of nested functions. In order to do this though, we need to extend the `(yul::Expression, FixedSize)` tuple to include some more information about type and location. Possibly adding a struct like the one bellow and making it the default return for each expression mapper.

```rust
/// Extended expression
pub struct ExtExpression {
    pub expression: yul::Expression,
    pub location: Location,
    pub typ: Type,
}
```

In the case of `self.messages`, we would have an extended expression that looks something like this:

```
expression: 0 // storage index
location: Storage
type: Map<u256, bytes[100]>
```

With this information, we would be able to form the final extended expression for `self.messages[5]` in `expr_subscript` by combining the left (`self.messages`) and right (`[5]`) side into:

```
expression: dualkeccak(0, 5) // indexed key
location: Storage
type: bytes[100]
```

This way we can actually fully implement the language and remove unnecessary code in the expressions module.

Hopefully this clears things up, I was pretty confused by the comment at first.

compiler/src/abi/builder.rs
```diff
@@ -0,0 +1,248 @@
...
+    } = stmt
+    {
+        let inputs = args
+            .iter()
+            .map(|arg| func_def_arg(type_defs, &arg.node))
+            .collect::<Result<Vec<FuncInput>, CompileError>>()?;
+
+        let outputs = if let Some(return_type) = return_type {
+            vec![FuncOutput {
+                name: "".to_string(),
```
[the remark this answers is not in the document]
g-r-a-n-t: I don't think so...

```
func_def:
    [func_qual] 'def' NAME '(' [arg_list] ')' ['->' base_type] ':' block
```
from [Vyper.grammer](https://github.com/ethereum/rust-vyper/blob/master/parser/src/grammar/Vyper.grammar#L49)

compiler/src/yul/runtime/abi.rs
g-r-a-n-t: agreed. might justify a utils module

compiler/src/yul/runtime/functions.rs
```diff
@@ -0,0 +1,161 @@
...
+        free(),
+        ccopy(),
+        mcopy(),
+        scopy(),
+        mloadn(),
+        sloadn(),
+        cloadn(),
+        mstoren(),
+        sstoren(),
+        dualkeccak256(),
```
g-r-a-n-t, line 18: This should be removed in favor of a generalized `keccak256` function available on all fixed size types.

``` rust
let a = literal_expression! { 42 };
let b = literal_expression! { 12 };
let my_tuple = FixedSize::Tuple(vec![FixedSize::U256, FixedSize::U256]);

my_tuple.keccak256(a, b) // equivalent to dualkeccak256()
```

Issue #29, opened 2020-09-16
Fix parser tests


g-r-a-n-t:
These macro-generated [tests](https://github.com/ethereum/fe/blob/680911006347b8a1ed7051e8ffc014da43a70e29/parser/tests/test_parsers.rs#L284-L729) are not compiling. We see the following error for each test function:

```
error[E0425]: cannot find function `test_guest_book` in this scope
   --> parser/tests/test_parsers.rs:118:9
    |
118 | /         fn $tester_name() {
119 | |             do_with_fixtures!(
120 | |                 assert_fixture_parsed_with!($parser),
121 | |                 $fixture_path,
122 | |             );
123 | |         }
    | |_________^ not found in this scope
```

The `$tester_name` identifier in this particular case is set to `test_guest_book`. If I attempt to only generate the guest book case by deleting the other test cases and then explicitly writing `test_guest_book` in the macro, it successfully compiles and the test passes. In more detail:

```rust
// replace the existing `parse_fixture_tests!` invocation with...
parser_fixture_tests! {
    (
        file_input,
        test_guest_book,
        write_guest_book,
        "fixtures/parsers/guest_book.ron",
    ),
}
```

```rust
// replace `$tester_name` with `test_guest_book` in the macro definition
macro_rules! parser_fixture_tests {
    ($(($parser:expr, $tester_name:ident, $writer_name:ident, $fixture_path:expr,)),+,) => {
        $(
        #[test]
        #[wasm_bindgen_test]
        fn test_guest_book() {
            do_with_fixtures!(
 ...
```

It looks like this may be related to [macro hygiene](https://danielkeep.github.io/tlborm/book/mbe-min-hygiene.html), but I'm not sure.

PR #30, opened 2020-09-16
Build fixes

g-r-a-n-t:
Cleaned some things up so that the CI build passes.

- featurized the solc bytecode compiler under `solc-backend`.
  - now you must run `cargo test --features solc-backend` to compile with solc and to run the contract tests
- ran `cargo fmt`, which updated alot of files
- commented out some parser tests which are now failing. see #29


cburgdorf:
Am I right to assume we do not run `cargo test --features solc-backend` in CI yet because the build takes too long? If that's the case, do you already have an idea how to overcome that. I create an issue here to discuss that further. https://github.com/ethereum/fe/issues/32

PR #33, opened 2020-09-17
Rename all Vyper things to Fe things

cburgdorf:
### What was wrong

Still many Vyper legacies

### How was it fixed

Renamed everything


g-r-a-n-t (approved):
looks good :+1: thanks

Issue #24, opened 2020-09-14
Arithmetic expression mapping for U256 values


g-r-a-n-t:
The following Fe expressions should be mapped to Yul, like so:

| Fe        | Yul           |
| ---        | ---             |
| `a + b` | `add(a', b')` |
| `a - b`  | `sub(a', b')` |
| `a * b`  | `mul(a', b')` |
| `a  /  b` | `div(a', b')`  |

Where *a* and *b* are `vyp::Expr`s and *a'* and *b'* are `yul::Expression`s.

**todo:**
- [ ] call a new function `expr_bin_operation` [here](https://github.com/ethereum/fe/blob/89d0e5ae1c0e618c8df54933ffdc0d0985687f15/compiler/src/yul/mappers/expressions.rs#L40) that returns an `ExtExpression`.
- [ ] add a match statement to `expr_bin_operation` that matches the `Add`, `Sub`, `Mult`, and `Div` operations and calls more new functions `expr_add`, `expr_sub`, `expr_mult`, and `expr_div`. (mark all other branches as unimplemented)
- [ ] match both the left and right `ExtExpression` types (i.e. `(Base::U256, Base::U256)`) in each of the `expr_` functions and return a new ExtExpression of type `U256` with an expression containing the correct arithmetic function.
- [ ] add a test contract with a function that takes two params, adds them together, and returns the result.

*Note: the `Base::U256` enum value should contain a `U256` struct with arithmetic functions attached to it. However, this is not necessary at the moment.*

Fits into #21.

PR #36, opened 2020-09-18
Add support for basic arithmetic operations

cburgdorf:
### What was wrong

As stated in #24 we need to add support for basic arithmetic operations.

### How was it fixed

- Resolving binary expressions for `add`, `sub`, `mul`, `div` to their YUL counterparts
- Added tests in expressions, assignments as well as evm_contracts.

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -50,7 +50,7 @@ pub fn expr(
         vyp::Expr::Attribute { .. } => expr_attribute(scope, exp),
         vyp::Expr::Ternary { .. } => unimplemented!(),
         vyp::Expr::BoolOperation { .. } => unimplemented!(),
-        vyp::Expr::BinOperation { .. } => unimplemented!(),
+        vyp::Expr::BinOperation { .. } => expr_bin_operation(scope, exp),
         vyp::Expr::UnaryOperation { .. } => unimplemented!(),
         vyp::Expr::CompOperation { .. } => unimplemented!(),
         vyp::Expr::Call { .. } => unimplemented!(),
@@ -60,6 +60,42 @@ pub fn expr(
         vyp::Expr::Str(_) => unimplemented!(),
         vyp::Expr::Ellipsis => unimplemented!(),
     }
+}
+
+pub fn expr_bin_operation<'a>(
+    scope: Shared<FunctionScope>,
+    exp: &Spanned<vyp::Expr<'a>>,
+) -> Result<ExtExpression, CompileError> {
+    if let vyp::Expr::BinOperation { left, op, right } = &exp.node {
+        let yul_left = expr(Rc::clone(&scope), left)?.expression;
+        let yul_right = expr(Rc::clone(&scope), right)?.expression;
+
+        return match op.node {
+            vyp::BinOperator::Add => Ok(ExtExpression {
+                expression: expression! { add([yul_left], [yul_right]) },
+                location: Location::Value,
+                typ: Type::Base(Base::U256),
+            }),
+            vyp::BinOperator::Sub => Ok(ExtExpression {
+                expression: expression! { sub([yul_left], [yul_right]) },
+                location: Location::Value,
+                typ: Type::Base(Base::U256),
+            }),
+            vyp::BinOperator::Mult => Ok(ExtExpression {
+                expression: expression! { mul([yul_left], [yul_right]) },
+                location: Location::Value,
+                typ: Type::Base(Base::U256),
+            }),
+            vyp::BinOperator::Div => Ok(ExtExpression {
+                expression: expression! { div([yul_left], [yul_right]) },
+                location: Location::Value,
+                typ: Type::Base(Base::U256),
+            }),
+            _ => unimplemented!(),
+        };
+    }
+
+    unreachable!()
 }
 
 /// Retrieves the &str value of a name expression.
@@ -271,6 +307,7 @@ mod tests {
         Type,
     };
     use fe_parser as parser;
+    use rstest::rstest;
     use std::rc::Rc;
 
     fn scope() -> Shared<FunctionScope> {
@@ -401,5 +438,19 @@ mod tests {
         assert_eq!(result.expression.to_string(), "caller()");
         assert_eq!(result.location, Location::Value);
         assert_eq!(result.typ, Type::Base(Base::Address));
+    }
+
+    #[rstest(
+        expression,
+        expected_yul,
+        case("1+1", "add(1, 1)"),
+        case("1-1", "sub(1, 1)"),
+        case("1*1", "mul(1, 1)"),
+        case("1/1", "div(1, 1)")
+    )]
+    fn arithmetic_expression(expression: &str, expected_yul: &str) {
+        let result = map(scope(), expression);
+
+        assert_eq!(result.expression.to_string(), expected_yul);
     }
 }
--- a/compiler/tests/evm_contracts.rs
+++ b/compiler/tests/evm_contracts.rs
@@ -8,6 +8,7 @@ use primitive_types::{
     H160,
     U256,
 };
+use rstest::rstest;
 use std::collections::BTreeMap;
 use std::fs;
 use std::iter;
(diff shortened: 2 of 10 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -62,6 +62,27 @@ pub fn expr(
...
+
+pub fn expr_bin_operation<'a>(exp: &Spanned<vyp::Expr<'a>>) -> Result<ExtExpression, CompileError> {
+    if let vyp::Expr::BinOperation{ left, op, right } = &exp.node {
+        // FIXME: How do I get what I need for the `add`
+        let yul_left = literal_expression! {(left.node)};
+        let yul_right = literal_expression! {(right.node)};
+        return match op.node {
+            vyp::BinOperator::Add => {
+                Ok(ExtExpression {
+                    expression: expression! { add([yul_left], [yul_right]) },
```
cburgdorf, line 74: @g-r-a-n-t Please ignore all the sloppiness in this code. Can you poke me in the right direction how to construct the `add(?, ?)` here? I need to get to "whats behind" the left and right nodes but I'm a bit lost how exactly.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -62,6 +62,27 @@ pub fn expr(
     }
 }
 
+
+pub fn expr_bin_operation<'a>(exp: &Spanned<vyp::Expr<'a>>) -> Result<ExtExpression, CompileError> {
+    if let vyp::Expr::BinOperation{ left, op, right } = &exp.node {
+        // FIXME: How do I get what I need for the `add`
+        let yul_left = literal_expression! {(left.node)};
+        let yul_right = literal_expression! {(right.node)};
```
g-r-a-n-t, line 70: looks good so far!

The correct left and right expressions can be found by calling `expr(...)` again, like we do [here](https://github.com/ethereum/fe/blob/4bd735adac4df4f7cf805a07de7a2348aee6ec0a/compiler/src/yul/mappers/expressions.rs#L154).

`let yul_left = expr(Rc::clone(&scope), left)?;`

We just need to pass in the scope `expr_bin_operation(scope: Shared<FunctionScope>, ...)`.

FYI: `Shared<T>` is an alias for `Rc<RefCell<T>>`. The pattern is described in more detail [here](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html#having-multiple-owners-of-mutable-data-by-combining-rct-and-refcellt). You won't need to do anything with scope here other than providing it to the `expr` call.

Let me know if what I've suggested doesn't work.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -62,6 +62,27 @@ pub fn expr(
...
+        return match op.node {
+            vyp::BinOperator::Add => {
+                Ok(ExtExpression {
+                    expression: expression! { add([yul_left], [yul_right]) },
+                    location: Location::Memory,
```
g-r-a-n-t, line 75: For now, U256 values on their own are always stored as values. So we'll need: `Location::Value`.

compiler/src/yul/mappers/expressions.rs
cburgdorf: Thanks, it worked!

>FYI: Shared<T> is an alias for Rc<RefCell<T>>. The pattern is described in more detail here. 

Thanks, I'm loosely familiar with the pattern. :+1:

compiler/tests/evm_contracts.rs
```diff
@@ -188,6 +188,20 @@ fn return_u256() {
...
+        let harness = deploy_contract(&mut executor, "return_product_u256.vy", "Foo");
+
+        harness.test_function(
+            &mut executor,
+            "bar",
+            vec![u256_token(42), u256_token(42)],
+            Some(u256_token(84)),
+        )
+    })
+}
```
cburgdorf, line 203: @g-r-a-n-t regarding #37 Instead of repeating the above test for multiplication, substraction etc, we could use parameters like this to put it all in one test.

```
#[rstest(input, expected,
    case("return_addition_u256.vy", vec![42, 42, 84]),
    ...
    case("return_substraction_u256.vy", vec![42, 42, 0),
    ...
    case("return_multiplication_u256.vy", vec![5, 5, 25),
    ...
)]
```

cburgdorf: At least, that is what I would do if this was Py-EVM / Trinity code. Interested to hear your thoughts..

g-r-a-n-t: I think it looks good.

I also wonder if we could do the same thing here #29, where we're trying to automatically generate tests with macros (which doesn't work anymore) :thinking:

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: awesome!

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: I looked at this a bit more. Any location would work here :| since we're not checking the location when [encoding](https://github.com/ethereum/fe/blob/4bd735adac4df4f7cf805a07de7a2348aee6ec0a/compiler/src/yul/namespace/types.rs#L101) base types, but it should still be `Location::Value`.

cburgdorf: Yep, I'll change that.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -62,6 +62,48 @@ pub fn expr(
...
+                })
+            }
+            vyp::BinOperator::Mult => {
+                Ok(ExtExpression {
+                    expression: expression! { mul([yul_left], [yul_right]) },
+                    location: Location::Value,
+                    typ: Type::Base(Base::U256),
+                })
+            },
+            vyp::BinOperator::Div => {
```
cburgdorf, line 93: @g-r-a-n-t  You mentioned in #24 that you would separate each arm into its own function. I was wondering if there was a strong argument in favor for that (beside just separation of concerns) because I believe it might just unnecessary bloat loc here. I can be convinced but wanted to here the reasoning :)

cburgdorf:
@g-r-a-n-t There are a bunch of other binary operations and I'm happy to implement these but since #24 was just about these basic four I thought I cut it short here and let you review this as a self-contained package first.

compiler/tests/evm_contracts.rs
```diff
@@ -188,6 +189,27 @@ fn return_u256() {
     })
 }
 
+
+#[rstest(fixture_file, input, expected,
+    case("return_addition_u256.vy", vec![42, 42], Some(84)),
+    case("return_subtraction_u256.vy", vec![42, 42], Some(0)),
+    case("return_multiplication_u256.vy", vec![42, 42], Some(1764)),
+    case("return_division_u256.vy", vec![42, 42], Some(1)),
```
cburgdorf, line 197: There's stuff like division by zero that we could test here but I don't want to open a can of worms and put it in an issue for now https://github.com/ethereum/fe/issues/41

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: I'm ok with this. Also, these functions will get even lighter with the semantic analysis pass.

compiler/tests/evm_contracts.rs
g-r-a-n-t: Yeah, for now I think it's ok to put things like this off. Below are some thoughts on how I think this should be done.

My thinking right now is that we will introduce *safe* math functions to the runtime, where we check for things like overflow and div-by-zero. Then instead of mapping `a / b` to `div(a', b')`, we map to `safe_div(a', b')`, where safe_div is a function that we've defined that reverts on a zero denominator.

I think the semantic analysis pass should be responsible for determining what functions are needed during runtime and then we'll include them in the Yul contract when mapping from the Fe contract. This approach would be extended to things like maps where we have generic type params. For example, defining the contract field `pub guest_book: map<address, bytes[100]>` would result in two function being added to the runtime named something like *map_get_address_bytes100* and *map_insert_address_bytes100*, which would be used in place of the expression `self.guest_book[0x00]` and statement `self.guest_book[0x00] = some_bytes`.

cburgdorf: Great feedback, I put it on #41

PR #39, opened 2020-09-22
Compiler decomposition

### LINKED ISSUE #31: Decompose Fe -> Yul compiler pass
Currently, we have a single compiler pass that maps Fe ASTs generated by the parser to Yul. This pass is responsible for both context-sensitive analysis and mapping Fe to Yul. As the project grows in complexity, it will become difficult to manage both of these responsibilities in one pass and would result in a codebase that's difficult to audit.

We can improve the situation by introducing a pass before the Fe to Yul mapper that is responsible for both deeper validation of the source program and for providing the mapper pass with enough information to correctly map expressions.

Let's illustrate with the guest book contract:

```
/* guest_book.vy */
type BookMsg = bytes[100]

contract GuestBook:
    pub guest_book: map<address, BookMsg>

    event Signed:
        idx book_msg: BookMsg

    pub def sign(book_msg: BookMsg):
        self.guest_book[msg.sender] = book_msg

        emit Signed(book_msg=book_msg)

    pub def get_msg(addr: address) -> BookMsg:
        return self.guest_book[addr]
```

**Compilation of the returned expression in `get_msg` goes roughly like this:**

*context/mapper pass*
- gets the type of `self.guest_book` from the contract scope and maps it to a Yul expression (its storage index) then returns this information to the caller.
- gets the type information for `addr` from the function scope and maps it to a Yul expression then returns this information to the caller
- determines the type of `self.guest_book[addr]` given the type of `self.guest_book` and maps it to a Yul expression then returns this information to the caller

**It should instead go like this:**

*context pass*
- gets the type of `self.guest_book` from the contract sope and stores it
- gets the type of `addr` from the function scope and stores it
- determines the type of `self.guest_book[addr]` and stores it

*mapper pass*
- maps `self.guest_book` to a Yul expression given the expression type from the previous pass
- maps `addr` to a Yul expression given the expression type from the previous pass
- maps `self.guest_book[addr]` to a Yul expression given the expression type from the previous pass

The second method separates concerns nicely and conforms more closely with standard compiler design.

**todo:**
Determine what the data structure built up by the context pass should be.

g-r-a-n-t:
closes #31 

The changes in this PR follow what's been described in the issue mentioned above. The general goal was to pull apart the existing compiler pass into two passes, where one focuses on analyzing the semantics of a source program and the other focuses on mapping Fe expressions to Yul.

In the semantic pass, we [traverse](https://github.com/ethereum/fe/tree/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/traversal) the AST similar to how we traverse the AST in the compiler [mapper](https://github.com/ethereum/fe/tree/598d4c03d3431c8681f3a4dc8dc274f79d50235d/compiler/src/yul/mappers) module. The difference, though, is that instead of mapping the Fe source nodes to Yul, we accumulate contextual information using [scopes](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/namespace/scopes.rs#L65), check the validity of the source program, and attribute contextual information to AST nodes using [`Context`](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/lib.rs#L53), which is then used by the mapper pass to correctly map nodes.

For example, the semantic pass would be responsible for determining whether or not `self.foo[bar]` is a valid expression given its contextual knowledge. It would also be responsible for storing contextual information that is needed by the mapper. In the example `self.foo[bar]`, the mapper needs to know the type of `self.foo` to properly form a Yul expression. The semantic pass needs to make this information readily available to the mapper using `context`. It does this by [attributing](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/lib.rs#L37) information about the expression to the AST node of `self.foo`, which is then used by the mapper to retrieve needed context information (`context.get_expression(value)`). Below is the refactored `expr_subscript` mapper, which uses the method just described.

https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/compiler/src/yul/mappers/expressions.rs#L163-L179

The [API](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/lib.rs#L61-L140) of context was built out only to the extent needed by this PR. As it expands, it should remain consistent in a couple regards: 1.) there should only be one pair of `add`/`get` methods for each type of AST node and 2.) if multiple things need to be attributed to one type of AST node, they should be grouped together using an attributes struct e.g. `ContractAttributes`. Also, attributing information to AST nodes should be done somewhat liberally. For example, there may be duplicate information attributed to both contract and function definitions. The goal is to make any information that's helpful during mapping immediately available at the point where a node is being mapped so that the mapper pass does not need to analyze context.

There are still some things that need to be refactored, but for the sake of keeping this PR at a reasonable size, they were ignored. The best example of this is that we still use methods defined within our [types](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/namespace/types.rs) module to form Yul expressions and statements. These Yul elements should instead be formed exclusively in the mapper pass.

There are a number of spots where I have left [todo stubs](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/traversal/assignments.rs#L60) for type checking. These should be taken care of in smaller subsequent PRs.


semantics/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Contains contextual information about a Fe module and can be queried using
+/// `Spanned` AST nodes.
+#[derive(Clone, Debug, Default)]
+pub struct Context {
+    expressions: HashMap<Span, ExpressionAttributes>,
+    emits: HashMap<Span, Event>,
+    functions: HashMap<Span, FunctionAttributes>,
+    declarations: HashMap<Span, FixedSize>,
+    contracts: HashMap<Span, Vec<FunctionAttributes>>,
```
cburgdorf, line 58: `contracts` holding a `HashMap<Span, Vec<FunctionAttributes>>` makes it look as if a contract would only hold functions but it can also hold declarations and events.

https://github.com/ethereum/fe/blob/503d7bb9c2d4241e6a6b38982756ae875714d775/compiler/tests/fixtures/guest_book.fe#L4-L7

I'm wondering if instead of `FunctionAttributes` a type should be used that would combine `FunctionAttributes`, `Event`  and `FixedSize`?

cburgdorf: Oh, I just finished reading the whole PR description and noticed:

>The API of context was built out only to the extent needed by this PR

>f multiple things need to be attributed to one type of AST node, they should be grouped together using an attributes struct e.g. `ContractAttributes`

I *think* that this is basically saying what I wrote above but waiting for you to confirm. If yes, than I think a small comment above line `58` would make a nice improvement in the meantime.

compiler/src/yul/mappers/contracts.rs
```diff
@@ -2,46 +2,30 @@ use crate::errors::CompileError;
...
             |mut statements, stmt| {
                 match &stmt.node {
-                    fe::ContractStmt::ContractField { .. } => {
-                        contract_field(Rc::clone(&contract_scope), stmt)?
-                    }
-                    fe::ContractStmt::EventDef { .. } => {
-                        event_def(Rc::clone(&contract_scope), stmt)?
-                    }
+                    fe::ContractStmt::ContractField { .. } => {}
+                    fe::ContractStmt::EventDef { .. } => {}
```
cburgdorf, line 26: These empty blocks look suspicious, did we lose existing functionality?

compiler/src/yul/mappers/expressions.rs
```diff
@@ -62,116 +38,54 @@ pub fn expr(
...
+                Ok(expression! { iszero([expression! { eq([yul_left], [yul_right]) }]) })
+            }
+            fe::CompOperator::Lt => Ok(expression! { lt([yul_left], [yul_right]) }),
+            fe::CompOperator::LtE => {
+                Ok(expression! { iszero([expression! {gt([yul_left], [yul_right])}]) })
+            }
+            fe::CompOperator::Gt => Ok(expression! { gt([yul_left], [yul_right]) }),
+            fe::CompOperator::GtE => {
+                Ok(expression! { iszero([expression! {lt([yul_left], [yul_right])}]) })
+            }
```
cburgdorf, line 61: Nice to see how compact this has become.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -215,310 +129,206 @@ pub fn spanned_expression<'a>(span: &Span, exp: &fe::Expr<'a>) -> Spanned<fe::Ex
...
-        assert_eq!(
-            result.expression.to_string(),
-            "sloadn(dualkeccak256(0, 3), 32)"
-        );
-        assert_eq!(result.location, Location::Value);
-        assert_eq!(result.typ, Type::Base(Base::U256));
-    }
-
-    #[test]
-    fn map_sload_array_and_address() {
```
cburgdorf, line 431: Was this test dropped without replacement?

compiler/src/yul/runtime/abi.rs
```diff
@@ -1,22 +1,16 @@
...
-pub fn dispatcher(
-    interface: &[String],
-    defs: &HashMap<String, ContractDef>,
-) -> Result<yul::Statement, CompileError> {
-    let arms = interface
+pub fn dispatcher(attributes: Vec<FunctionAttributes>) -> Result<yul::Statement, CompileError> {
+    let arms = attributes
         .iter()
-        .map(|name| dispatch_arm(name.to_owned(), defs))
+        .map(|a| dispatch_arm(a.to_owned()))
```
cburgdorf, line 13: As a team convention we try to avoid single letter variable names as much as possible. (Except for life time annotations and generic type params)

semantics/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "fe-semantics"
+version = "0.1.0"
+authors = ["Grant Wuerker <gwuerker@gmail.com>"]
```
cburgdorf, line 4: Can we make this `"Ethereum Foundation <snakecharmers@ethereum.org>"`. This also reminds me that I still need to double check why the license was changed (a long time ago) from MIT to  Apache 2.0. I think if we want to stick with Apache then at least we should dual license with MIT (that's what Rust does) because most of our projects are MIT licensed.

cburgdorf (approved):
Great work!

This is quite a huge PR and touches a lot of parts that I'm not yet super familiar with. But as far as I understand the changes, they make sense to me and look good.

I left a few comments inline but just some minor things to address.

semantics/src/lib.rs
g-r-a-n-t: Yeah, I can add a comment here.

compiler/src/yul/mappers/contracts.rs
g-r-a-n-t: Both contract fields and event definitions only modify scope, so there is nothing to map.

Modifications to scope happen [here](https://github.com/ethereum/fe/blob/598d4c03d3431c8681f3a4dc8dc274f79d50235d/semantics/src/traversal/contracts.rs#L55-L91) in the semantic pass.

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: Yeah, it looks like this test was originally added to make sure we can map expressions with multiple storage values in scope. It didn't really seem necessary to keep this anymore since things are more straight forward with context. It would basically test the other two map tests in the same test, which seems a bit redundant.

compiler/src/yul/runtime/abi.rs
g-r-a-n-t: I can update these.

Issue #34, opened 2020-09-17
CI doesn't seem to run on PRs


cburgdorf:
CI doesn't seem to run on PRs. Also, *maybe* it's worth migrating to Github Actions.

PR #45, opened 2020-09-22
Setup Github Actions

cburgdorf:
### What was wrong

As stated in #34. We do currently not run Circle CI on PRs. Or better, we do, but they don't show up here and I couldn't figure out how to resolve that.

### How was it fixed

Migrated to Github Actions instead


cburgdorf:
image

@g-r-a-n-t this should be doing everything that the CircleCI build used to do.

.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    - name: Lint with clippy
+      uses: actions-rs/cargo@v1
+      with:
+        command: clippy
+        args: --all-targets --all-features
+
+  test:
+
+    # TODO: Need to enable Mac and Windows builds [ubuntu-latest, macos-latest windows-latest]
```
cburgdorf, line 53: @g-r-a-n-t On Windows we do have an actual bug that should be fixable. Mac seemed to be a GitHub Actions issue (others reported the same issue). We need to look into this but since that's not a regression I don't think it should hold us back now.

g-r-a-n-t (approved):
looks good!

Issue #40, opened 2020-09-22
Start language specification


g-r-a-n-t:
We will need to write a specification so that it's possible to audit this project. It would also just be useful to have in general for ourselves and for users.

This is the [specification](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf) for c++. It might help give us an idea of how our specification should be structured.

For this issue we should do two things:
- [x] create a new directory named *spec* with an empty README 
- [x] create a couple issues for things that can be easily specified at the moment

cburgdorf:
Throwing in the [Rust reference](https://doc.rust-lang.org/stable/reference/introduction.html) (Not to be confused with the [Rust book](https://doc.rust-lang.org/book/) or [Std library docs](https://doc.rust-lang.org/stable/std/index.html) ) as another good example. Even though they mention it is not 100 % the specification but the closest thing to a specification that exists today.

>Finally, this book is not normative. It may include details that are specific to rustc itself, and should not be taken as a specification for the Rust language. We intend to produce such a book someday, and until then, the reference is the closest thing we have to one.

cburgdorf:
For reference: We have an [early draft](https://github.com/ethereum/fe/blob/master/spec/index.md) in master. Basic structure / system is heavily influenced by the [Rust reference](https://doc.rust-lang.org/stable/reference/introduction.html)

PR #50, opened 2020-09-23
Start language specification

cburgdorf:
### What was wrong?

As stated in #40 we need to start writing a language specification.

### How was it fixed?

Not fixed but wanted to break ground after implementing the arithmetic operations.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1,0 +1,58 @@
+# Fe Language Specification
+
+<div class="warning">
+  Warning: This is a work in progress document. It is incomplete and specifications aren't stable yet.
+</div>
+
+## Arithmetic Operators
+
+> **<sup>Syntax</sup>**\
+> _ArithmeticExpression_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_] `+` [_Expression_]\
+> &nbsp;&nbsp; | [_Expression_] `-` [_Expression_]\
+> &nbsp;&nbsp; | [_Expression_] `*` [_Expression_]\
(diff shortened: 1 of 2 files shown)


cburgdorf:
This is heavily influenced by https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html

spec/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+<div class="warning">
+  Warning: This is a work in progress document. It is incomplete and specifications aren't stable yet.
+</div>
+
+## Arithmetic Operators
+
+> **<sup>Syntax</sup>**\
+> _ArithmeticExpression_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_Expression_] `+` [_Expression_]\
```
g-r-a-n-t, line 11:
```suggestion
> &nbsp;&nbsp; | [_Expression_] `+` [_Expression_]\
```

cburgdorf:
@g-r-a-n-t are you ok to merge this? I think we can just continue go make changes to this document over time to build out the specification piece by piece.

spec/index.md
cburgdorf: Ah, that's on purpose. It's because the `|` is used as an `or` and since this is the first operator it doesn't have an or and just uses `&nbsp;&nbsp;` for alignment. See the original for reference https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html

g-r-a-n-t: I see :+1:

g-r-a-n-t (approved):

Issue #8, opened 2019-12-17
Possible introduction of prefix keywords or qualifiers


davesque:
There are a lot of situations where Rust Vyper could emulate Python Vyper's syntax, but it seems sub-optimal to do that because Python Vyper's syntax seems to have been designed mostly to accommodate the Python parser.

From the point of view of Rust Vyper, it seems more and more like the Python parser is underpowered because it was designed for a dynamic language in which methods and variables don't require much annotation.  This is because there aren't many restrictions on how methods and variables can be used in Python.

Here's a concrete example.  If Rust Vyper followed Python Vyper's syntax, here are two different ways that things would be annotated to indicate that they're "public":
```
contract GuestBook:
    guest_book: public(map[address, bytes[100]])

    @public
    def sign(book_msg: bytes[100]):
        ...
```
Frankly, this is ugly and here are some reasons why:
* Storage variables and methods are being annotated in two entirely different ways.  Storage variables use something like a "pseudo-function" called `public` that "returns" a new type `public(map[address, bytes[100]])`.  On the other hand, methods use the python decorator syntax.  Public methods are annotated by placing a `@public` directive on the line preceding the method declaration.
* Being forced to use the pseudo-function syntax for variable annotations means that having more than one annotation (if we ever want a feature like that) makes things very unreadable and forces the inclusion of lots of nested parentheses.

Now consider this alternative syntax:
```
contract GuestBook:
    pub guest_book: map[address, bytes[100]]

    pub def sign(book_msg: bytes[100]):
        ...
```
In my opinion, this is far more readable for the following reasons:
* Both property and method annotations use *the same* syntax.
* The annotations are at the front of the line where they are most visible.
* The keyword "pub" is shorter but no less expressive.

It seems like we could also elect to keep the decorator syntax in case we also want modifiers (which are a feature that I think was unfairly excluded from Vyper).  But the qualification of a method as public or not could use a prefix keyword on the line of the method declaration.

davesque:
By the way, just to head off this discussion if it happens, I'm aware that there's really no difference in Python semantically between the `public(...)` syntax and the `@public` syntax.  The latter is just syntactic sugar for application of a decorator to a function object which could also be done by using the function call syntax `public(...)`.  However, in Vyper, these semantics are irrelevant and shouldn't inform any design decisions about the syntax.

cburgdorf:
[...]

PR #51, opened 2020-09-24
More specifications

cburgdorf:
### What was wrong?

As I continue to dive deeper I realized that beginning with the language specification might actually be one of the most valuable things we can do. It ensures everyone is working towards a common, shared vision of the language rather than "making things up as we go".

I think that by writing the language specification early on, we sooner get to the point where we create some value for the community. Even if we don't have anything that compiles, just having the spec alone and be able to gather community feedback is a great benefit in itself.

This builds on top of #40 

### How was it fixed?

This is very incomplete and I'm sure it has tons of box. I'm leaning heavily on the  [Rust reference](https://doc.rust-lang.org/stable/reference/introduction.html) for structure and syntax of the document (We love you Rust :crab: :heart: )

All links in the document should be working and a human readable version with working links can be reviewed at: https://github.com/ethereum/fe/blob/4c8481b205823beda0b75dcbfb847a61c372be83/spec/index.md

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -4,11 +4,385 @@
   Warning: This is a work in progress document. It is incomplete and specifications aren't stable yet.
 </div>
 
-## Arithmetic Operators
+# 1. Notation
+
+## 1.1 Grammar
+
+The following notations are used by the *Lexer* and *Syntax* grammar snippets:
+
+| Notation          | Examples                      | Meaning                                   |
+|-------------------|-------------------------------|-------------------------------------------|
+| CAPITAL           | KW_IF                         | A token produced by the lexer             |
+| _ItalicCamelCase_ | _Item_                        | A syntactical production                  |
+| `string`          | `x`, `while`, `*`             | The exact character(s)                    |
+| \\x               | \\n, \\r, \\t, \\0            | The character represented by this escape  |
+| x<sup>?</sup>     | `pub`<sup>?</sup>             | An optional item                          |
+| x<sup>\*</sup>    | _OuterAttribute_<sup>\*</sup> | 0 or more of x                            |
+| x<sup>+</sup>     |  _MacroMatch_<sup>+</sup>     | 1 or more of x                            |
+| x<sup>a..b</sup>  | HEX_DIGIT<sup>1..6</sup>      | a to b repetitions of x                   |
+| \|                | `u8` \| `u16`, Block \| Item  | Either one or another                     |
+| [ ]               | [`b` `B`]                     | Any of the characters listed              |
+| [ - ]             | [`a`-`z`]                     | Any of the characters in the range        |
+| ~[ ]              | ~[`b` `B`]                    | Any characters, except those listed       |
+| ~`string`         | ~`\n`, ~`*/`                  | Any characters, except this sequence      |
+| ( )               | (`,` _Parameter_)             | Groups items                              |
+
+
+
+## 2. Lexical structure
+
+### 2.1. Keywords
+
+Fe divides keywords into two categories:
+
+* [strict](#strict-keywords)
+* [reserved](#reserved-keywords)
+
+#### 2.1.1. Strict keywords
+
+These keywords can only be used in their correct contexts. They cannot
+be used as the names of:
+
+* [Items]
+* [Variables] and function parameters
+* Fields and [variants]
+
+> **<sup>Lexer:<sup>**\
+> KW_AS             : `as`\
+> KW_BREAK          : `break`\
+> KW_CONST          : `const`\
+> KW_CONTINUE       : `continue`\
+> KW_CONST          : `contract`\
+> KW_DEF            : `def`\
+> KW_ELIF           : `elif`\
+> KW_ELSE           : `else`\
+> KW_EMIT           : `emit`\
+> KW_ENUM           : `enum`\
+> KW_EVENT          : `event`\
+> KW_FALSE          : `false`\
+> KW_FOR            : `for`\
+> KW_IDX             : `idx`\
+> KW_IF             : `if`\
+> KW_IN             : `in`\
+> KW_LET            : `let`\
+> KW_NONPAYABLE     : `nonpayable`\
+> KW_PASS           : `pass`\
+> KW_PAYABLE        : `payable`\
+> KW_PUB            : `pub`\
+> KW_RETURN         : `return`\
+> KW_REVERT         : `revert`\
+> KW_SELFVALUE      : `self`\
+> KW_STRUCT         : `struct`\
+> KW_TRUE           : `true`\
+> KW_WHILE          : `while`
+
+
+#### 2.1.2. Reserved keywords
+
+These keywords aren't used yet, but they are reserved for future use. They have
+the same restrictions as strict keywords. The reasoning behind this is to make
+current programs forward compatible with future versions of Fe by forbidding
+them to use these keywords.
+
+> **<sup>Lexer</sup>**\
+> KW_ABSTRACT       : `abstract`\
+> KW_ABSTRACT       : `async`\
+> KW_ABSTRACT       : `await`\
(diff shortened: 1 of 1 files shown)


g-r-a-n-t:
more info on `pub` [here](https://github.com/ethereum/fe/issues/8)

cburgdorf:
Ha! Great digging. I originally [gave :+1: on that](https://github.com/ethereum/fe/issues/8#issuecomment-566667835) but I'm not entirely sure if I still agree with it. Let's discuss that in our meeting later.

cburgdorf:
Re-reading what was written back then, it seems the main motivation was not having two different ways (in different scenarios) to define things as public. I believe the concrete example could also be overcome if both would use decorator syntax.

```
contract GuestBook:
   @public 
   guest_book: map[address, bytes[100]]
   @public
   def sign(book_msg: bytes[100]):
```

g-r-a-n-t (commented):
looks sharp!

spec/index.md
```diff
@@ -0,0 +1,308 @@
...
+**External:** External functions are part of the contract interface, which means they can be called from other contracts and via transactions.
+
+**Public:** Public functions are part of the contract interface and can be either called internally or from other contracts.
+
+**Private:** Those functions and state variables can only be accessed internally from within the same contract. This is the default visibility.
+
+For example, this is a function that can be called externally from a transaction:
+
+```python
+@external
```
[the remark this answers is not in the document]
cburgdorf: I looked into vyper and it seems that they indeed only get away with `internal` and `external` so I'm happy to stick with `pub`. I think we still may want to keep decorator syntax for things like `payable` or other attributes a function may have. But I'm happy to keep that undefined for now.

That [grammar](https://github.com/ethereum/fe/blob/master/parser/src/grammar/Vyper.grammar) file that you linked, is that being used for anything or does it only serve as a form of minimal syntax spec?

I certainly don't want to spec out far away future features just yet. That said, I think it's valuable to at least *draft spec* the fundamental syntax of the core concepts the language. I think that is valuable to build up common understanding and a shared goal of what the language will eventually will be.

spec/index.md
```diff
@@ -0,0 +1,308 @@
...
+When referred to, a _function_ yields a first-class *value* of the
+corresponding zero-sized [*function item type*], which
+when called evaluates to a direct call to the function.
+
+A function header ends with a colon (`:`) after which the function body begins.
+
+For example, this is a simple function:
+
+```python
+def answer_to_life_the_universe_and_everything() -> u32:
```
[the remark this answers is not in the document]
cburgdorf: Yeah, I can change that to `u256` but I think we eventually want support for all multiples of 8 up to 256, right?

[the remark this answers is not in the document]
g-r-a-n-t: Yeah, we should have `u8`, `u16`, `u32`, ..., `u256`. No need to change, was just double checking since some type systems use bytes for sizing.

spec/index.md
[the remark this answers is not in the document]
g-r-a-n-t: > That grammar file that you linked, is that being used for anything or does it only serve as a form of minimal syntax spec?

minimal syntax spec

> I think we still may want to keep decorator syntax

we definitely do

spec/index.md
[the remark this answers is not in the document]
g-r-a-n-t: So I just noticed that the test contracts have `uint256` instead of `u256`... This is something that isn't specified anywhere, even in the grammar. I would be open to changing this if you think `u256` is cleaner, as it's a fairly trivial change.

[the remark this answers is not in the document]
cburgdorf: Sounds good. I'm :+1: using `uN`  and `iN` mainly because it saves some space without compromising on readability and is inline with rust.

spec/index.md
```diff
@@ -4,11 +4,383 @@
...
+> KW_PASS           : `pass`\
+> KW_PAYABLE        : `payable`\
+> KW_PUB            : `pub`\
+> KW_RETURN         : `return`\
+> KW_SELFVALUE      : `self`\
+> KW_SELFTYPE       : `Self`\
+> KW_STRUCT         : `struct`\
+> KW_TRUE           : `true`\
+> KW_WHILE          : `while`
+
```
cburgdorf, line 77: @g-r-a-n-t These were the keywords that I anticipate will end up using if we finished up the spec completely. Let me know if you think we should degrade some from `strict` to `reserved` (or drop entirely) or if you think some are missing.

spec/index.md
```diff
@@ -4,11 +4,383 @@
...
+> KW_ELIF           : `elif`\
+> KW_ELSE           : `else`\
+> KW_EMIT           : `emit`\
+> KW_ENUM           : `enum`\
+> KW_EVENT          : `event`\
+> KW_FALSE          : `false`\
+> KW_FOR            : `for`\
+> KW_IF             : `if`\
+> KW_IN             : `in`\
+> KW_LET            : `let`\
```
cburgdorf, line 66: I left this in because I actually think I'm in favor of requiring it. Maybe even making `let` bindings immutable by default and introducing `let mut` for anything mutable. I think that is a good thing for auditability.

spec/index.md
```diff
@@ -55,4 +427,161 @@ Here are examples of these operators being used.
...
+
+### 5.1.1.12. Bytes type
+
+MISSING
+
+### 5.1.1.13. Event types
+
+An *event type* is the type denotated by the name of an [`event` item].
+
+
```
cburgdorf, line 552: These :point_down: are all just links within the markdown

spec/index.md
```diff
@@ -4,11 +4,383 @@
...
+> _EventIndexability_ :\
+> &nbsp;&nbsp; `indexed`<sup>?</sup>
+
+An _event_ is a nominal [event type] defined with the keyword `event`. It is emitted with the keyword `emit`.
+
+An example of a `event` item and its use:
+
+```
+event Transfer:
+    indexed sender: address
```
cburgdorf, line 271: @g-r-a-n-t , I guess this should be `idx` instead, right?

spec/index.md
```diff
@@ -4,11 +4,383 @@
...
+> KW_FALSE          : `false`\
+> KW_FOR            : `for`\
+> KW_IF             : `if`\
+> KW_IN             : `in`\
+> KW_LET            : `let`\
+> KW_NONPAYABLE     : `nonpayable`\
+> KW_PASS           : `pass`\
+> KW_PAYABLE        : `payable`\
+> KW_PUB            : `pub`\
+> KW_RETURN         : `return`\
```
cburgdorf, line 71: Ah, I guess `revert` should be in here, too.

spec/index.md
g-r-a-n-t: I think `Self` should be dropped or reserved. As far as I can tell, this would only be used in the case where we introduce Rust style struct impls, which comes with alot of other key words and complexity. Is there another use case that you have in mind?

cburgdorf: Ah yes, good point. It's a copy pasta leftover.

spec/index.md
```diff
@@ -4,11 +4,383 @@
...
+
+An _event_ is a nominal [event type] defined with the keyword `event`. It is emitted with the keyword `emit`.
+
+An example of a `event` item and its use:
+
+```
+event Transfer:
+    indexed sender: address
+    indexed receiver: address
+    value: uint256
```
g-r-a-n-t, line 273:
```suggestion
    value: u256
```

spec/index.md
g-r-a-n-t: Yeah, this is what's currently used as I recall.

spec/index.md
g-r-a-n-t: yup!

g-r-a-n-t (approved):

PR #54, opened 2020-09-25
Rename uint256 to u256 according to spec update

cburgdorf:
### What was wrong?

We decided to follow rust by renaming `uint256` to `u256` so the tokenizer needed to be changed.

### How was it fixed?

Changed tokenizer + tests to use `u256` instead of `uint256`

### To-Do

- [x] Clean up commit history

### DIFF
--- a/parser/tests/fixtures/tokenizer/validator_registration.v.py.json
+++ b/parser/tests/fixtures/tokenizer/validator_registration.v.py.json
@@ -1,10 +1,10 @@
-MIN_DEPOSIT_AMOUNT: constant(uint256) = 1000000000  # Gwei
-DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
-MAX_DEPOSIT_COUNT: constant(uint256) = 4294967295 # 2**DEPOSIT_CONTRACT_TREE_DEPTH - 1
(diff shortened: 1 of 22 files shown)


parser/tests/fixtures/tokenizer/validator_registration.v.py.json
```diff
@@ -530,85 +530,85 @@ def deposit(pubkey: bytes[PUBKEY_LENGTH],
...
-    "line": "PUBKEY_LENGTH: constant(uint256) = 48  # bytes\n"
+    "line": "PUBKEY_LENGTH: constant(u256) = 48  # bytes\n"
   },
   {
     "typ": "COMMENT",
     "string": "# bytes",
     "start": [
       4,
-      39
+      36
```
cburgdorf, line 592: For the record: I updated this file by hand but by hand. 33 % through I thought to myself: "There's probably an easy way to just regenerate it", 66 % I was almost ready to give up and look into regenerating but then followed through. :sweat_smile:

g-r-a-n-t: :pray:

g-r-a-n-t (approved):
looks like this got very tedious.. thanks for fixing all the spans :)

PR #55, opened 2020-09-28
Map comparision operators

cburgdorf:
### What was wrong?

We do not yet map various comparison expressions (e.g. `==`, `!=` etc)

### How was it fixed?

1. Had to add support for `bool` in various places (I'm being vague because I don't always exactly now yet what I'm doing :joy: )
2. Mapped expressions to their YUL counterparts which for some expressions meant to generate chained calls where the result of one opcode is the input to an `iszero` call
3. Updated the spec

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/abi/builder.rs
+++ b/compiler/src/abi/builder.rs
@@ -156,6 +156,7 @@ fn type_desc<'a>(
 
     match typ {
         fe::TypeDesc::Base { base: "u256" } => Ok(VarType::Uint256),
+        fe::TypeDesc::Base { base: "bool" } => Ok(VarType::Bool),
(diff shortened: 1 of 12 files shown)


cburgdorf:
@g-r-a-n-t if this is making your refactoring in #31  harder then I'm also happy to let this sit and rebase whenever yours landed.

g-r-a-n-t (approved):
looks great!

g-r-a-n-t:
> @g-r-a-n-t if this is making your refactoring in #31 harder then I'm also happy to let this sit and rebase whenever yours landed.

It will only take a minute to update this, so not a problem. I'd rather have this merged.

Issue #62, opened 2020-09-30
Ensure we run CI against Windows too


cburgdorf:
### What is wrong?

After landing https://github.com/ethereum/fe/pull/63 Windows is the only major platform left that we do not yet test against in CI.

### How can it be fixed

1. investigate [the failure](https://github.com/ethereum/fe/runs/1198035048?check_suite_focus=true) in https://github.com/ethereum/fe/pull/64

>No SMT solver found (or it has been forcefully disabled). Optional SMT checking will not be available.  
  Please install Z3 or CVC4 or remove the option disabling them (USE_Z3, USE_CVC4).

The failure :point_up: is weird because both should actually [be disabled](https://github.com/spalladino/solc-rust/blob/4171d70a5d904339e123d28afa637c2df873f14b/build.rs#L13-L14).

cburgdorf:
long happened

PR #63, opened 2020-10-01
Run CI against Mac

cburgdorf:
### What was wrong?

We only test against Linux

### How was it fixed?

Added support for Mac

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,16 +21,14 @@ jobs:
       uses: actions/cache@v1.1.2
       with:
         path: target
-        key: ${{ runner.OS }}-build-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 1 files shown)


cburgdorf:
Ok, finally hitting the point where a cache entry was restored

image

As [explained here](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) cache entries don't work cross PRs. A cache entry created in one PR won't be accessible from another. However, a cache entry created from master will be accessible to all PRs.

cburgdorf:
@g-r-a-n-t I was fighting with Windows support for some hours but eventually gave up and udpated https://github.com/ethereum/fe/issues/62

For now, merging this PR ensures we at least target Linux and Mac.

g-r-a-n-t (approved):

PR #66, opened 2020-10-02
Readme goals updates and build doc.

g-r-a-n-t:
Updated goals with our latest progress and added erc20 as a new short-term goal. Also provided build instructions.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -1,0 +1,23 @@
+# Build and test
+
+Please make sure Rust is [installed](https://www.rust-lang.org/learn/get-started).
+
+**Basic**
+
+The following commands only build the Fe -> Yul compiler components.
+
+- build the CLI: `cargo build`
+- test: `cargo test --workspace`
+
+**Full**
+
+The Fe compiler depends on the Solidty compiler for transforming Yul IR to EVM bytecode. We currently use [solc-rust](https://github.com/axic/solc-rust) to perform this. In order to compile solc-rust, the following must be installed on your system:
+
+- cmake
+- libboost
+- libclang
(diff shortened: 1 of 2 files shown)


cburgdorf:
Ah, I think I know the fix for that CI issue. We need to exclude serde from caching

cburgdorf:
@g-r-a-n-t I believe https://github.com/ethereum/fe/pull/69 will fix the CI issue.

docs/build.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Full**
+
+The Fe compiler depends on the Solidty compiler for transforming Yul IR to EVM bytecode. We currently use [solc-rust](https://github.com/axic/solc-rust) to perform this. In order to compile solc-rust, the following must be installed on your system:
+
+- cmake
+- libboost
+- libclang
+
+Once these have been installed, you can run `cargo build --features solc-backend` to build the full Fe compiler. This may take up to an hour.
```
cburgdorf, line 17: I think this needs to be `cargo build --workspace --features solc-backend` when given from the root.

g-r-a-n-t: I think this part is ok, since we're building the CLI, which depends on the other libraries.

We should probably add another line though that tells them how to test the whole project, which would be:

`cargo test --workspace --features solc-backend`

cburgdorf: Ah, absolutely. I was *thinking*  about running the tests when in fact this is only about building the project :man_facepalming:

cburgdorf (approved):

Issue #59, opened 2020-09-30
Compiling a method without parameters gives parser error


cburgdorf:
### What is wrong?

This should compile but does not:

```
contract Foo:
    pub def bar() -> u256:
        return 0
```

### How can it be fixed

:hocho:

PR #72, opened 2020-10-05
Fix issue with parser not handling functions without arguments

cburgdorf:
### What was wrong?

As explained in #59, the parser does currently error on functions that do not define any arguments such as:

```
contract Foo:
    pub def bar() -> u256:
        return 42
```

### How was it fixed?

1. Unrelated to the issue, but the first commit refactors an existing test into a parameter of another existing test.

2. The `arg_list` function of the parser was augmented to handle the case where the next token is a closing `)`. It then returns an empty vector of `Spanned<FuncDefArg>`
3. Added a test to cover the functionality

### To-Do

- [ ] Clean up commit history


parser/src/parsers.rs
```diff
@@ -569,7 +569,10 @@ pub fn func_def(input: Cursor) -> ParseResult<Spanned<ContractStmt>> {
 }
 
 pub fn arg_list(input: Cursor) -> ParseResult<Vec<Spanned<FuncDefArg>>> {
-    separated(arg_def, op(","), true)(input)
+    match input[0] {
+        Token { string: ")", ..} => Ok((input, vec![])),
```
cburgdorf, line 573: @g-r-a-n-t not sure if that's the intended way to handle this?

parser/src/parsers.rs
```diff
@@ -569,7 +569,10 @@ pub fn func_def(input: Cursor) -> ParseResult<Spanned<ContractStmt>> {
 }
 
 pub fn arg_list(input: Cursor) -> ParseResult<Vec<Spanned<FuncDefArg>>> {
-    separated(arg_def, op(","), true)(input)
+    match input[0] {
+        Token { string: ")", .. } => Ok((input, vec![])),
```
cburgdorf, line 573: @g-r-a-n-t  not sure if this is the intended way of handling it?

cburgdorf:
Ooops, cancelling the review request. Apparently there's still a bug that I missed locally.

cburgdorf:
Ok, should be ready now :see_no_evil:

parser/src/parsers.rs
g-r-a-n-t: If it's the case that `separated` should not be able handle an empty list, I think this is ok. But, it seems odd to me. Maybe this is not the intended behavior of `separated` and we should fix that instead?

I tried changing [this line](https://github.com/ethereum/fe/blob/5b9d302cef06595c6c76486415c6a109e899cb99/parser/src/builders.rs#L278) to

```rust
let (input, first) = match parser(input) {
    Ok(result) => result,
    Err(_) => return Ok((input, vec![]))
};
```

which I thought would allow it to handle empty lists, but it ended breaking a bunch of other stuff, so I'm not quite sure.

cburgdorf: >I tried changing this line to

Wouldn't that swallow legitimate errors by treating them as an empty argument list?

My instinct says it isn't too bad to let `separated`  continue to work on the assumption of dealing with at least one item and handle allowed empty cases explicitly. Another possible idea would be to introduce another parameter `allow_empty` to `separated` to manage this. My intuition says we could leave it as is and refactor it if this pops up more often. What do you think?

g-r-a-n-t: I'm alright with this

g-r-a-n-t (approved):

PR #73, opened 2020-10-05
Type checking for assignments and declarations.

### LINKED ISSUE #68: Type checking for variable declarations and assignments.
The following todo stubs should be replaced with code that checks the validity of the nearby expressions.

- [x] [subscript assignments](https://github.com/ethereum/fe/blob/e56d55597c7afe3779bfb138c7719384995e7790/semantics/src/traversal/assignments.rs#L60)
- [x] [name assignments](https://github.com/ethereum/fe/blob/e56d55597c7afe3779bfb138c7719384995e7790/semantics/src/traversal/assignments.rs#L71)
- [x] [variable declarations](https://github.com/ethereum/fe/blob/e56d55597c7afe3779bfb138c7719384995e7790/semantics/src/traversal/declarations.rs#L28)

g-r-a-n-t:
closes #68 

Adds type checking to assignments and declarations.

### To-Do

- [ ] Clean up commit history


semantics/src/traversal/assignments.rs
```diff
@@ -53,11 +54,15 @@ fn assign_subscript(
...
-        let _index_attributes = expressions::slices_index(scope, context, slices)?;
+        let target_attributes = expressions::expr(Rc::clone(&scope), Rc::clone(&context), target)?;
+        let value_attributes = expressions::expr(Rc::clone(&scope), Rc::clone(&context), value)?;
+        let index_attributes = expressions::slices_index(scope, context, slices)?;
 
-        // TODO: perform type checking
+        let indexed_type = operations::index(target_attributes.typ, index_attributes.typ)?;
+
+        if indexed_type != value_attributes.typ {
+            return Err(SemanticError::TypeError);
```
cburgdorf, line 64: Not for this PR but should the different branches of `SemanticError` allow to carry a string to be able to give the user better feedback about the specific type error?

cburgdorf (approved):
LGTM

semantics/src/traversal/assignments.rs
g-r-a-n-t: Yeah, I'd like to add this at some point. My thinking was to build the semantic pass out a bit more and then find a good approach to handling errors. It would also be nice to include span data and have it so that multiple errors can be returned.

cburgdorf: >it would also be nice to include span data and have it so that multiple errors can be returned.

Yep, that sounds good.

PR #74, opened 2020-10-06
Implement revert

cburgdorf:
### What was wrong?

We need to have support for `revert` expressions.

### How was it fixed?

1. Refactored `test_function` into two functions in order to re-use it for revert assertion
2. Mapped  `revert` to `revert(0,0)` in YUL (Ignoring revert reasons #75 for now)
3. Added a contract test for revert
4. Added coverage for `revert` in the spec

### To-Do

- [x] **Update [Spec](https://github.com/ethereum/fe/blob/master/spec/index.md)**
- [x] Clean up commit history

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -374,7 +374,24 @@ contract GuestBook:
 
 ### 4.1. Statements
 
-MISSING
+### 4.1.1 `revert` statement
+
+
+> **<sup>Syntax</sup>**\
+> _RevertStatement_ :\
+> &nbsp;&nbsp; `revert`<sup>?</sup>
+
+The revert statement is denoted with the keyword `revert`. Evaluating a `revert`
+statement will cause to revert all state changes made by the call and return with an revert error to the caller.
+
+An example of a `revert` statement:
+
+```
+def transfer(to : address, value : u256):
+    if not self.in_whitelist(to):
+        revert
+    # more logic here
+```
 
 ### 4.2 Expressions
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -431,9 +431,7 @@ Here are examples of these operators being used.
 212 >> 1 == 106
 ```
 
-## 4.2.2 Comparision Operators
-
-## Comparison Operators
+### 4.2.2 Comparision Operators
 
 > **<sup>Syntax</sup>**\
 > _ComparisonExpression_ :\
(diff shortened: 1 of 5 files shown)


spec/index.md
```diff
@@ -465,6 +463,29 @@ Here are examples of the comparison operators being used.
 11 <= 11
 ```
 
+### 4.2.4 `revert` expression
```
g-r-a-n-t, line 466: I think we should refer to `revert` as a statement

spec/index.md
```diff
@@ -465,6 +463,29 @@ Here are examples of the comparison operators being used.
...
+
+
+> **<sup>Syntax</sup>**\
+> _RevertExpression_ :\
+> &nbsp;&nbsp; `revert`<sup>?</sup>
+
+The `revert` expression is a way to trigger exceptionsfunction is another way to trigger exceptions from within other code blocks to flag an error and revert the current call. The function takes an optional string message containing details about the error that is passed back to the caller.
```
g-r-a-n-t, line 473:
```suggestion
The `revert` statement is a way to trigger exceptions from within other code blocks to flag an error and revert the current call. The function takes an optional string message containing details about the error that is passed back to the caller.
```

spec/index.md
```diff
@@ -465,6 +463,29 @@ Here are examples of the comparison operators being used.
...
+The revert expressions is denoted with the keyword `revert`. Evaluating a `revert`
+expression will cause to revert all state changes made by the call and return with an revert error to the caller.
+
+An example of a `revert` expression:
+
+```
+def transfer(to : address, value : u256):
+    if not self.in_whitelist(to):
+        revert
+  # more logic here
```
g-r-a-n-t, line 484:
```suggestion
    # more logic here
```

g-r-a-n-t (approved):
Code looks good. Just a few things in the spec that should be improved, mainly that we're referring to `revert` as an expression when it is considered to be a [statement](https://github.com/ethereum/fe/blob/fc0657f6482c7d6d0a4b8f88761c263ea545c250/parser/src/grammar/Vyper.grammar#L74).

spec/index.md
cburgdorf: Ah, you are absolutely right. Will change.

Issue #67, opened 2020-10-02
Determine required features for ERC20 contract support


g-r-a-n-t:
Compilation of an ERC20 token contract is the next big goal. To start, we should complete two things:

- [x] Write an ERC20 token contract in Fe. It should be similar to the OpenZeppelin ERC20 [contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol).
- [ ] Create a list of language features that need to be implemented to support compilation of the contract.

PR #76, opened 2020-10-06
ERC20 contract fixture.

g-r-a-n-t:
addresses #67 

I'll list the needed language features here and then create issues for each one.

### To-Do

- [ ] Create issues for the features mentioned below
- [x] Create a test that parses the contract
- [x] Clean up commit history


g-r-a-n-t:
**create issues for the following features:**
- [ ] constructors
 - we should be able initialize storage values with a constructor
- [ ] string type
 - create a new *string* type
 - support the use of strings in storage
 - support the use of strings in memory
 - support string abi encoding 
- [ ] u8 type
 - create a new *u8* base type
 - support the use of u8 in storage
 - support the use of u8 in memory
 - support u8 abi encoding 
- [ ] u256 type
 - support the use of u256 in storage
- [ ] tuple type
 - create a new *Tuple* type
 - support the use of tuples in storage
 - support the use of tuples in memory
 - support tuple abi encoding
- [ ] assert statement support
- [ ] support for events with any set of parameters

compiler/tests/fixtures/erc20_token.fe
```diff
@@ -0,0 +1,94 @@
...
+        emit Transfer(sender, recipient, amount)
+
+    def _mint(account: address, amount: u256):
+        assert account != address(0)
+
+        self._beforeTokenTransfer(address(0), account, amount)
+
+        self._totalSupply = self._totalSupply + amount
+        self._balances[account] = self._balances[account] + amount
+        emit Transfer(address(0), account, amount)
```
cburgdorf, line 72: Looks like `Transfer`  and `Approval` should be defined anywhere in this file.

g-r-a-n-t: good catch :)

g-r-a-n-t:
@cburgdorf let me know if the contract and feature list generally look good. I'm going to spend some time working on #77 now

cburgdorf (approved):
LGTM

PR #78, opened 2020-10-09
Implement call expressions

cburgdorf:
### What was wrong?

We currently do not support call expressions (e.g. `return self.foo()`)

### How was it fixed?

- mapped the `expr_call` in the mapper
- in the semantic pass, check if the called function does have a return value and raise a `TypeError` if not
- Added a fixture contract to test the success case
- Started a new category of fixture tests that check for compile errors and added a fixture for the described error case

### To-Do

- [x] Clean up commit history
- [x] Create issue for call expressions with arguments
- [x] Created issue for initial sweep of function definitions

### DIFF
--- a/semantics/src/traversal/expressions.rs
+++ b/semantics/src/traversal/expressions.rs
@@ -42,7 +42,7 @@ pub fn expr(
         fe::Expr::BinOperation { .. } => expr_bin_operation(scope, Rc::clone(&context), exp)?,
         fe::Expr::UnaryOperation { .. } => unimplemented!(),
         fe::Expr::CompOperation { .. } => expr_comp_operation(scope, Rc::clone(&context), exp)?,
-        fe::Expr::Call { .. } => unimplemented!(),
+        fe::Expr::Call { .. } => expr_call(scope, exp)?,
         fe::Expr::List { .. } => unimplemented!(),
         fe::Expr::ListComp { .. } => unimplemented!(),
         fe::Expr::Tuple { .. } => unimplemented!(),
@@ -251,6 +251,34 @@ fn expr_bin_operation(
             typ: Type::Base(Base::U256),
             location: Location::Value,
         });
+    }
+
+    unreachable!()
+}
+
+fn expr_call(
+    scope: Shared<FunctionScope>,
+    exp: &Spanned<fe::Expr>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    if let fe::Expr::Call { args: _, func } = &exp.node {
+        if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let contract_scope = &scope.borrow().parent;
+            let called_func = contract_scope.borrow().def(attr.node.to_string());
+            return match called_func {
+                Some(ContractDef::Function {
+                    params: _,
+                    returns: Some(return_type),
+                }) => Ok(ExpressionAttributes {
+                    typ: return_type.into_type(),
+                    location: Location::Value,
+                }),
+                Some(ContractDef::Function {
+                    params: _,
+                    returns: None,
+                }) => Err(SemanticError::NotAnExpression),
+                _ => unreachable!(),
+            };
+        }
     }
 
     unreachable!()
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -166,7 +166,7 @@ fn func_return(
     stmt: &Spanned<fe::FuncStmt>,
 ) -> Result<(), SemanticError> {
     if let fe::FuncStmt::Return { value: Some(value) } = &stmt.node {
-        let _attributes = expressions::expr(scope, context, value);
+        let _attributes = expressions::expr(scope, context, value)?;
 
         // TODO: Perform type checking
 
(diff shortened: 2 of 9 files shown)


semantics/src/traversal/expressions.rs
```diff
@@ -256,6 +256,31 @@ fn expr_bin_operation(
...
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    exp: &Spanned<fe::Expr>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    if let fe::Expr::Call { func, args } = &exp.node {
+
+        if let fe::Expr::Attribute { value:_, attr } = &func.node {
+            let contract_scope = &scope.borrow().parent;
+            let called_func = contract_scope.borrow().def(attr.node.to_string());
+            panic!("{:?}, {:?}, {}", contract_scope.borrow().defs, called_func, attr.node.to_string());
```
cburgdorf, line 269: @g-r-a-n-t If I understood you correctly, then given the following file:

```
contract Foo:
    pub def bar() -> u256:
        return self.foo()

    pub def foo() -> u256:
        return 42
``` 

We should be performing a check on the definition of `foo` here to throw an error in case it turns out it doesn't have an return value. Because in that case the expression `return self.foo()` wouldn't be legal.

I'm trying to do that so I'm going up to the contract_scope with the intention to lookup the definition of `foo`. However, to my surprise it doesn't exist here.

The result of the panic is:

>`thread 'test_method_return::case_01' panicked at '{"bar": Function { params: [], returns: Some(Base(U256)) }}, None, foo', semantics/src/traversal/expressions.rs:269:13
`

So, there is a definition for `bar` but not for `foo`.

Am I doing it wrong?

g-r-a-n-t: At the moment, function definitions are [analyzed one at a time](https://github.com/ethereum/fe/blob/de034e05af021904be516ebd93cf9ed63640b1d5/semantics/src/traversal/contracts.rs#L39) from top to bottom. So, since `foo` is defined after `bar`, we can't see `foo` in `bar`s function body.

We should modify the [contract analysis](https://github.com/ethereum/fe/blob/de034e05af021904be516ebd93cf9ed63640b1d5/semantics/src/traversal/contracts.rs#L29) function to instead do an initial sweep on all function definitions that adds their signatures to the scope.

If you'd like, we can just swap the order of the functions in this contract and create an issue for function signature analysis.

semantics/src/traversal/expressions.rs
```diff
@@ -256,6 +256,34 @@ fn expr_bin_operation(
...
+                    params: _,
+                    returns: Some(_),
+                }) => Ok(ExpressionAttributes {
+                    typ: Type::Base(Base::U256),
+                    location: Location::Value,
+                }),
+                Some(ContractDef::Function {
+                    params: _,
+                    returns: None,
+                }) => Err(SemanticError::TypeError),
```
cburgdorf, line 278: I'm not exactly sure if `TypeError` is the right choice here. If we had a `void` type, I'd say it is but since Fe doesn't have a concept of returning `void` if it doesn't explicitly returns something, maybe a different error type needs to be introduced. What do you think @g-r-a-n-t ?

g-r-a-n-t: `NotAnExpression` might be more fitting. 

In general, feel free to create new error types on a whim. I'm thinking it would be better to perform a refactor of errors once the library is built out more fully than to try to get each error right in every PR.

Eventually it would be nice to print error like Rust does:

```
return self.foo()
       ----------
           ^
           error: not an expression

hint: consider giving `foo` a return value.
```

So we'll have to designate some time to focus solely on improving errors at some point.

cburgdorf: Alright cool! Will update with `NotAnExpression` :+1:

semantics/src/traversal/functions.rs
```diff
@@ -166,7 +166,7 @@ fn func_return(
     stmt: &Spanned<fe::FuncStmt>,
 ) -> Result<(), SemanticError> {
     if let fe::FuncStmt::Return { value: Some(value) } = &stmt.node {
-        let _attributes = expressions::expr(scope, context, value);
+        let _attributes = expressions::expr(scope, context, value)?;
```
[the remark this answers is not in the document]
cburgdorf: I was scratching my head why I don't catch the error in my test when a panic proved the code path was getting hit :sweat_smile:

semantics/src/traversal/expressions.rs
```diff
@@ -256,6 +256,34 @@ fn expr_bin_operation(
...
+    if let fe::Expr::Call { args: _, func } = &exp.node {
+        if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let contract_scope = &scope.borrow().parent;
+            let called_func = contract_scope.borrow().def(attr.node.to_string());
+            return match called_func {
+                Some(ContractDef::Function {
+                    params: _,
+                    returns: Some(_),
+                }) => Ok(ExpressionAttributes {
+                    typ: Type::Base(Base::U256),
```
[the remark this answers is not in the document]
cburgdorf: Ah, nice. Updated.

semantics/src/traversal/functions.rs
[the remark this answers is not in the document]
g-r-a-n-t: yeah, let me double check and make sure there aren't any other instances like this

g-r-a-n-t (approved):

PR #79, opened 2020-10-12
Support AST and Tokens as output targets

cburgdorf:
### What was wrong?

There's currently no way to instruct the compiler to show the internal representation of Tokens or AST for a given code snippet. Being able to emit these can be beneficial for learning purposes and debugging.

### How was it fixed?

- Unrelated but included in this PR: Map `-e`, `-o` to `--emit` and `--output-dir`
- add support for `ast` and `tokens` target
- internal changed the yul compiler from only emitting the bare yul str to emitting a `CompilerOutput`

```rust
pub struct CompilerOutput {
    pub tokens: String,
    pub ast: String,
    pub yul: String,
}
```

### DIFF
--- a/compiler/src/yul/mod.rs
+++ b/compiler/src/yul/mod.rs
@@ -5,8 +5,14 @@ use crate::errors::CompileError;
 mod mappers;
 mod runtime;
 
+pub struct CompilerOutput {
+    pub tokens: String,
+    pub ast: String,
+    pub yul: String,
+}
+
 /// Compiles Fe to Yul.
-pub fn compile(src: &str) -> Result<String, CompileError> {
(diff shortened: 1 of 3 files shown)


compiler/src/yul/mod.rs
```diff
@@ -16,7 +22,11 @@ pub fn compile(src: &str) -> Result<String, CompileError> {
         .values()
         .next()
     {
-        return Ok(first_contract.to_string());
+        return Ok(CompilerOutput {
+            tokens: format!("{:?}", tokens),
+            ast: format!("{:?}", fe_module),
```
cburgdorf, line 27: The formatting could be nicer but I think I'm just going to create an issue for that for now.

g-r-a-n-t (approved):

Issue #19, opened 2020-09-10
Better defined type properties with traits


g-r-a-n-t:
The module that defines the behavior of [types](https://github.com/ethereum/rust-vyper/blob/master/compiler/src/yul/namespace/types.rs) should be tightened up a bit. Instead of generating Yul statements on the types themselves, the types should have various properties attributed to them with traits. The traits would then be used by the compiler to generate various Yul operations. For example, if a given type does not implement the sized trait, it is impossible to perform data operations with it. If it does implement the sized trait, we can use its size value to generate Yul data operations.

The two types of operations that need be refactored are:

- [x] ABI encoding/decoding (#99 via `AbiEncoding` trait)
- [x] data operations (#93 via the `FeSized` trait)

g-r-a-n-t:
Another

g-r-a-n-t:
accidentally closed

PR #81, opened 2020-10-12
Nested maps

### LINKED ISSUE #77: Nested maps
We should be able to support maps of the form: `map<address, map<address, u256>>`

My thinking is that the storage location for the final value (`u256` in this case) should be found like so:

```
self.foo[0x00][0x01]
|------|
    0.) storage index
|------------|
       1.) intermediary storage index
|------------------|
         2.)  storage location of value
```

The expression attributes for each of the above would be as follows:
0.) `type: Map<...>, location: Storage { index: n }` 
1.) `type: Map<...>, location: Storage { index: 0 }` with the *real* index being determined during runtime `keccak256(n, 0x00)`
2.) `type: u256, location: Value`  

*Note: This is a bit strange in that we are using what would normally be a pointer to a segment in storage (`self.foo[0x00]`) as a new storage index. I don't see a problem with this though.*

g-r-a-n-t:
### What was wrong?

closes #77 

We weren't able to compile maps with nested map values.

### How was it fixed?

- [x] Update the definition of maps to include maps as a value type
- [x] Add semantic analysis and mapping for map value types
- [x] Add testing
- [x] Update [Spec](https://github.com/ethereum/fe/blob/master/spec/index.md)
- [x] Clean up commit history

There's a file I added named [operations](https://github.com/ethereum/fe/blob/fed15791844da5717a4c1c5bf4f14feeeb71a2d1/compiler/src/yul/mappers/operations.rs) that should be used to form basic Yul operations. I'll follow up this PR with another that addresses #19 (the issue is a bit outdated) and moves the Yul code generation taking place in the semantic library to the operations module. This is mentioned briefly in #39:

> There are still some things that need to be refactored, but for the sake of keeping this PR at a reasonable size, they were ignored. The best example of this is that we still use methods defined within our types module to form Yul expressions and statements. These Yul elements should instead be formed exclusively in the mapper pass.

### DIFF
--- a/compiler/src/yul/runtime/functions.rs
+++ b/compiler/src/yul/runtime/functions.rs
@@ -24,6 +24,7 @@ pub fn avail() -> yul::Statement {
     function_definition! {
         function avail() -> ptr {
             (ptr := mload(0x00))
+            (if (eq(ptr, 0x00)) { (ptr := 0x20) })
         }
     }
 }
--- a/semantics/src/traversal/expressions.rs
+++ b/semantics/src/traversal/expressions.rs
@@ -5,13 +5,10 @@ use crate::namespace::scopes::{
     FunctionScope,
     Shared,
 };
-use crate::namespace::types::{
-    Array,
-    Map,
-};
+
+use crate::namespace::operations;
 use crate::namespace::types::{
     Base,
-    FixedSize,
     Type,
 };
 use crate::{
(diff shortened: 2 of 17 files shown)


compiler/src/yul/runtime/functions.rs
```diff
@@ -24,6 +24,7 @@ pub fn avail() -> yul::Statement {
     function_definition! {
         function avail() -> ptr {
             (ptr := mload(0x00))
+            (if (eq(ptr, 0x00)) { (ptr := 0x20) })
```
g-r-a-n-t, line 27: There was a bug here where `avail()` was returning `0x00` as the lowest available memory slot, which is reserved for storing the lowest available memory slot, causing us to set the lowest available memory slot to a very large number and triggering an out-of-gas error.

semantics/src/traversal/expressions.rs
```diff
@@ -144,52 +141,24 @@ fn expr_subscript(
...
-                    }),
-                location: Location::Memory,
-            } => ExpressionAttributes {
-                typ: Type::Base(inner),
-                location: Location::Value,
-            },
-            _ => unimplemented!(),
+        let index_attributes = slices_index(scope, context, slices)?;
+
+        let typ = operations::index(value_attributes.typ.clone(), index_attributes.typ)?;
```
g-r-a-n-t, line 146: type checking happens here

g-r-a-n-t:
I made a few adjustments to the type section of the spec @cburgdorf. Let me know if you disagree.

spec/index.md
```diff
@@ -496,36 +496,32 @@ limited capabilities.
...
+* Reference types:
     * [Tuple]
     * [Array]
     * [Bytes]
+    * [Map]
 * User-defined types:
     * [Struct]
     * [Enum]
-    * [Contract]
-    * [Event]
```
cburgdorf, line 510: What's the reason for deleting `Contract` and `Event`?

spec/index.md
```diff
@@ -496,36 +496,32 @@ limited capabilities.
...
+Example:
+
 ```rust
 x = true
 ```
 
-### 5.1.1.2. Contract type
-
-An *contract type* is the type denotated by the name of an [`contract` item].
-
```
cburgdorf, line 527: Same here, it means that e.g. `GuestBook` would be a type (assuming it is defined as contract)

https://github.com/ethereum/fe/blob/7cd0b98a31d745f8d2c399553cab5f6f49827be8/compiler/tests/fixtures/guest_book.fe#L3-L15

cburgdorf (approved):
Looks good. Just wondering about the removal of `Event` and `Contract` from the spec.

spec/index.md
g-r-a-n-t: When I removed these, I was thinking that we should reserve the word *type* for data types. However, I'm now thinking this is too limited.

It would probably be a good idea to come up with a solid definition of what a type in Fe is before we classify things as types. So maybe the correct definition would be:

> A type is anything that can be instantiated.

So under that definition, a contract would be a type if we implement contract loading during runtime. Like this:

```
foo = FooContract.at(contract_address)
bar = foo.some_method()
```

So `FooContract.at(contract_address)` would be an expression that returns an instance of the contract type that gets stored it in `foo`.

With events, I'm not quite sure there's a situation where one would be instantiated, so I'm inclined to leave that out of the types spec.

Maybe this is a better classification:

- Data Types
  - Base types:
    - [Boolean] — `true` or `false`
    - [Address] - Ethereum address
    - [Numeric] — integer
  - Reference types:
    - [Tuple]
    - [Array]
    - [Bytes]
    - [Map]
    - [Struct]
    - [Enum]
- Special types:
  - [Function]
  - [Contract]

g-r-a-n-t: I suppose the current definition is ok.

I'm still not sure if events fits into it though. To me, event definitions exist to simplify writing log statements. You can't really store instances of them anywhere. Also, are you ok with classifying base and reference types under data types?

I agree that contracts should be considered a type.

cburgdorf: >I'm still not sure if events fits into it though. To me, event definitions exist to simplify writing log statements. You can't really store instances of them anywhere.

That's true but I'm not sure if that should count as a requirement. You could certainly have a shared library holding common event types. I think everything that could appear in a type annotation can be considered a type.

>Also, are you ok with classifying base and reference types under data types?

Yes, I think that classification makes sense :+1:

Issue #89, opened 2020-10-13
Support for call expressions with parameters


cburgdorf:
### What is wrong?

After https://github.com/ethereum/fe/pull/78 lands, Fe will have support for call expressions (e.g. `return self.foo()`). However it does not handle yet call expressions with arguments such as `return self.foo(42)`. We need to implement that.

### How can it be fixed

Probably drilling deeper into what https://github.com/ethereum/fe/pull/78 does :sweat_smile:

g-r-a-n-t:
I'm going close #88 and move this under the erc20 milestone.

We should also probably create a separate issue for call statements.

PR #91, opened 2020-10-13
Support for arguments in call expressions.

cburgdorf:
### What was wrong?

As mentioned in #89 we do currently not have support for arguments in call expressions.

### How was it fixed?

- Resolving the arguments with `expr`
   - for that, I had to manually span them reusing the span that is around the `CallArg`
- manually built a `yul::Expression::FunctionCall` as I'm not sure how to do that using a macro.
- Added a test that demonstrates the various ways of using arguments in call expressions

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -29,7 +29,7 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
         fe::Expr::BinOperation { .. } => expr_bin_operation(context, exp),
         fe::Expr::UnaryOperation { .. } => unimplemented!(),
         fe::Expr::CompOperation { .. } => expr_comp_operation(context, exp),
-        fe::Expr::Call { .. } => expr_call(exp),
+        fe::Expr::Call { .. } => expr_call(context, exp),
         fe::Expr::List { .. } => unimplemented!(),
         fe::Expr::ListComp { .. } => unimplemented!(),
         fe::Expr::Tuple { .. } => unimplemented!(),
@@ -38,11 +38,34 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
     }
 }
 
-pub fn expr_call(exp: &Spanned<fe::Expr>) -> Result<yul::Expression, CompileError> {
-    if let fe::Expr::Call { args: _, func } = &exp.node {
+pub fn call_arg(
+    context: &Context,
+    arg: &Spanned<fe::CallArg>,
+) -> Result<yul::Expression, CompileError> {
+    match &arg.node {
+        fe::CallArg::Arg(value) => {
+            let spanned = spanned_expression(&arg.span, value);
+            expr(context, &spanned)
+        }
+        fe::CallArg::Kwarg(fe::Kwarg { name: _, value }) => expr(context, value),
+    }
+}
+
+pub fn expr_call(
+    context: &Context,
+    exp: &Spanned<fe::Expr>,
+) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::Call { args, func } = &exp.node {
         if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let arguments = &args.node;
+            let yul_args: Vec<yul::Expression> = arguments
+                .iter()
+                .map(|val| call_arg(context, val))
+                .collect::<Result<_, _>>()?;
+
             let func_name = identifier! { (attr.node) };
-            return Ok(expression! { [func_name]() });
+
+            return Ok(expression! { [func_name]([yul_args...]) });
         }
     }
 
--- a/compiler/tests/fixtures/compile_errors/return_call_to_fn_with_param_type_mismatch.fe
+++ b/compiler/tests/fixtures/compile_errors/return_call_to_fn_with_param_type_mismatch.fe
@@ -1,0 +1,7 @@
+contract Foo:
+
+    pub def foo(val: address) -> u256:
+        return 42
+
+    pub def bar() -> u256:
+        return self.foo(100)
--- a/compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
+++ b/compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
@@ -1,0 +1,11 @@
+contract Foo:
+    baz: map<u256, u256>
+    pub def foo(val1: u256, val2: u256, val3: u256, val4: u256, val5: u256) -> u256:
+        return val1 + val2 + val3 + val4 + val5
+
+    pub def cem() -> u256:
+        return 100
+
+    pub def bar() -> u256:
+        self.baz[0] = 43
+        return self.foo(5, 2, self.cem(), 25 + 25, self.baz[0])
--- a/semantics/src/traversal/_utils.rs
+++ b/semantics/src/traversal/_utils.rs
@@ -1,0 +1,15 @@
+use fe_parser::ast as fe;
+use fe_parser::span::{
+  Span,
+  Spanned,
+};
+
+
+/// Creates a new spanned expression. Useful in cases where an `Expr` is nested
+/// within the node of a `Spanned` object.
+pub fn spanned_expression<'a>(span: &Span, exp: &fe::Expr<'a>) -> Spanned<fe::Expr<'a>> {
+  Spanned {
+      node: (*exp).clone(),
+      span: (*span).to_owned(),
+  }
+}
--- a/semantics/src/traversal/_utils.rs
+++ b/semantics/src/traversal/_utils.rs
@@ -1,15 +1,33 @@
+use crate::namespace::types::FixedSize;
+use crate::{
+    ExpressionAttributes,
+    Type,
+};
 use fe_parser::ast as fe;
 use fe_parser::span::{
-  Span,
-  Spanned,
+    Span,
+    Spanned,
 };
-
 
 /// Creates a new spanned expression. Useful in cases where an `Expr` is nested
 /// within the node of a `Spanned` object.
 pub fn spanned_expression<'a>(span: &Span, exp: &fe::Expr<'a>) -> Spanned<fe::Expr<'a>> {
-  Spanned {
-      node: (*exp).clone(),
-      span: (*span).to_owned(),
-  }
+    Spanned {
+        node: (*exp).clone(),
+        span: (*span).to_owned(),
+    }
+}
+
+pub fn expression_attributes_to_types(attributes: Vec<ExpressionAttributes>) -> Vec<Type> {
+    attributes
+        .iter()
+        .map(|attributes| attributes.typ.clone())
+        .collect()
+}
+
+pub fn fixed_sizes_to_types(sizes: Vec<FixedSize>) -> Vec<Type> {
+    sizes
+        .iter()
+        .map(|param| param.clone().into_type())
+        .collect()
 }
--- a/compiler/src/yul/mappers/functions.rs
+++ b/compiler/src/yul/mappers/functions.rs
@@ -1,5 +1,5 @@
 use crate::errors::CompileError;
-use crate::yul::mappers::expressions::spanned_expression;
+use crate::yul::mappers::expressions::call_arg;
 use crate::yul::mappers::{
     assignments,
     declarations,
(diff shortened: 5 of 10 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -38,6 +38,28 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
...
+
+pub fn expr_call(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::Call { args, func } = &exp.node {
+        if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let arguments = &args.node;
+            if let fe::CallArg::Arg(fe::Expr::Num(num)) = arguments[0].node {
+                // let yul_arg = expr(context, val)?;
+                let func_arg = literal_expression! {(num)};
+                let func_name = identifier! { (attr.node) };
+                return Ok(expression! { [func_name]([func_arg])});
```
cburgdorf, line 50: @g-r-a-n-t  I'm working on a follow up to support arguments to call expressions but I'm wondering how to dynamically build yul expressions for a dynamic number of arguments (as opposed to the hardcoded one for a single argument seen above)

Do we already have an example for something like that in the code base?

cburgdorf: Ok, figured out how to do that. I bypassed the macros for now. Will change it back if there's a macro way to do that :)

compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
```diff
@@ -0,0 +1,10 @@
+contract Foo:
+
+    pub def foo(val1: u256, val2: u256) -> u256:
+        return val1 + val2
+
+    pub def jip() -> u256:
+        return 44
+
+    pub def bar() -> u256:
```
pipermerriam, line 9: I'm curious about the decision to **not** include `self` in the call signature.  Here are the arguments I see for both sides.

Including it is a little confusing because it means that you have to strip it out to compute the function signature.  One example of this is the 4byte.directory which uses regexes to parse solidity source files and extract function signatures.

Excluding it makes the `self` variable kind of *magic*.  Admittedly, it is already a little magic in python to begin with, but I have always liked that it isn't any different from other variables in terms of scope.  It's possible also that the `ContractClass.some_method(instance, *args)` call semantics might be useful for something like making library calls.

```python
class MyContract:
    def foo(self):
        return MyLibrary.some_library_function(self)
```

cburgdorf: >Excluding it makes the self variable kind of magic. Admittedly, it is already a little magic in python to begin with, but I have always liked that it isn't any different from other variables in terms of scope

Yes, I'm on the same page. I'm actually in favor of requiring `self` for instance methods (so, without `self` means its a class method by definition). I discussed that with @g-r-a-n-t very briefly recently and I think the main argument is just simplicity.

I would be very open (and actually in favor) to make that change.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -38,11 +38,33 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
...
         if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let arguments = &args.node;
+            let yul_args = arguments
+                .iter()
+                .map(|val| {
+                    if let fe::CallArg::Arg(arg_expr) = val.node.clone() {
+                        let spanned = Spanned {
+                            node: arg_expr,
+                            span: val.span,
+                        };
```
cburgdorf, line 55: Not sure if this is pointing to a bigger problem but I had to wrap this in a span, reusing the span around the `CallArg`.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -38,11 +38,33 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
...
+                })
+                .filter_map(Result::ok);
+
             let func_name = identifier! { (attr.node) };
-            return Ok(expression! { [func_name]() });
+
+            return Ok(yul::Expression::FunctionCall(yul::FunctionCall {
+                identifier: func_name,
+                arguments: yul_args.collect(),
+            }));
```
cburgdorf, line 67: @g-r-a-n-t I bypassed the macro here. There's probably another way to do that?

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: see: 

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/semantics/src/traversal/expressions.rs#L84

This helper function exists in both the compiler and mapper libraries. I think we'll just need to accept using it in these situations.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -38,11 +38,33 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
...
+                    if let fe::CallArg::Arg(arg_expr) = val.node.clone() {
+                        let spanned = Spanned {
+                            node: arg_expr,
+                            span: val.span,
+                        };
+                        return expr(context, &spanned);
+                    }
+                    unimplemented!()
+                })
+                .filter_map(Result::ok);
```
g-r-a-n-t, line 60: we should just collect here

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: yup! `expression! { [func_name]([yul_args...]) }` should do the trick

compiler/src/yul/mappers/expressions.rs
```diff
@@ -38,11 +38,33 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
...
+    context: &Context,
+    exp: &Spanned<fe::Expr>,
+) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::Call { args, func } = &exp.node {
         if let fe::Expr::Attribute { value: _, attr } = &func.node {
+            let arguments = &args.node;
+            let yul_args = arguments
+                .iter()
+                .map(|val| {
+                    if let fe::CallArg::Arg(arg_expr) = val.node.clone() {
```
g-r-a-n-t, line 51: noting that we have a function that pretty much does the same thing here. it maps a call arg to an expression. I dont think we want to call back into the function mapper, so it might be worth just copying that into here for the time being. later on we'll want to add support for named parameters, so we'll probably create a separate *arguments* module that can deal with this

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/semantics/src/traversal/functions.rs#L143

g-r-a-n-t: ah, linked to the semantic analysis version. here's the mapper `call_arg` function

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/compiler/src/yul/mappers/functions.rs#L138

compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
g-r-a-n-t: I'm open to this change.

One of the things Christoph and have also discussed is adding explicit mutability similar to rust. I think this could work well with a self parameter.

```python
contract Foo:
  pub def bar(self): # can't modify state (equivalent to Solidity's `view` modifier)
    pass

  pub def baz(mut self): # can modify state
    pass

def foobar(): # can't view or modify state (equivalent to Solidity's `pure`modifier). could also be defined in a contract scope, but is defined in the module scope here and can be shared between contracts in this module and outside of it.
  pass
```

pipermerriam: This has me thinking about re-entrancy protection...

g-r-a-n-t:
Looks good so far @cburgdorf, here are my general thoughts:

We still need to perform semantic analysis on on the call args. If I understand correctly, the tests are currently passing because none of the parameters being passed in utilize `context`. If we were to introduce a arg like say `self.my_map[0]`, compilation would fail because the expression has not been analyzed and we're not quite sure what the type of `self.my_map` is.

I think the best way for us to go forward would be to create a couple new modules for handing arguments. One in the semantics library under traversals and the other in the compiler module under mappers. We should copy the existing `call_arg` functions ([1](https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/semantics/src/traversal/functions.rs#L143) [2](https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/compiler/src/yul/mappers/functions.rs#L138)) used for handling event emit arguments into these modules and then use them for analyzing and mapping arguments for both events and functions calls.

It would also be nice to add type checking for call arguments. This would entail getting the parameter types assigned to a function name from the contract scope and verifying that they indeed match the types of the call args we get from performing semantic analysis on them.

compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
cburgdorf: I created an issue for it https://github.com/ethereum/fe/issues/96

compiler/src/yul/mappers/expressions.rs
cburgdorf: Ah, nice. Maybe we can adopt the snakecharmer Python convention that would go like this:

1. Create a new module `_utils` in `mappers`
2. `_utils` is only meant to be imported from anywhere within `mappers` (which is what the `_` is hinting at)

So, we do not cross import from `expressions` and `functions` but we also don't copy the snippet around.

cburgdorf: Ah, ok..maybe not a good candidate for `_utils` because the it would need to import from `expressions` which defeats the purpose. But maybe the whole `call_arg` helper should be *moved* into `expressions` because `functions` imports from `expressions` anyway but not the other way around.

compiler/src/yul/mappers/expressions.rs
cburgdorf: Good point. I also think I was suppressing any errors with the `filter_map` and the `.collect::<Result<_, _>>()?;` throws them as it should be.

compiler/src/yul/mappers/expressions.rs
cburgdorf: Beautiful :sweat_smile:

cburgdorf:
>If we were to introduce a arg like say `self.my_map[0]`, compilation would fail because the expression has not been analyzed

Yes, it fails and hits the `unreachable` in `expr_subscript`

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/compiler/src/yul/mappers/expressions.rs#L188

And that is because the expression can't be found here in the `context` (as you said. I'm just trying to follow and confirm)

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/compiler/src/yul/mappers/expressions.rs#L175

I think I can figure that out :)

compiler/tests/fixtures/return_u256_from_called_fn_with_args.fe
```diff
@@ -0,0 +1,11 @@
...
+    baz: map<u256, u256>
+    pub def foo(val1: u256, val2: u256, val3: u256, val4: u256, val5: u256) -> u256:
+        return val1 + val2 + val3 + val4 + val5
+
+    pub def cem() -> u256:
+        return 100
+
+    pub def bar() -> u256:
+        self.baz[0] = 43
+        return self.foo(5, 2, self.cem(), 25 + 25, self.baz[0])
```
cburgdorf, line 11: I put `self.baz[0]` in here to cover the case that didn't work so far.

compiler/tests/fixtures/compile_errors/return_call_to_fn_with_param_type_mismatch.fe
```diff
@@ -0,0 +1,7 @@
+contract Foo:
+
+    pub def foo(val: address) -> u256:
```
cburgdorf, line 3: This test proves the type checking for the arguments works as expected (trying to pass `100` where an `address` is expected)

semantics/src/traversal/_utils.rs
```diff
@@ -0,0 +1,33 @@
...
+        .map(|attributes| attributes.typ.clone())
+        .collect()
+}
+
+pub fn fixed_sizes_to_types(sizes: Vec<FixedSize>) -> Vec<Type> {
+    sizes
+        .iter()
+        .map(|param| param.clone().into_type())
+        .collect()
+}
```
cburgdorf, line 33: I'm not super happy with these two helper functions. I initially wanted to write an `impl` for `Vec<ExpressionAttributes>` but that isn't legal as I can't write an impl for a foreign type. There might be a more idiomatic way to handle that.

g-r-a-n-t: We should be able to implement the relevant `From` traits and then use something like: `attributes.iter().collect<Type>()`, but this is ok as is.

PR #93, opened 2020-10-14
FeSized trait and data opertions.

g-r-a-n-t:
### What was wrong?

Partially addresses #19 and something that was mentioned in #39.

> There are still some things that need to be refactored, but for the sake of keeping this PR at a reasonable size, they were ignored. The best example of this is that we still use methods defined within our types module to form Yul expressions and statements. These Yul elements should instead be formed exclusively in the mapper pass. 

### How was it fixed?

Created a new [`FeSized` trait](https://github.com/ethereum/fe/blob/f67359d68d63f4a8885a22a067654143d4174d1b/semantics/src/namespace/types.rs#L8) that is analogous to Rust's very own [`Sized` trait](https://doc.rust-lang.org/beta/std/marker/trait.Sized.html), then used this trait to implement a set of [data operations](https://github.com/ethereum/fe/blob/80486998c0fdc41fd31eb223ea2b72ed0b24cb59/compiler/src/yul/mappers/operations.rs) and refactored the compiler to use these operations.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/assignments.rs
+++ b/compiler/src/yul/mappers/assignments.rs
@@ -1,8 +1,14 @@
 use crate::errors::CompileError;
 use crate::yul::mappers::expressions;
+use crate::yul::mappers::operations;
 use fe_parser::ast as fe;
 use fe_parser::span::Spanned;
-use fe_semantics::namespace::types::Type;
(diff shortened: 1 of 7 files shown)


cburgdorf (approved):
LGTM

PR #94, opened 2020-10-14
Create ownable.fe

Ro5s:
adapt ownable.fe from oz ownable.sol implement: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol

### DIFF
--- a/compiler/tests/fixtures/ownable.fe
+++ b/compiler/tests/fixtures/ownable.fe
@@ -12,12 +12,12 @@ contract Ownable:
     return self._owner
   
   pub def renounceOwnership():
-    assert msg.sender == owner
-    _owner = address(0)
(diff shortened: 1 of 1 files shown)


g-r-a-n-t:
Hi @Ro5s, thanks for this! I'd definitely like to add this to our collection of fixtures.

One thing that would be nice is to have a test that makes sure this parses. We are already doing this with the erc20 contract [here](https://github.com/ethereum/fe/blob/master/compiler/tests/erc20.rs) and I'm thinking we could repurposed this file to check that all fixture files parse, instead of just erc20. That way we can perform some level of validation on PRs like this. Do you have any interest in doing this? If not, we can go ahead and merge this and I'll do this later.

PR #97, opened 2020-10-16
Data layout spec.

g-r-a-n-t:
### What was wrong?

Data layout was not specified.

### How was it fixed?

Described how each data type should be stored in each location.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -210,7 +210,7 @@ def answer_to_life_the_universe_and_everything() -> u256:
 
 These two terms are often used interchangeably, and what they are attempting to convey is the answer to the question "Can this item be used at this location?"
 
-Fe knows two different types of visibility for functions and state variables: `public` and `private`. Visibility of `private` is the default and is used if no other visiblity is specified. 
+Fe knows two different types of visibility for functions and state variables: `public` and `private`. Visibility of `private` is the default and is used if no other visiblity is specified.
 
 **Public:** External functions are part of the contract interface, which means they can be called from other contracts and via transactions.
 
@@ -502,11 +502,12 @@ The list of types is:
         * [Address] - Ethereum address
         * [Numeric] — integer
     * Reference types:
-        * [Tuple]
-        * [Array]
-        * [Bytes]
-        * [Struct]
-        * [Enum]
+        * Sequence types
+            * [Tuple]
+            * [Array]
+            * [Bytes]
+            * [Struct]
+            * [Enum]
         * [HashMap]
 * Other types:
     * [Event]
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -598,8 +599,102 @@ MISSING
...
+The value of a base type in storage is found by simply loading the value from storage at the
+given pointer.
+
+To find a value inside of a sequence type. The relative location of the value is added to the
+given pointer.
+
+#### 6.2.2. Maps in storage
+
+Maps are not assigned pointers, because they do not have a location in storage. They are instead 
+assigned a unique index value that is used to derive the location of keyed values.
```
pipermerriam, line 660: What do you think about the term `nonce` instead of `index`?

https://www.merriam-webster.com/dictionary/nonce

It denotes *single use* but without the notion of ordering or *indexing* into a sequence.

g-r-a-n-t: I was actually trying to think of a word other than index. Nonce makes alot of sense.

spec/index.md
```diff
@@ -598,8 +599,102 @@ MISSING
...
-An *event type* is the type denotated by the name of an [`event` item].
+An *event type* is the type denoted by the name of an [`event` item].
 
+## 6. Data Layout
+
+There are three places where data can be stored on the EVM:
+
+- stack
+- storage
+- memory
```
cburgdorf, line 610: Should we count `calldata` as well?

g-r-a-n-t: I initially had it in there, but took it out since there's not a whole lot to say about it. The data will always be laid out according to the contract ABI spec.

spec/index.md
```diff
@@ -598,8 +599,129 @@ MISSING
...
+
+To find an element inside of a sequence type, the relative location of the element is added to the
+given pointer.
+
+### 6.4. Function calls
+
+Constant size values stored on the stack or in memory can be passed into and returned by functions.
+
+Memory is owned by a function if it is allocated in the function call or in the function's body.
+When a function exits, all memory owned by the function is freed.
```
cburgdorf, line 702: I like that we free the memory on function exit. I noticed that Solidity [does not do that](https://solidity.readthedocs.io/en/v0.7.4/internals/layout_in_memory.html):

>memory is never freed (this might change in the future).

I'd love to talk about this some more in our next weekly call.

cburgdorf (approved):

spec/index.md
g-r-a-n-t: I removed the stuff about memory ownership and de-allocation, since gas-savings are questionable and it complicates compiler implementation.

PR #100, opened 2020-10-23
Support for events with multiple params.

### LINKED ISSUE #87: Full event support
We are currently able to emit a `Signed` event in the guestbook contract, but not much more. For ERC-20, we should be able to emit events of any parameter set or at least just any parameter set of base types.

- [ ] add multi-parameter support for events

g-r-a-n-t:
### What was wrong?

closes #87 

Events had very limited support. Basically, you could only emit events with a single `bytes` parameter.

### How was it fixed?

Added ABI encoding for any combination of values and hooked this into event emits.

When an event definition is encountered, an encoding operation is created for the given set of parameters. This encoding operation is then used whenever an event is emitted. The same thing is done with public functions and their external returns.

Here are the encoding operations generated for [events.fe](https://github.com/ethereum/fe/blob/285025b167f0361664f51d0bc53b3cb97c79e11a/compiler/tests/fixtures/events.fe):

```
function abi_encode_uint256_address_uint256_bytes100(val_0, val_1, val_2, val_3) -> ptr
{
    ptr := avail()
    pop(alloc_mstoren(val_0, 32))
    pop(alloc_mstoren(val_1, 32))
    pop(alloc_mstoren(val_2, 32))
    for { let i := 0 } lt(i, 100) { i := add(i, 1) }
    {
        let val_ptr := add(val_3, mul(i, 1))
        let val := mloadn(val_ptr, 1)
        pop(alloc_mstoren(val, 1))
    }
}
function abi_encode_uint256_address(val_0, val_1) -> ptr
{
    ptr := avail()
    pop(alloc_mstoren(val_0, 32))
    pop(alloc_mstoren(val_1, 32))
}
function abi_encode_uint256_uint256(val_0, val_1) -> ptr
{
    ptr := avail()
    pop(alloc_mstoren(val_0, 32))
    pop(alloc_mstoren(val_1, 32))
}
function abi_encode_address2(val_0) -> ptr
{
    ptr := avail()
    for { let i := 0 } lt(i, 2) { i := add(i, 1) }
    {
        let val_ptr := add(val_0, mul(i, 20))
        let val := mloadn(val_ptr, 20)
        pop(alloc_mstoren(val, 32))
    }
}
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/operations.rs
+++ b/compiler/src/yul/operations.rs
@@ -1,3 +1,4 @@
+use crate::yul::runtime::functions;
 use fe_semantics::namespace::events::Event;
 use fe_semantics::namespace::types::{
     AbiEncoding,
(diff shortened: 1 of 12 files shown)


g-r-a-n-t:
To illustrate, these are the encoding functions generated for the events.fe contract:

```
            function abi_encode_uint256_uint256(val_0, val_1) -> ptr                                                                                                                                                                           
            {                                                                                                                                                                                                                                  
                ptr := avail()                                                                                                                                                                                                                 
                pop(alloc_mstoren(val_0, 32))                                                                                                                                                                                                  
                pop(alloc_mstoren(val_1, 32))                                                                                                                                                                                                  
            }                                                                                                                                                                                                                                  
            function abi_encode_uint256_address(val_0, val_1) -> ptr                                                                                                                                                                           
            {                                                                                                                                                                                                                                  
                ptr := avail()                                                                                                                                                                                                                 
                pop(alloc_mstoren(val_0, 32))                                                                                                                                                                                                  
                pop(alloc_mstoren(val_1, 32))                                                                                                                                                                                                  
            }                                                                                                                                                                                                                                  
            function abi_encode_uint256_address_uint256_bytes100(val_0, val_1, val_2, val_3) -> ptr                                                                                                                                            
            {                                                                                                                                                                                                                                  
                ptr := avail()                                                                                                                                                                                                                 
                pop(alloc_mstoren(val_0, 32))                                                                                                                                                                                                  
                pop(alloc_mstoren(val_1, 32))                                                                                                                                                                                                  
                pop(alloc_mstoren(val_2, 32))                                                                                                                                                                                                  
                for { let i := 0 } lt(i, 100) { i := add(i, 1) }                                                                                                                                                                               
                {                                                                                                                                                                                                                              
                    let val_ptr := add(val_3, mul(i, 1))                                                                                                                                                                                       
                    let val := mloadn(val_ptr, 1)                                                                                                                                                                                              
                    pop(alloc_mstoren(val, 1))                                                                                                                                                                                                 
                }                                                                                                                                                                                                                              
            }
```

compiler/src/yul/operations.rs
```diff
@@ -111,3 +94,38 @@ pub fn decode(typ: FixedSize, ptr: yul::Expression) -> yul::Expression {
...
+        Base,
+        FixedSize,
+    };
+    use yultsur::*;
+
+    #[test]
+    fn test_encode() {
+        assert_eq!(
+            encode(vec![FixedSize::Base(Base::U256)], vec![expression! { 42 }]).to_string(),
+            "abi_encode_uint256(42)"
```
cburgdorf, line 115: I guess this is another case where we will need to ensure generated method names can not clash with user defined method names. Maybe just generating them with a leading `$` would be sufficient for now? I'm going to create an issue for that.

cburgdorf (approved):
Great work! It seems to me that it would be a relatively small change to enforce that all generated functions begin with some character that is illegal in Fe but legal in YUL such as `$`. Maybe worth putting in here as a general protection until we have something more sophisticated?

compiler/src/yul/operations.rs
cburgdorf: I created an issue for this https://github.com/ethereum/fe/issues/102

PR #101, opened 2020-10-26
Add support for returning boolean literal

cburgdorf:
### What was wrong?

Fe does currently not support returning boolean literals. The following won't parse:

```
contract Foo:
    pub def bar() -> bool:
        return false
```

### How was it fixed?

1. Added a new expression type `Bool`
2. Changed the parser to map the literals `true`  and `false` to the `Bool` expression
3. Added handling for the new expression in semantic and mapper pass

### To-Do

- [x] Clean up commit history

### DIFF
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -236,6 +236,7 @@ pub enum Expr<'a> {
     Tuple {
         elts: Vec<Spanned<Expr<'a>>>,
     },
+    Bool(bool),
     Name(&'a str),
     Num(&'a str),
     Str(Vec<&'a str>),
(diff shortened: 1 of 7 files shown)


parser/src/ast.rs
```diff
@@ -236,6 +236,7 @@ pub enum Expr<'a> {
     Tuple {
         elts: Vec<Spanned<Expr<'a>>>,
     },
+    Bool(bool),
```
cburgdorf, line 239: I'm unsure if this should be of type `&str` and then be mapped to `Expr::Bool("false")` and `Expr::Bool("true")`  instead. The current system works and we can map using a `literal_expression` but I *guess* that's just by coincidence because rust and Fe both use the same literals. If we would choose `yes` and `no`, then I think we would have to use strings.

g-r-a-n-t: I'm fine with this the way it is.

g-r-a-n-t (approved):

PR #103, opened 2020-10-27
Implement call statements

### LINKED ISSUE #92: Function call statement support
We should be able to compile function call statements.

e.g.: `self.foo(42, 26, my_var)`

This will be very similar to what has been implemented for #78 and #89. We can just discard the return value (if there is one).

cburgdorf:
### What was wrong?

closes #92 

As stated in  #92, we do currently not have support for call statements (e.g. `self.assign(10, 11)`. 

### How was it fixed?

This is based on #111

- Moved `spanned_expression` helper into _utils module
- Mapped `fe::FuncStmt::Expr` to either the expression directly or to `pop([expr])` in case the return type is anything else than `()`
- In the semantic pass, I just analyzed the `fe::FuncStmt::Expr` as any other expression.
- A few helpers to ease checking for `()` type

### DIFF
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -141,12 +141,25 @@ fn func_stmt(
         fe::FuncStmt::While { .. } => unimplemented!(),
         fe::FuncStmt::If { .. } => unimplemented!(),
         fe::FuncStmt::Assert { .. } => unimplemented!(),
-        fe::FuncStmt::Expr { .. } => unimplemented!(),
+        fe::FuncStmt::Expr { .. } => expr(scope, context, stmt),
         fe::FuncStmt::Pass => unimplemented!(),
         fe::FuncStmt::Break => unimplemented!(),
         fe::FuncStmt::Continue => unimplemented!(),
         fe::FuncStmt::Revert => Ok(()),
     }
+}
+
+fn expr(
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::Expr { value } = &stmt.node {
+        let spanned = spanned_expression(&stmt.span, value);
+        let _attributes = expressions::expr(scope, context, &spanned)?;
+    }
+
+    Ok(())
 }
 
 fn emit(
--- a/compiler/src/yul/mappers/_utils.rs
+++ b/compiler/src/yul/mappers/_utils.rs
@@ -1,0 +1,14 @@
+use fe_parser::ast as fe;
+use fe_parser::span::{
+    Span,
+    Spanned,
+};
+
+/// Creates a new spanned expression. Useful in cases where an `Expr` is nested
+/// within the node of a `Spanned` object.
+pub fn spanned_expression<'a>(span: &Span, exp: &fe::Expr<'a>) -> Spanned<fe::Expr<'a>> {
+    Spanned {
+        node: (*exp).clone(),
+        span: (*span).to_owned(),
+    }
+}
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -1,10 +1,8 @@
 use crate::errors::CompileError;
+use crate::yul::mappers::_utils::spanned_expression;
 use crate::yul::operations;
 use fe_parser::ast as fe;
-use fe_parser::span::{
-    Span,
-    Spanned,
-};
+use fe_parser::span::Spanned;
 use fe_semantics::namespace::types::{
     Array,
     FeSized,
@@ -153,15 +151,6 @@ pub fn slices_index(
     }
 
     unreachable!()
-}
-
-/// Creates a new spanned expression. Useful in cases where an `Expr` is nested
-/// within the node of a `Spanned` object.
-pub fn spanned_expression<'a>(span: &Span, exp: &fe::Expr<'a>) -> Spanned<fe::Expr<'a>> {
-    Spanned {
-        node: (*exp).clone(),
-        span: (*span).to_owned(),
-    }
 }
 
 pub fn slice_index(
--- a/compiler/src/yul/mappers/functions.rs
+++ b/compiler/src/yul/mappers/functions.rs
@@ -1,4 +1,5 @@
 use crate::errors::CompileError;
+use crate::yul::mappers::_utils::spanned_expression;
 use crate::yul::mappers::expressions::call_arg;
 use crate::yul::mappers::{
     assignments,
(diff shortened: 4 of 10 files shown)


semantics/src/traversal/functions.rs
```diff
@@ -106,14 +106,28 @@ fn func_stmt(
...
 }
 
+fn expr(
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::Expr { value } = &stmt.node {
+        let spanned = spanned_expression(&stmt.span, value);
+        // TODO: Whats the reason the expr isn't spanned?
```
g-r-a-n-t, line 124: If a child node fits entirely within the parent, we don't give them a new span. I guess this would prevent us from having duplicate spans, but I'm not sure if this is necessary.

semantics/src/traversal/functions.rs
```diff
@@ -106,14 +106,28 @@ fn func_stmt(
         fe::FuncStmt::While { .. } => unimplemented!(),
         fe::FuncStmt::If { .. } => unimplemented!(),
         fe::FuncStmt::Assert { .. } => unimplemented!(),
-        fe::FuncStmt::Expr { .. } => unimplemented!(),
+        fe::FuncStmt::Expr { .. } => expr(scope, context, stmt),
```
g-r-a-n-t, line 109: I could be wrong, but I think this variant would only be used for function call statements.

If I were working on this, I'd probably start by directing this straight into a method named something like `func_stmt_call` that is pretty similar to the [`expr_call`](https://github.com/ethereum/fe/blob/de661d0e8d81d006b09319413b9fab5cff93c956/semantics/src/traversal/expressions.rs#L247) method. The only difference would be that we don't return a semantic error for empty return calls and that we don't create expression attributes for the call statement.

semantics/src/traversal/functions.rs
```diff
@@ -106,14 +106,28 @@ fn func_stmt(
...
 
+fn expr(
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::Expr { value } = &stmt.node {
+        let spanned = spanned_expression(&stmt.span, value);
+        // TODO: Whats the reason the expr isn't spanned?
+        let _attributes = expressions::expr(scope, context, &spanned)?;
```
g-r-a-n-t, line 125: If we were to compile a function call statement that does not return a value, we would get a `NotAnExpression` error, since the `expr` method expects something that has some value.

I'm personally in favor of just treating call statements and call expressions as two separate node types and handling them in distinct branches of code, but arguments could be made in favor introducing a `void` type for situations like this.

cburgdorf: >If we were to compile a function call statement that does not return a value, we would get a NotAnExpression error,

Yeah, that makes a lot of sense. I didn't run into it because I accidentally gave `assign` a return value which the semantic pass did not complain about. I filed a separate issue for that https://github.com/ethereum/fe/issues/104

>I'm personally in favor of just treating call statements and call expressions as two separate node types and handling them in distinct branches of code

Yeah, I remember that but when I noticed that function call statements are hitting `fe::FuncStmt::Expr` I was scratching my head what else I might need to handle here. I guess I was expecting something like `fe::FuncStmt::Call` to exist.

>but arguments could be made in favor introducing a void type for situations like this

In general, I think that expression based programming is very powerful and I consider it one of the beauties of Rust to be so expression focused that pretty much everything (if/else, loops etc) is an expression. Expressions often tighten the rules and hence eliminate sources of bugs.

E.g. this simple if/else won't catch the typo in the assignment.
```
if something:
   some_var = "foo"
else:
   zome_var = "bar"
```

The same thing can not happen if a ternary expression was used.

```
some_var = "foo" if something else "bar"
```

What would be your main argument against introducing a `void` type?

cburgdorf: I started to address #104 in a separate PR https://github.com/ethereum/fe/pull/105/

g-r-a-n-t: > What would be your main argument against introducing a void type?

I don't really have a strong argument against it and the more I think about it, the more I think we should add it. My initial hesitation just stemmed from adding another type variant, which we would then need to handle throughout the codebase. 

What do you think about using an empty tuple as our void/none type?

cburgdorf: >the more I think about it, the more I think we should add it

Cool :+1: 

>What do you think about using an empty tuple as our void/none type?

Yes, I think we should do that. I wasn't fully aware that the `()` type rust uses [is in fact just an empty tuple](https://stackoverflow.com/a/31107736/288703). That seems to make things more streamlined and avoids adding a special type to the type system (assuming that we want to have tuples anyway)

semantics/src/traversal/functions.rs
cburgdorf: Wouldn't we also hit this for any other expression such as `1+1` in the following example.

```
    pub def bar() -> u256:
        1 + 1
        return 100
```

Obviously, the `1 + 1` doesn't really do anything and an optimizer would probably optimize it away but I assume it is still valid.

semantics/src/traversal/functions.rs
cburgdorf: I don't fully understand what you mean by that

> fits entirely within the parent

What is the difference in terms of spans for these two snippets:

```
def foo():
    return 1
```
or

```
def foo():
    self.bar()
```

What makes `self.bar()` *fit entirely within the parent* but `return 1` doesn't?

semantics/src/traversal/functions.rs
```diff
@@ -141,14 +141,27 @@ fn func_stmt(
...
 }
 
+fn expr(
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::Expr { value } = &stmt.node {
+        let spanned = spanned_expression(&stmt.span, value);
+        let _attributes = expressions::expr(scope, context, &spanned)?;
```
cburgdorf, line 159: Is there anything we should type check here? I'm under the impression that `expressions::expr(scope, context, &spanned)?;` does all the heavy lifting already, right?

cburgdorf:
@g-r-a-n-t I rebased this on #111 and think its ready to review.

semantics/src/traversal/functions.rs
g-r-a-n-t: In the case of a return statement, the expression that is nested within it (`1`) does not occupy the entire span of `return 1`, so it gets its own span. The span for `return 1` would be something like `(17, 24)`, while the span of `1` would be `(23, 24)`

In the case of the expression statement, the expression occupies the entire span of the parent node (`self.bar()` is `self.bar()`), which will always be true for expression statements,  so we don't bother assigning the same span to the expression.

semantics/src/traversal/functions.rs
g-r-a-n-t: I'm ok with allowing this sort of thing now.

semantics/src/traversal/functions.rs
cburgdorf: Oh! Now I understand. Thanks for taking the time to explain it in that detailed way :+1:

semantics/src/traversal/functions.rs
g-r-a-n-t: Nothing to check, these attributes get thrown away :+1:

g-r-a-n-t (approved):

PR #105, opened 2020-10-28
Validate return values of function body

### LINKED ISSUE #104: Throw an error if a function does not return what the signature says it does
### What is wrong?

Currently, code such as this compiles without error:

```
contract Foo:
    baz: map<u256, u256>

    pub def assign() -> u256:
        self.baz[0] = 100
```

Notice that assign is defined to return `u256` but doesn't return anything.

### How can it be fixed

Throw an error if:

1. A method is defined with return value but does not return any value
2. A method is defined with a return value but the actual returned value doesn't match the signature
3. A method is defined to **not** return a value but does return a value

cburgdorf:
Fixes #104

### What was wrong?

As explained in #104 we do currently not score well in terms of return value checking. We do not detect when:

1  A method is defined with return value but does not return any value
2. A method is defined with a return value but the actual returned value doesn't match the signature
3. A method is defined to **not** return a value but does return a value

### How was it fixed?

For case 1 and 3, I added checks directly from `func_def` to scan if all code paths return or revert (or the opposite)
For case 2 I gave `func_return` access to the `FunctionAttributes` of the host function to perform type checking

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/compiler/tests/compile_errors.rs
+++ b/compiler/tests/compile_errors.rs
@@ -8,6 +8,9 @@ use std::fs;
 #[rstest(
     fixture_file,
     error,
+    case("mismatch_return_type.fe", "[Str(\"semantic error: TypeError\")]"),
+    case("unexpected_return.fe", "[Str(\"semantic error: UnexpectedReturn\")]"),
+    case("missing_return.fe", "[Str(\"semantic error: MissingReturn\")]"),
     case(
         "return_call_to_fn_without_return.fe",
         "[Str(\"semantic error: NotAnExpression\")]"
--- a/semantics/src/lib.rs
+++ b/semantics/src/lib.rs
@@ -106,8 +106,8 @@ impl Context {
     }
 
     /// Get information that has been attributed to an expression node.
-    pub fn get_expression(&self, spanned: &Spanned<fe::Expr>) -> Option<&ExpressionAttributes> {
-        self.expressions.get(&spanned.span)
+    pub fn get_expression<T: Into<Span>>(&self, span: T) -> Option<&ExpressionAttributes> {
+        self.expressions.get(&span.into())
     }
 
     /// Attribute contextual information to an emit statement node.
@@ -116,8 +116,8 @@ impl Context {
     }
 
     /// Get information that has been attributed to an emit statement node.
-    pub fn get_emit(&self, spanned: &Spanned<fe::FuncStmt>) -> Option<&Event> {
-        self.emits.get(&spanned.span)
+    pub fn get_emit<T: Into<Span>>(&self, span: T) -> Option<&Event> {
+        self.emits.get(&span.into())
     }
 
     /// Attribute contextual information to a function definition node.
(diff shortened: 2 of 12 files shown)


compiler/tests/compile_errors.rs
```diff
@@ -8,6 +8,9 @@ use std::fs;
 #[rstest(
     fixture_file,
     error,
+    case("mismatch_return_type.fe", "[Str(\"semantic error: TypeError\")]"),
```
cburgdorf, line 11: @g-r-a-n-t 

I guess to satisfy this :point_up: test, I would have to add type checking here:

https://github.com/ethereum/fe/blob/de661d0e8d81d006b09319413b9fab5cff93c956/semantics/src/traversal/functions.rs#L168-L182

But I'm a bit confused, how do I lookup the function name of the function that hosts the `return`?

cburgdorf: Mmh..one way I can think of is to pass `name` into `func_stmt` and pass it on to each function that needs it.

https://github.com/ethereum/fe/blob/de661d0e8d81d006b09319413b9fab5cff93c956/semantics/src/traversal/functions.rs#L72-L74

Then it can be looked up from the context. Alternatively, instead of the `name` we could pass on the whole `FunctionAttributes` which would eliminate the need to do the lookup. Not sure if that's the right approach but I'm going to try it out to make progress on this.

cburgdorf:
The failing build is due to the general build error with MacOS. I have a PR #107 open to disable MacOS builds (but would prefer another option) But I think this can be reviewed ignoring the build error (but I won't merge it anyway before we haven't resolved the build error)

semantics/src/namespace/scopes.rs
```diff
@@ -52,6 +54,7 @@ pub struct ContractScope {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct FunctionScope {
+    pub span: Span,
```
cburgdorf, line 57: I tried to put the `FunctionAttributes` here but there's a cyclic dependency so the best I could have come up with would have been `Option<FunctionAttributes>` which leads to more verbose code down the road. So, I decided to put the `Span` here because that's all we need to lookup the `FunctionAttributes` on the `Context` (After a small refactoring of the context https://github.com/ethereum/fe/pull/105/commits/1c8a21721127f428dfe26f23b92fb1d8578f7fe3)

semantics/src/lib.rs
```diff
@@ -106,8 +106,8 @@ impl Context {
     }
 
     /// Get information that has been attributed to an expression node.
-    pub fn get_expression(&self, spanned: &Spanned<fe::Expr>) -> Option<&ExpressionAttributes> {
-        self.expressions.get(&spanned.span)
+    pub fn get_expression<T: Into<Span>>(&self, span: T) -> Option<&ExpressionAttributes> {
```
cburgdorf, line 109: These getters previously worked with `Spanned<T>` but given the fact that they only actually care about the `Span` it was a bit unfortunate that one could not perform a lookup with a bare `Span`. So now they accept `Into<Span>` which means all the previous lookups still work as they did before but one can also use a bare `Span` to perform a lookup.

semantics/src/traversal/functions.rs
```diff
@@ -79,6 +83,38 @@ pub fn func_def(
...
+
+        if let fe::FuncStmt::Revert { .. } = &statement.node {
+            return Ok(());
+        }
+    }
+
+    Err(SemanticError::MissingReturn)
+}
+
+fn validate_no_path_returns(body: &[Spanned<fe::FuncStmt>]) -> Result<(), SemanticError> {
```
cburgdorf, line 105: This rule will probably become obsolete when we introduce the implicit return of `()` because I believe such a method will then be perfectly legal:

```
def no_return_in_signature():
    return ()
```

The same thing is legal in rust:

```rust
fn no_return() {
    return ();
}
```

g-r-a-n-t (approved):

PR #107, opened 2020-10-29
Disable MacOS builds on CI

cburgdorf:
### What was wrong?

CI is [currently failing for Mac Builds](https://github.com/ethereum/fe/pull/106/checks?check_run_id=1322145918)

### What I tried

1. Pin to an older rust nightly
2. Pin to an older Mac Version

None of that worked.

### How was it fixed?

Disabled Mac for the lack of a better fix :cry: 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,7 +72,8 @@ jobs:
       matrix:
         include:
         - os: ubuntu-latest
-        - os: macOS-latest
+        # Ignoring Mac until we find a solution to: https://github.com/ethereum/fe/pull/106/checks?check_run_id=1322145918
+        # - os: macOS-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


cburgdorf:
@g-r-a-n-t I tried to investigate the CI error that currently happens on MacOS. I couldn't figure out what is wrong and I have no Mac available to investigate it locally. The only mitigation I have right now would be to disable MacOS builds for now. I don't really like that so maybe you have more ideas for what we could try.

g-r-a-n-t:
I tried a few things. Had one build pass, but then it failed after rebasing :thinking: 

Solidity is due for an update anyway, might be better to try updating solc-rust from solc 0.6 -> 0.7 and seeing if this persists.

Going to merge.

g-r-a-n-t:
see #108

PR #109, opened 2020-10-29
Support for __init__ functions. 

### LINKED ISSUE #83: Implement constructors
It should be possible to write a constructor that accepts arguments and sets storage values. I'm thinking the constructor should look like Python inits, but am open to doing something else.

```Python
def __init__(name: string, symbol: string):
      self._name = name
      self._symbol = symbol
      self._decimals = 18
```

The first code block in the Yul [ERC-20 example](https://solidity.readthedocs.io/en/latest/yul.html#complete-erc20-example) is basically where *construction* will happen. It should look something like this:

```
code {
      function __init__(name, symbol) {
          ...
       }

      // Call the constructor
      __init__(...) // Call init with string params

      // Deploy the contract
      datacopy(0, dataoffset("runtime"), datasize("runtime"))
      return(0, datasize("runtime"))
}
```

*note: strings are not required for this issue. We can just test with different types that are already supported.*

g-r-a-n-t:
### What was wrong?

closes #83 

We did not have basic support for init functions.

### How was it fixed?

Added the following:

- `__init__` functions are added to the json ABI as constructor functions.
- if an `__init__` function is defined, it will be added to the contract constructor and called before the contract is deployed.
- a contract test that validates this behavior.

One limitation right now is that other functions defined by the user can not be called by the init function. This is because user defined functions are not added to the constructor code block. If this is something that we want to support, we could add the user defined functions to the constructor code block so that users can call any function they like during initialization.

### To-Do

- [x] Clean up commit history


compiler/src/yul/constructor.rs
```diff
@@ -0,0 +1,74 @@
...
+    } else {
+        // build a constructor without an init function
+        block! { [deploy_stmts...] }
+    };
+
+    yul::Code { block }
+}
+
+#[test]
+fn test_constructor_w_o_func() {
```
cburgdorf, line 69: Nittiest of nitpicks but I'd prefer to stick to the rule of not using single letter acronyms and call this `test_constructor_without_func`

cburgdorf (approved):
LGTM :tada:

PR #111, opened 2020-11-02
Implicitly return empty tuple when nothing else is returned

cburgdorf:
### What was wrong?

Currently, functions that do not return anything are a special case from the perspective of the type system. There is no way to capture anything from a call to such functions. In fact, a call to a function that does not return anything is treated as a semantic error.

E.g.

```
contract Foo:

def child():
    y = 1

def parent():
    x = self.child()  # semantic error
```

We [recently discussed](https://github.com/ethereum/fe/pull/103#discussion_r514078077) to change this and follow the model of rust which treats any function that does not explicitly return a value as one that returns an empty tuple `()`.

The core motivation for that is that it generalizes the handling of functions by getting rid of different branching logic based on whether a function does or does not return a value. In the new models, functions simply always return, even if only a zero sized empty tuple.

This means, that the code above would be valid and that the type of `x` would be `()` since this is what `child` returns.

These function definitions should be equivalent and the shorter ones can be treated as syntactic sugar to the first.

Full length form:
```
def child() -> ():
    y = 1
    return ()
```

No explicit return in signature:
```
def child():
    y = 1
    return ()
```

No explicit return statement:
```
def child() -> ():
    y = 1
```

No return in signature and function body:
```
def child():
    y = 1
```

**Notice that support for the `()` syntax is out of scope for this PR and is only meant to explain the planned changes that this PR is working towards.** 

### How was it fixed?

1. Introduced a `Tuple` type to the type system
2. Changed `return_type` of `FunctionAttributes` from `Option<FixedSize>` to `FixedSize` which means that function return values cease to be *optional* past the parser stage (They continue to be optional in the parser because of the shorthand function definition syntax)
3. Changed checks that had branching logic for returning / non-returning functions to be strictly focused on the return type

### DIFF
--- a/compiler/src/yul/mappers/functions.rs
+++ b/compiler/src/yul/mappers/functions.rs
@@ -42,39 +42,41 @@ pub fn func_def(
             .collect::<Result<Vec<_>, _>>()?;
 
         // Different return types require slightly different functions.
-        if let Some(return_type) = attributes.return_type.to_owned() {
-            return match return_type {
-                // Base types are returned by value. All memory used by the function is cleared.
-                FixedSize::Base(_) => Ok(function_definition! {
-                    function [function_name]([param_names...]) -> return_val {
-                        (let ptr := avail())
-                        [function_statements...]
-                        (free(ptr))
-                    }
-                }),
-                // Arrays need to keep memory allocated after completing.
-                // FIXME: Copy arrays to the lowest available pointer to save memory.
-                FixedSize::Array(array) => {
-                    let size = literal_expression! {(array.size())};
-
-                    Ok(function_definition! {
-                        function [function_name]([param_names...]) -> return_val {
-                            [function_statements...]
-                            (free((add(return_val, [size]))))
-                        }
-                    })
+        return match attributes.return_type.to_owned() {
+            // Base types are returned by value. All memory used by the function is cleared.
+            FixedSize::Base(_) => Ok(function_definition! {
+                function [function_name]([param_names...]) -> return_val {
+                    (let ptr := avail())
+                    [function_statements...]
+                    (free(ptr))
                 }
-            };
-        }
+            }),
+            // Arrays need to keep memory allocated after completing.
+            FixedSize::Array(array) => {
+                let size = literal_expression! {(array.size())};
 
-        // Nothing is returned. All memory used by the function is freed.
-        return Ok(function_definition! {
-            function [function_name]([param_names...]) {
-                (let ptr := avail())
-                [function_statements...]
-                (free(ptr))
+                Ok(function_definition! {
+                    function [function_name]([param_names...]) -> return_val {
+                        [function_statements...]
+                        (free((add(return_val, [size]))))
+                    }
+                })
             }
-        });
+            FixedSize::Tuple(tuple) => {
+                if tuple.size() == 0 {
+                    // Nothing is returned. All memory used by the function is freed.
+                    return Ok(function_definition! {
+                        function [function_name]([param_names...]) {
+                            (let ptr := avail())
+                            [function_statements...]
+                            (free(ptr))
+                        }
+                    });
+                }
+
+                unimplemented!();
+            }
+        };
     }
 
     unreachable!()
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -48,6 +48,9 @@ pub enum TypeDesc<'a> {
         from: Box<Spanned<TypeDesc<'a>>>,
         to: Box<Spanned<TypeDesc<'a>>>,
     },
+    Tuple {
+        items: Vec<Spanned<TypeDesc<'a>>>,
+    },
 }
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
--- a/semantics/src/lib.rs
+++ b/semantics/src/lib.rs
@@ -67,7 +67,7 @@ impl ExpressionAttributes {
 pub struct FunctionAttributes {
     pub name: String,
     pub param_types: Vec<FixedSize>,
-    pub return_type: Option<FixedSize>,
+    pub return_type: FixedSize,
 }
 
 /// Contains contextual information about a Fe module and can be queried using
--- a/semantics/src/namespace/operations.rs
+++ b/semantics/src/namespace/operations.rs
@@ -13,7 +13,8 @@ pub fn index(value: Type, index: Type) -> Result<Type, SemanticError> {
     match value {
         Type::Array(array) => index_array(array, index),
         Type::Map(map) => index_map(map, index),
-        Type::Base(_) => Err(SemanticError::TypeError),
+        Type::Base(_) => Err(SemanticError::NotSubscriptable),
+        Type::Tuple(_) => Err(SemanticError::NotSubscriptable),
     }
 }
 
--- a/semantics/src/namespace/types.rs
+++ b/semantics/src/namespace/types.rs
@@ -48,12 +48,14 @@ pub enum Type {
     Base(Base),
     Array(Array),
     Map(Map),
+    Tuple(Tuple),
 }
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum FixedSize {
     Base(Base),
     Array(Array),
+    Tuple(Tuple),
 }
 
 #[derive(Clone, Debug, PartialEq)]
@@ -76,11 +78,17 @@ pub struct Map {
     pub value: Box<Type>,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Tuple {
+    pub items: Vec<Base>,
+}
+
 impl FeSized for FixedSize {
     fn size(&self) -> usize {
         match self {
             FixedSize::Base(base) => base.size(),
             FixedSize::Array(array) => array.size(),
+            FixedSize::Tuple(tuple) => tuple.size(),
         }
     }
 }
@@ -90,6 +98,7 @@ impl AbiEncoding for FixedSize {
         match self {
             FixedSize::Array(array) => array.abi_name(),
             FixedSize::Base(base) => base.abi_name(),
+            FixedSize::Tuple(tuple) => tuple.abi_name(),
         }
     }
 
@@ -97,6 +106,7 @@ impl AbiEncoding for FixedSize {
         match self {
             FixedSize::Base(base) => base.abi_size(),
             FixedSize::Array(array) => array.abi_size(),
+            FixedSize::Tuple(tuple) => tuple.abi_size(),
         }
     }
 
@@ -104,6 +114,7 @@ impl AbiEncoding for FixedSize {
         match self {
             FixedSize::Base(base) => base.abi_padding(),
             FixedSize::Array(array) => array.abi_padding(),
+            FixedSize::Tuple(tuple) => tuple.abi_padding(),
         }
     }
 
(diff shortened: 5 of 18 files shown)


cburgdorf:
@g-r-a-n-t Can you help me make sense of the failing tests here?

```
failures:

---- address_bytes10_map stdout ----
thread 'address_bytes10_map' panicked at 'Unable to compile to bytecode: CompileError { errors: [Str("{\"errors\":[{\"component\":\"general\",\"formattedMessage\":\"input.yul:1:2368: TypeError: Expected expression to return one item to the stack, but did return 0 items.\\n ... se 0x281388a3 { return(abi_encode_(write_bar(calldataload(4), ccopy(36, 10))), 0) }  }  } }\\n                                        ^---------------------------------------^\\n\",\"message\":\"Expected expression to return one item to the stack, but did return 0 items.\",\"severity\":\"error\",\"sourceLocation\":{\"end\":2408,\"file\":\"input.yul\",\"start\":2367},\"type\":\"TypeError\"}]}")] }', compiler/tests/evm_contracts.rs:145:47

---- guest_book stdout ----
thread 'guest_book' panicked at 'Unable to compile to bytecode: CompileError { errors: [Str("{\"errors\":[{\"component\":\"general\",\"formattedMessage\":\"input.yul:1:2485: TypeError: Expected expression to return one item to the stack, but did return 0 items.\\n ... se 0x7f8fe731 { return(abi_encode_(sign(ccopy(4, 100))), 0) }  }  } }\\n                                        ^-----------------^\\n\",\"message\":\"Expected expression to return one item to the stack, but did return 0 items.\",\"severity\":\"error\",\"sourceLocation\":{\"end\":2503,\"file\":\"input.yul\",\"start\":2484},\"type\":\"TypeError\"}]}")] }', compiler/tests/evm_contracts.rs:145:47

---- nested_map stdout ----
thread 'nested_map' panicked at 'Unable to compile to bytecode: CompileError { errors: [Str("{\"errors\":[{\"component\":\"general\",\"formattedMessage\":\"input.yul:1:2511: DeclarationError: Function name abi_encode_ already taken in this scope.\\n ... () pop(alloc_mstoren(val_0, 32)) } function abi_encode_(val_0) -> ptr { ptr := avail() pop(alloc_mstoren(val_0, 0)) } switch cloadn(0, 4) case 0xa5ad475 ...\\n                                        ^--------------------------------------------------------------------------------^\\n\",\"message\":\"Function name abi_encode_ already taken in this scope.\",\"severity\":\"error\",\"sourceLocation\":{\"end\":2592,\"file\":\"input.yul\",\"start\":2510},\"type\":\"DeclarationError\"}]}")] }', compiler/tests/evm_contracts.rs:145:47

---- events stdout ----
thread 'events' panicked at 'Unable to compile to bytecode: CompileError { errors: [Str("{\"errors\":[{\"component\":\"general\",\"formattedMessage\":\"input.yul:1:3328: DeclarationError: Function name abi_encode_ already taken in this scope.\\n ... 0) pop(alloc_mstoren(val, 32)) } } function abi_encode_(val_0) -> ptr { ptr := avail() pop(alloc_mstoren(val_0, 0)) } function abi_encode_uint256_addres ...\\n                                        ^--------------------------------------------------------------------------------^\\n\",\"message\":\"Function name abi_encode_ already taken in this scope.\",\"severity\":\"error\",\"sourceLocation\":{\"end\":3409,\"file\":\"input.yul\",\"start\":3327},\"type\":\"DeclarationError\"}]}")] }', compiler/tests/evm_contracts.rs:145:47

---- u256_u256_map stdout ----
thread 'u256_u256_map' panicked at 'Unable to compile to bytecode: CompileError { errors: [Str("{\"errors\":[{\"component\":\"general\",\"formattedMessage\":\"input.yul:1:2258: TypeError: Expected expression to return one item to the stack, but did return 0 items.\\n ... se 0xc6d8ca82 { return(abi_encode_(write_bar(calldataload(4), calldataload(36))), 0) }  }  } }\\n                                        ^------------------------------------------^\\n\",\"message\":\"Expected expression to return one item to the stack, but did return 0 items.\",\"severity\":\"error\",\"sourceLocation\":{\"end\":2301,\"file\":\"input.yul\",\"start\":2257},\"type\":\"TypeError\"}]}")] }', compiler/tests/evm_contracts.rs:145:47


failures:
    address_bytes10_map
    events
    guest_book
    nested_map
    u256_u256_map

test result: FAILED. 44 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out

```

I *think* what's happening is that I generate the wrong abi encoding (I definitely generate it wrong, just not sure if that's the problem). But I'm thinking that in terms of abi encoding a method that now returns an empty tuple should be treated as if it were to not return anything (just like before) so I think there's probably a place somewhere that needs to be adjusted that I'm not seeing. Sorry for the vague error description..if you don't find the time to look at it, don't worry, I'll return to it with a fresh brain tomorrow.

cburgdorf:
Ok, I have found the error by inspecting what has changed in the YUL code. I've put out a PR that lets us generate YUL code even when the generated YUL code would not translate into valid bytecode. This has helped me find this bug https://github.com/ethereum/fe/pull/112/

semantics/src/lib.rs
```diff
@@ -67,7 +67,7 @@ impl ExpressionAttributes {
 pub struct FunctionAttributes {
     pub name: String,
     pub param_types: Vec<FixedSize>,
-    pub return_type: Option<FixedSize>,
+    pub return_type: FixedSize,
```
cburgdorf, line 70: I like how this has streamlined a bunch of code. There's simply no `None` case anymore to handle..

cburgdorf:
@g-r-a-n-t There's still some WIPyness about this code including leftover comments but it feels like approaching MVP for empty tuple support. Would appreciate your input especially about the places that I left comments in.

semantics/src/lib.rs
g-r-a-n-t: yeah, definitely nice not having to guard against these

semantics/src/namespace/operations.rs
```diff
@@ -33,6 +35,14 @@ fn index_map(map: Map, index: Type) -> Result<Type, SemanticError> {
     Ok(*map.value)
 }
 
+fn index_tuple(_tuple: Tuple, index: Type) -> Result<Type, SemanticError> {
+    if index != Type::Base(Base::U256) {
+        return Err(SemanticError::TypeError);
+    }
+    // Not sure how I would get the type here given the inputs..
```
g-r-a-n-t, line 42: I don't think it should be possible to index tuples with a runtime value like we do with maps and arrays. We could consider adding something like `my_tuple.0`, but the index will have to be known at compile time.

semantics/src/namespace/types.rs
```diff
@@ -219,13 +242,48 @@ impl Array {
...
+    fn abi_size(&self) -> usize {
+        self.size()
+    }
+
+    fn abi_padding(&self) -> AbiPadding {
+        unimplemented!();
+    }
+
+    fn abi_type(&self) -> AbiType {
+        AbiType::Terminal
```
g-r-a-n-t, line 275: Currently, we have `Terminal` and `UniformRecursive` ABI types. I think when we get around to adding support for non-empty tuples, we'll want to create a new ABI type named `Recursive` that contains any set of other ABI types.

For now, I think all of these ABI functions should be marked as unimplemented.

semantics/src/traversal/expressions.rs
```diff
@@ -157,10 +157,12 @@ fn expr_subscript(
                     // Index value is ignored. We may want to introduce a new location
                     // variant named StorageRuntime or something to suit this case.
                     Type::Map(_) => Location::Storage { index: 0 },
+                    Type::Tuple(_) => Location::Value,
```
g-r-a-n-t, line 160: should be memory

semantics/src/traversal/expressions.rs
```diff
@@ -157,10 +157,12 @@ fn expr_subscript(
...
                 }
             }
             Type::Array(_) => Location::Value,
             Type::Base(_) => unreachable!(),
+            Type::Tuple(_) => unimplemented!(),
```
g-r-a-n-t, line 165: Again, I don't think it would be possible to index a tuple like the other sequence types, so I'd think this would be a semantic error. Also, I'm wondering why indexing a base type is marked as `unreachable`, I'd think this would throw a semantic error too. I realize I probably wrote this code :D

parser/src/ast.rs
```diff
@@ -100,6 +103,8 @@ pub enum ContractStmt<'a> {
         qual: Option<Spanned<FuncQual>>,
         name: Spanned<&'a str>,
         args: Vec<Spanned<FuncDefArg<'a>>>,
+        // I guess this continues to be an option because we are strictly concerned about the
+        // syntax in which return type or no return type are both valid options?
```
g-r-a-n-t, line 107: Yeah, I think we should leave this as is.

compiler/src/yul/mappers/functions.rs
```diff
@@ -42,39 +42,42 @@ pub fn func_def(
...
+            // Base types are returned by value. All memory used by the function is cleared.
+            FixedSize::Base(_) => Ok(function_definition! {
+                function [function_name]([param_names...]) -> return_val {
+                    (let ptr := avail())
+                    [function_statements...]
+                    (free(ptr))
+                }
+            }),
+            // Arrays need to keep memory allocated after completing.
+            // FIXME: Copy arrays to the lowest available pointer to save memory.
```
g-r-a-n-t, line 55: We might as well delete this comment since we're on the fence about memory de-allocation.

semantics/src/namespace/operations.rs
cburgdorf: Ah, interesting! I didn't know that this is how rust handles it but it makes sense now that I think about it. I'm onboard :+1:

semantics/src/traversal/expressions.rs
cburgdorf: Assuming that each Tuple is always a fixed size set of base types, why wouldn't it be `Location::Value`? I'm not arguing against it, there's probably a good reason for that, that I want to learn :)

semantics/src/traversal/expressions.rs
cburgdorf: Makes sense. I created `SemanticError::NotSubscriptable` and changed both arms.

cburgdorf:
@g-r-a-n-t I made the changes you suggested so maybe this qualifies as MVP.

semantics/src/traversal/expressions.rs
g-r-a-n-t: I'm thinking that a tuple could also contain arrays or other tuples, in addition to base types.

Even if we did just allow base types in tuples, though, I'd think we'd still need to store them in memory, as I'm not sure how we would manage storing multiple values associated with a tuple on the stack.

Say we have something like this:

```
foo: (u256, address, u256) = (42, 0x01, 26)
```

The total size of this value is 84 bytes, so it can not be stored in one stack slot. If we were to commit to storing tuple values on the stack, then we would need to map it to something like this:

```
let foo_0 := 42
let foo_1 := 0x01
let foo_2 := 26
```

This seems pretty messy to deal with... Imagine trying to pass a whole tuple into another function like this :thinking:. Using memory seems alot simpler imo because we just have to deal with one pointer:

```
let foo := alloc(84)
mstoren(foo, 42, 32)
mstoren(add(foo, 32), 0x01, 20)
mstoren(add(foo, 52), 26, 32)
```

Then looking up values in the tuple is pretty similar to looking up values in an array, the only difference being that we must know the index at compile time so that we can perform type analysis.

g-r-a-n-t (approved):

semantics/src/traversal/expressions.rs
cburgdorf: Ah, that makes sense!

PR #112, opened 2020-11-03
Allow compile to YUL even when YUL is invalid

cburgdorf:
### What was wrong?

Sometimes, especially during development, we would like to inspect a piece of generated YUL code even if it is invalid, meaning that we can not derive proper bytecode from it.

Naturally one would think that by using `<path-to-fe-file> --emit yul` instead of `<path-to-fe-file> --emit yul, bytecode` we could simply remove the request for compiling to bytecode to see the faulty YUL code. But that's not how the internals of the compiler work up to now because compilation is currently an all-or-nothing operation. 

### How was it fixed?

1. The compile module got support for a new `CompileStage` enum that controls if we should stop after generating YUL (which currently implies generating tokens and AST) or generate everything up to the bytecode

2. The CLI only uses `CompileStage::AllUpToBytecode` only if the generation of `bytecode` is actually requested by the user.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/evm/mod.rs
+++ b/compiler/src/evm/mod.rs
@@ -10,11 +10,26 @@ pub struct CompilerOutput {
     pub bytecode: String,
 }
 
+pub enum CompileStage {
+    AllUpToYul,
+    AllUpToBytecode,
+}
+
 /// Compiles Fe to bytecode. It uses Yul as an intermediate representation.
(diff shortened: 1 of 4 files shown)


g-r-a-n-t (approved):
Makes sense

PR #113, opened 2020-11-05
Assert statements.

### LINKED ISSUE #86: Support assert statements
Statements of the form `assert <bool expression>` should check if the expression is false and revert the transaction if it is.

- [x] type check the bool expression
- [x] map the assert statement to a Yul if statement that reverts if entered

g-r-a-n-t:
### What was wrong?

closes #86 

We did not support assert statements.

### How was it fixed?

Added support for assert statements. This does not include support for revert messages if the assert fails.

For example, the message in the statement `assert x > 5, "not greater than five"` will just be ignored. We can add this after basic string support is implemented, but it's not necessarily required for the Fe erc20 contract.

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
Nice!

PR #116, opened 2020-11-06
Add support for if / else statements

cburgdorf:
### What was wrong?

Fe does currently not support `if` / `else`  statements.

### How was it fixed?

1. Added support by mapping to the equivalent switch block in YUL
2. Added a check in the semantic pass that the tested expression is of type `Base::Bool`
3. Changed the `validate_all_paths_return_or_revert` check to follow down the branches of `if` / `else` to ensure all paths `return` or `revert`
4. Added a bunch of test cases to cover the functionality.

### DIFF
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -94,18 +94,29 @@ pub fn func_def(
 }
 
 fn validate_all_paths_return_or_revert(
-    body: &[Spanned<fe::FuncStmt>],
+    block: &[Spanned<fe::FuncStmt>],
 ) -> Result<(), SemanticError> {
-    // This will need to become more sophisticated when we introduce branching logic
-    // because we then need to follow different code paths and check that they
-    // all return or revert.
-    for statement in body {
+    for statement in block.iter().rev() {
         if let fe::FuncStmt::Return { .. } = &statement.node {
             return Ok(());
         }
 
         if let fe::FuncStmt::Revert { .. } = &statement.node {
             return Ok(());
+        }
+
+        if let fe::FuncStmt::If {
+            test: _,
+            body,
+            or_else,
+        } = &statement.node
+        {
+            let body_returns = validate_all_paths_return_or_revert(body).is_ok();
+            let or_else_returns =
+                or_else.is_empty() || validate_all_paths_return_or_revert(or_else).is_ok();
+            if body_returns && or_else_returns {
+                return Ok(());
+            }
         }
     }
 
@@ -141,7 +152,7 @@ fn func_stmt(
         fe::FuncStmt::AugAssign { .. } => unimplemented!(),
         fe::FuncStmt::For { .. } => unimplemented!(),
         fe::FuncStmt::While { .. } => unimplemented!(),
-        fe::FuncStmt::If { .. } => unimplemented!(),
+        fe::FuncStmt::If { .. } => if_statement(scope, context, stmt),
         fe::FuncStmt::Assert { .. } => assert(scope, context, stmt),
         fe::FuncStmt::Expr { .. } => expr(scope, context, stmt),
         fe::FuncStmt::Pass => unimplemented!(),
@@ -149,6 +160,26 @@ fn func_stmt(
         fe::FuncStmt::Continue => unimplemented!(),
         fe::FuncStmt::Revert => Ok(()),
     }
+}
+
+fn if_statement(
+    scope: Shared<FunctionScope>,
+    context: Shared<Context>,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::If {
+        test,
+        body: _,
+        or_else: _,
+    } = &stmt.node
+    {
+        let attributes = expressions::expr(scope, context, &test)?;
+        if let Type::Base(Base::Bool) = attributes.typ {
+            return Ok(());
+        }
+    }
+
+    Err(SemanticError::TypeError)
 }
 
 fn expr(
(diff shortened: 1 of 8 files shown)


semantics/src/traversal/functions.rs
```diff
@@ -94,19 +94,34 @@ pub fn func_def(
...
+            or_else,
+        } = &statement.node
+        {
+            let body_returns = validate_all_paths_return_or_revert(body)?;
+            let or_else_returns = if !or_else.is_empty() {
+                validate_all_paths_return_or_revert(or_else)?;
+            };
+            if vec![body_returns, or_else_returns]
+                .iter()
+                .all(|result| result == &())
```
cburgdorf, line 120: Not, super proud of this. There's probably a more rusty way to do this. I thought about something like:

```rust
[
validate_all_paths_return_or_revert(body),
validate_all_paths_return_or_revert(or_else),
].collect::<Something, Something>()?;
```
But I gave up on that since the `or_else` needs to be checked for emptiness.

semantics/src/traversal/functions.rs
```diff
@@ -94,19 +94,34 @@ pub fn func_def(
...
 
 fn validate_all_paths_return_or_revert(
-    body: &[Spanned<fe::FuncStmt>],
+    block: &[Spanned<fe::FuncStmt>],
 ) -> Result<(), SemanticError> {
-    // This will need to become more sophisticated when we introduce branching logic
-    // because we then need to follow different code paths and check that they
-    // all return or revert.
-    for statement in body {
+    for statement in block {
```
g-r-a-n-t, line 99: Looks like this might be better off as a single match statement.

semantics/src/traversal/functions.rs
g-r-a-n-t: If I understand correctly, the following would fail to compile:

```
contract Foo:
    pub def bar(val: u256) -> u256:
        if val > 1:
            return 5
        else:
            x = 1
        
        return 42
```

as ` validate_all_paths_return_or_revert(or_else)?` would return an error before we get to the final statement in `bar`. This could probably be fixed and cleaned up by doing something like this:

```rust
let body_returns = validate_all_paths_return_or_revert(body).is_ok();
let or_else_returns = if !or_else.is_empty() {
  validate_all_paths_return_or_revert(or_else).is_ok();
};

if body_returns && or_else_return {
  // code definitely terminates
  return Ok(());
}
// continue checking statements
```

However, I think calling `validate_all_paths_return_or_revert` on code blocks that don't need to terminate is a bit misleading. It might be better to have separate function named `is_terminal` that takes a list of statements and returns a bool, instead of a result. The base case of this function would be `false`, and any absolutely terminal statements that we encounter would return `true`. We could then wrap this function inside of `validate_all_paths_return_or_revert` where it expects a terminal code block and fails if it does not get one. Does this seem cleaner?

semantics/src/traversal/functions.rs
```diff
@@ -151,6 +166,28 @@ fn func_stmt(
...
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::If {
+        test,
+        body: _,
+        or_else: _,
+    } = &stmt.node
+    {
+        let attributes = expressions::expr(scope, context, &test)?;
+        if let Type::Base(base) = attributes.typ {
```
g-r-a-n-t, line 181: I would suggest:

```
if Type::Base(Base::Bool) == attributes.typ {
  return Ok(());
}
```

semantics/src/traversal/functions.rs
g-r-a-n-t: Something worth noting is that `if` statements without an `or_else` are not guaranteed to terminate, so we don't need to check the `if` code block in that case. We only need to check the contents of an `if` statement if there is an `or_else` attached. In that case, we would check to make sure that both blocks are guaranteed to terminate.

Even in the case of:

```
def bar() -> u256:
  if true:
    return 42
```

We would throw an error since sematic analysis isn't concerned about what values expressions evaluate to.

With something like this:

```
def bar(x: u256) -> u256:
  # non-terminal
  if <case1>:
    return 42

  # terminal
  if <case2>:
    # terminal
    return 26
  else:
    # non-terminal
    if <case3>:
      return 4226

    # terminal
    return 9999

```

We can just think about whether or not a statement is terminal or non-terminal and require that all code branches end with a terminal statement.

cburgdorf: Ah, you are making great points here. The snippet you shared is in fact falsely categorized as *not returning* even though it clearly does.

>is there any reason to iterate over each statement, or can we just recursively grab the last one and make sure it terminates?

That's an interesting point. I think that would certainly make things easier to implement. That said, in Rust (as well as many other languages) such code would be valid and just trigger a warning because of an unreachable statement.

```rust
fn something() -> bool {
    return true;
    let x = 1;
}
```

In our case, it would lead to a hard error because the last statement would not be a `return` or `revert`.

I think I have to play a bit with the code to get to an informed opinion :)

semantics/src/traversal/functions.rs
cburgdorf: Ah, of course. I sometimes still fail to recognize when there are nicer options available with Rust :)

semantics/src/traversal/functions.rs
cburgdorf: The reason I didn't do that is because I would have to add one arm with an empty `{}` which doesn't look very elegant imho. Or did you think of something else?

semantics/src/traversal/functions.rs
cburgdorf: Regarding my previous comment. If we don't want to fixate on the last statement of the block (which would diverge from what other languages are doing), I think the closest do that would be to just check the statements in reverse order and return `Ok(())` as soon as we find the first termination. I added a test case to cover the case that previously failed.

cburgdorf:
@g-r-a-n-t Are we good to merge this?

g-r-a-n-t (approved):

g-r-a-n-t:
LGTM :partying_face:

PR #119, opened 2020-11-10
Basic string support and encoding/decoding of dynamically sized ABI types.

g-r-a-n-t:
### What was wrong?

We do not support strings or ABI encoding of dynamically sized types. 

### How was it fixed?

Added a new string type and made the necessary updates to our ABI encoding components to support dynamically sized types.

What this does add support for:
- strings in memory
- encoding/decoding of dynamically sized arrays

What this does not support:
- strings in storage
- string literals
- any operations on strings

Strings in the ABI are treated like dynamically sized byte arrays. This is somewhat different than the encoding of statically sized values.

Say we want to encode the following statically sized values `(bytes[100], u256)`, the encoding will look roughly like this:

`|------bytes[100]-------| u256 |`

Both of the values are encoded in-place, since we know the size of each value at compile time.

With dynamically sized values, we need to include the size of each array in the encoding. The encoding is split into two sections: the first being where static values are encoded in-place along with the head's of dynamic values, and the second being where the data of the dynamic values is stored.

So say we want to encode `(bytes, u256)`, that would look like this:

`| bytes_head | u256 | bytes_size |-----bytes_data-----|`

`bytes_head` in this case would be the offset value `64`, which points to the start of bytes in the dynamic section.

In order to encode the heads of dynamically sized values, we must know the size of each dynamic array preceding it. Strings and other "dynamically" sized values in Fe will be stored the same way as their statically sized counterparts, with the exception of having a 32-byte size appended to the front. So, to lookup the size of a dynamically sized type, we just need to call `mload(array_ptr)`. 

When we encounter a dynamically sized value during encoding, we store the sum of the static section and current size of the dynamic section, then move on. When we're finished with the static section, we take all of dynamically sized values and encode them one at a time.

Here is what a generated encoding function looks like:

```javascript
function abi_encode_string26_uint256_string42_string100_address(val_0, val_1, val_2, val_3, val_4) -> ptr
{
    // we return the start of the encoding in memory
    ptr := avail()

    // store headers and static values

    // string26 header
    pop(alloc_mstoren(add(160, 0), 32))
    // uint256 in-place encoding
    pop(alloc_mstoren(val_1, 32))
    // string42 header
    pop(alloc_mstoren(add(160, add(32, ceil32(mul(mload(val_0), 1)))), 32))
    // string100 header (these header statements grow because we need to sum the sized of the other dynamically sized arrays)
    pop(alloc_mstoren(add(160, add(add(32, ceil32(mul(mload(val_0), 1))), add(32, ceil32(mul(mload(val_2), 1))))), 32))
    // address in-place encoding
    pop(alloc_mstoren(val_4, 32))
    
    // store the dynamic arrays

    // store string26
    {
        pop(alloc_mstoren(mload(val_0), 32))
        {
            let array_size := mload(val_0)
            let array_data_size := mul(array_size, 1)
            for { let i := 0 } lt(i, array_size) { i := add(i, 1) }
            {
                let val_ptr := add(add(val_0, 32), mul(i, 1))
                let val := mloadn(val_ptr, 1)
                pop(alloc_mstoren(val, 1))
            }
            // allocate up until the next multiple of 32 so our value is properly padded
            pop(alloc(sub(ceil32(array_data_size), array_data_size)))
        }
    }
    // store string42
    {
        pop(alloc_mstoren(mload(val_2), 32))
        {
            let array_size := mload(val_2)
            let array_data_size := mul(array_size, 1)
            for { let i := 0 } lt(i, array_size) { i := add(i, 1) }
            {
                let val_ptr := add(add(val_2, 32), mul(i, 1))
                let val := mloadn(val_ptr, 1)
                pop(alloc_mstoren(val, 1))
            }
            pop(alloc(sub(ceil32(array_data_size), array_data_size)))
        }
    }
    // store string100
    {
        pop(alloc_mstoren(mload(val_3), 32))
        {
            let array_size := mload(val_3)
            let array_data_size := mul(array_size, 1)
            for { let i := 0 } lt(i, array_size) { i := add(i, 1) }
            {
                let val_ptr := add(add(val_3, 32), mul(i, 1))
                let val := mloadn(val_ptr, 1)
                pop(alloc_mstoren(val, 1))
            }
            pop(alloc(sub(ceil32(array_data_size), array_data_size)))
        }
    }
}
```

[add info about decoding]

### To-Do

- [x] Add a new string type
- [x] Add ABI encoding for dynamically sized ABI types
- [x] Implement ABI encoding for strings
- [ ] Create issue for max-sizes in dynamically sized ABI types
- [ ] Create issue for pack/unpack operations
- [x] Update syntax of strings
- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/abi/functions.rs
+++ b/compiler/src/yul/abi/functions.rs
@@ -1,0 +1,264 @@
+use crate::yul::abi::utils::{
+    decode_name,
+    encode_name,
+    head_offsets,
+};
+use crate::yul::operations;
+use fe_semantics::namespace::types::{
+    AbiArraySize,
+    AbiDecodeLocation,
+    AbiEncoding,
+    AbiType,
+    AbiUintSize,
+};
+use yultsur::*;
+
+/// Generates an encoding function for any set of type parameters.
+pub fn encode<T: AbiEncoding>(types: Vec<T>) -> yul::Statement {
+    // the name of the function we're generating
+    let func_name = encode_name(&types);
+
+    // create a vector of identifiers and a vector of tuples, which contain
+    // expressions that correspond to the identifiers.
+    //
+    // The identifier vector is injected into the parameter section of our
+    // encoding function and the expressions are used to reference the parameters
+    // while encoding.
+    let (params, typed_params): (Vec<_>, Vec<_>) = types
+        .iter()
+        .enumerate()
+        .map(|(i, typ)| {
+            let ident = identifier! { (format!("val_{}", i)) };
+            let expr = identifier_expression! { [ident.clone()] };
+            (ident, (expr, typ))
+        })
+        .unzip();
+
+    // get the size of the static section
+    let (_, static_size) = head_offsets(&types);
+    let static_size = literal_expression! { (static_size) };
+
+    // we need to keep track of the dynamic section size to properly encode
+    // dynamically-sized array headers.
+    let mut dyn_sizes = vec![];
+    // we also keep a separate vector of dynamically-sized arrays, which we
+    // encode after finishing the static section.
+    let mut dyn_array_params = vec![];
+
+    // map the typed params to encoding statements for headers and statically-sized
+    // values
+    let static_encode_stmts = typed_params
+        .into_iter()
+        .map(|(param, typ)| match typ.abi_type() {
+            AbiType::Array {
+                inner,
+                size: AbiArraySize::Static { size },
+            } => encode_static_array(param, *inner, size),
+            AbiType::Array {
+                inner,
+                size: AbiArraySize::Dynamic,
+            } => {
+                let dyn_offset =
+                    expression! { add([static_size.clone()], [operations::sum(dyn_sizes.clone())]) };
+                let inner = *inner;
+                dyn_sizes.push(dyn_array_data_size(param.clone(), inner.clone()));
+                dyn_array_params.push((param, inner));
+                encode_uint(dyn_offset)
+            }
+            AbiType::Uint { .. } => encode_uint(param),
+        })
+        .collect::<Vec<_>>();
+
+    // map the dynamically-sized arrays to encoding statemnts
+    let dyn_encode_stmts = dyn_array_params
+        .into_iter()
+        .map(|(param, inner)| encode_dyn_array(param, inner))
+        .collect::<Vec<_>>();
+
+    function_definition! {
+        function [func_name]([params...]) -> ptr {
+            (ptr := avail())
+            [static_encode_stmts...]
+            [dyn_encode_stmts...]
+        }
+    }
+}
+
+fn dyn_array_data_size(val: yul::Expression, inner: AbiType) -> yul::Expression {
+    match inner {
+        AbiType::Array { .. } => unimplemented!(),
+        AbiType::Uint {
+            size: AbiUintSize { padded_size, .. },
+        } => {
+            let inner_padded_size = literal_expression! { (padded_size) };
+            let array_size = expression! { mload([val]) };
+            let elements_size = expression! { mul([array_size], [inner_padded_size]) };
+            expression! { add(32, (ceil32([elements_size]))) }
+        }
+    }
+}
+
+fn encode_uint(val: yul::Expression) -> yul::Statement {
+    statement! { pop((alloc_mstoren([val], 32))) }
+}
+
+fn encode_dyn_array(val: yul::Expression, inner: AbiType) -> yul::Statement {
+    let array_size = expression! { mload([val.clone()]) };
+    let array_start = expression! { add([val], 32) };
+    block_statement! {
+        [encode_uint(array_size.clone())]
+        [encode_array(array_start, inner, array_size)]
+    }
+}
+
+fn encode_static_array(val: yul::Expression, inner: AbiType, size: usize) -> yul::Statement {
+    let array_size = literal_expression! { (size) };
+    encode_array(val, inner, array_size)
+}
+
+fn encode_array(
+    val: yul::Expression,
+    inner: AbiType,
+    array_size: yul::Expression,
+) -> yul::Statement {
+    let (inner_data_size, inner_padded_size) = match inner {
+        AbiType::Uint {
+            size:
+                AbiUintSize {
+                    data_size,
+                    padded_size,
+                },
+        } => (
+            literal_expression! { (data_size) },
+            literal_expression! { (padded_size) },
+        ),
+        AbiType::Array { .. } => unimplemented!(),
+    };
+
+    block_statement! {
+        (let array_size := [array_size])
+        (let array_data_size := mul(array_size, [inner_padded_size.clone()]))
+        (for {(let i := 0)} (lt(i, array_size)) {(i := add(i, 1))}
+        {
+            (let val_ptr := add([val], (mul(i, [inner_data_size.clone()]))))
+            (let val := mloadn(val_ptr, [inner_data_size]))
+            (pop((alloc_mstoren(val, [inner_padded_size]))))
+        })
+        (pop((alloc((sub((ceil32(array_data_size)), array_data_size))))))
+    }
+}
+
+/// Generates an decoding function for a single type parameter in either
+/// calldata or memory.
+pub fn decode<T: AbiEncoding>(typ: T, location: AbiDecodeLocation) -> yul::Statement {
+    let func_name = decode_name(&typ, location.clone());
+
+    let decode_expr = match typ.abi_type() {
+        AbiType::Uint { .. } => decode_uint(location),
+        AbiType::Array { inner, size } => decode_array(*inner, size, location),
+    };
+
+    function_definition! {
+         // `start` refers to the beginning of the encoding
+         // `head_ptr` refers to the pointer at which the head is located
+         function [func_name](start, head_ptr) -> val {
+            (val := [decode_expr])
+         }
+    }
+}
+
+fn decode_uint(location: AbiDecodeLocation) -> yul::Expression {
+    match location {
+        AbiDecodeLocation::Memory => expression! { mload(head_ptr) },
+        AbiDecodeLocation::Calldata => expression! { calldataload(head_ptr) },
+    }
+}
+
+fn decode_array(
+    inner: AbiType,
+    size: AbiArraySize,
+    location: AbiDecodeLocation,
+) -> yul::Expression {
+    match inner {
+        // encoding of nested arrays
+        AbiType::Array { .. } => unimplemented!(),
+        // encoding of values that do not need to be unpacked
+        AbiType::Uint {
+            size:
+                AbiUintSize {
+                    padded_size,
+                    data_size,
+                },
+        } if padded_size == data_size => match size {
+            AbiArraySize::Static { size } => {
+                let total_size = size * data_size;
+                let total_size = literal_expression! { (total_size) };
+
+                match location {
+                    AbiDecodeLocation::Calldata => expression! { ccopy(head_ptr, [total_size]) },
+                    AbiDecodeLocation::Memory => expression! { head_ptr },
+                }
+            }
+            AbiArraySize::Dynamic => match location {
+                AbiDecodeLocation::Calldata => {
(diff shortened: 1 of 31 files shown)


compiler/tests/fixtures/strings.fe
```diff
@@ -0,0 +1,3 @@
+contract Foo:
+    pub def bar(s: string[100]) -> string[100]:
```
cburgdorf, line 2: Just repeating what we discussed in chat: As it stands, I think the syntax is a bit confusing because we have `u256[100]` defined as "An array with 100 `u256` items" so following that logic one could easily believe `string[100]` means "an array of 100 dynamically sized strings". Maybe an alternative syntax could be `string100`. That would be pretty much in line with something like `u8`, `u16`, ...; `u256`.

compiler/src/yul/runtime/functions.rs
```diff
@@ -88,6 +85,25 @@ pub fn mcopy() -> yul::Statement {
     }
 }
 
+/// Duplicates a segment of memory.
+pub fn mdup() -> yul::Statement {
```
g-r-a-n-t, line 89: delete this function

semantics/src/namespace/types.rs
```diff
@@ -9,37 +9,47 @@ pub trait FeSized {
...
-pub enum AbiPadding {
-    /// The element is padded on the left with some number of bytes.
-    Left { size: usize },
-    /// The element is padded on the right with some number of bytes.
-    Right { size: usize },
-    /// There is no padding. For example, `u256` are already 32 bytes, so there
-    /// is no need to pad them.
-    None,
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+pub struct AbiUintSize {
```
g-r-a-n-t, line 13: document

semantics/src/namespace/types.rs
```diff
@@ -9,37 +9,47 @@ pub trait FeSized {
...
+    pub data_size: usize,
+    pub padded_size: usize,
+}
+
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+pub enum AbiArraySize {
```
g-r-a-n-t, line 19: document

semantics/src/namespace/types.rs
```diff
@@ -9,37 +9,47 @@ pub trait FeSized {
...
+    Static { size: usize },
+    Dynamic,
```
g-r-a-n-t, line 21: create issue for dynamic types with max sizes. we should be able to attach a static max-size value here that is checked against the dynamic size during runtime. we should also allocate the max size when decoding

compiler/src/yul/abi/functions.rs
```diff
@@ -0,0 +1,205 @@
...
+    AbiArraySize,
+    AbiDecodeLocation,
+    AbiEncoding,
+    AbiType,
+    AbiUintSize,
+};
+use yultsur::*;
+
+/// Dynamically creates an encoding function for any set of type parameters.
+pub fn encode<T: AbiEncoding>(types: Vec<T>) -> yul::Statement {
```
g-r-a-n-t, line 17: add inline comments explaining whats going on for both functions in here. also, make sure all public functions in this module have api docs

compiler/src/yul/abi/functions.rs
```diff
@@ -0,0 +1,205 @@
...
+            literal_expression! { (data_size) },
+            literal_expression! { (padded_size) },
+        ),
+        AbiType::Array { .. } => unimplemented!(),
+    };
+
+    block_statement! {
+        (let array_size := [array_size])
+        (let array_data_size := mul(array_size, [inner_padded_size.clone()]))
+        (for {(let i := 0)} (lt(i, array_size)) {(i := add(i, 1))}
```
g-r-a-n-t, line 124: if the inner data_size and padded_size are the same, we can just copy the whole thing into our encoding, which is much more efficient. implement that in this PR or create a separate issue for it

compiler/src/yul/abi/functions.rs
```diff
@@ -0,0 +1,205 @@
...
+                    let total_size = expression! { add(32, (mul([array_size], [data_size]))) };
+
+                    expression! { ccopy([data_ptr], [total_size]) }
+                }
+                AbiDecodeLocation::Memory => {
+                    expression! { add(start, (mload(head_ptr))) }
+                }
+            },
+        },
+        AbiType::Uint { .. } => unimplemented!(),
```
g-r-a-n-t, line 189: the above arm on line 175 matches anything where `padded_size == data_size`, which means that no packing needs to take place. the arm tagged here is intended to handle anything that needs to be packed

for the time being, we are unable to decode arrays that contain things like addresses, since they are padded in the abi encoding. once pack/unpack operations are implemented, we can support this

compiler/src/yul/abi/operations.rs
```diff
@@ -0,0 +1,121 @@
...
+pub fn encode<T: AbiEncoding>(
+    types: Vec<T>,
+    vals: Vec<yul::Expression>,
+) -> (yul::Expression, yul::Expression) {
+    let func_name = encode_name(&types);
+    let size = encode_size(&types, vals.clone());
+    (expression! { [func_name]([vals...]) }, size)
+}
+
+fn encode_size<T: AbiEncoding>(types: &[T], vals: Vec<yul::Expression>) -> yul::Expression {
```
g-r-a-n-t, line 30: make this its own public function and remove its use in `encode`

compiler/src/yul/abi/operations.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    let static_size = literal_expression! { (static_size) };
+    return expression! { add([static_size], [operations::sum(dyn_size)]) };
+}
+
+/// Decode a sized value in calldata.
+///
+/// We currently support byte and u256 arrays and all base types.
```
g-r-a-n-t, line 65: update this doc

compiler/src/yul/abi/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+    };
+    full_name.push('_');
+    full_name.push_str(loc);
+    full_name.push('_');
+    full_name.push_str(&typ.abi_safe_name());
+
+    identifier! { (full_name) }
+}
+
+pub fn ceil_32(n: usize) -> usize {
```
g-r-a-n-t, line 67: use the simpler 1-line expression thats used in the runtime version of this function

compiler/src/yul/constructor.rs
```diff
@@ -50,11 +31,12 @@ pub fn build(init: Option<(yul::Statement, Vec<FixedSize>)>) -> yul::Code {
         // init function.
         block! {
             (let code_start := datasize("Contract"))
-            (let mem_start := alloc([params_size.clone()]))
```
g-r-a-n-t, line 53: these constructor variables should have better names

compiler/src/yul/mappers/functions.rs
```diff
@@ -63,6 +63,16 @@ pub fn func_def(
                     }
                 })
             }
+            FixedSize::String(string) => {
```
g-r-a-n-t, line 66: see if these function mappings can be cleaned up. matching the exact return type is a weird. also, get rid of the memory freeing stuff

compiler/src/yul/runtime/functions.rs
```diff
@@ -21,6 +16,8 @@ pub fn std() -> Vec<yul::Statement> {
         mstoren(),
         sstoren(),
         dualkeccak256(),
+        ceil32(),
+        mdup(),
```
g-r-a-n-t, line 20: remove this

compiler/src/yul/abi/functions.rs
```diff
@@ -0,0 +1,205 @@
...
+            AbiType::Uint { .. } => encode_uint(param),
+        })
+        .collect::<Vec<_>>();
+
+    let dyn_encode_stmts = dyn_array_params
+        .into_iter()
+        .map(|(param, inner)| encode_dyn_array(param, inner))
+        .collect::<Vec<_>>();
+
+    return function_definition! {
```
g-r-a-n-t, line 61: remove return

semantics/src/lib.rs
```diff
@@ -33,10 +34,15 @@ pub enum Location {
...
 /// Operations that need to be made available during runtime.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
 pub enum RuntimeOperations {
     /// Encode a set of fixed sized values.
     AbiEncode { params: Vec<FixedSize> },
+    /// Decode a value.
+    AbiDecode {
+        param: FixedSize,
+        loc: AbiDecodeLocation,
```
cburgdorf, line 44: When I first read the usage of `loc` somewhere further down I got confused for a second, thinking about "lines of code". Thoughts on renaming this to `location`?

compiler/src/yul/abi/utils.rs
```diff
@@ -0,0 +1,134 @@
+use fe_semantics::namespace::types::{
+    AbiArraySize,
+    AbiDecodeLocation,
+    AbiEncoding,
+    AbiType,
+    AbiUintSize,
+};
+use yultsur::*;
+
+/// returns the offset at which each head is located in the static section
```
cburgdorf, line 10: I guess docstrings should always begin with a capitalized word, right?

semantics/src/namespace/types.rs
```diff
@@ -320,6 +380,12 @@ pub fn type_desc(
         fe::TypeDesc::Base { base: "bool" } => Ok(Type::Base(Base::Bool)),
         fe::TypeDesc::Base { base: "bytes" } => Ok(Type::Base(Base::Byte)),
         fe::TypeDesc::Base { base: "address" } => Ok(Type::Base(Base::Address)),
+        fe::TypeDesc::Base { base } if &base[..6] == "string" => {
+            let max_size = base[6..]
+                .parse::<usize>()
```
cburgdorf, line 385: I wonder if there is a reasonable upper bound that should be placed on the `max_size` other than `usize`. Especially because, if I understand correctly, `usize` will be determined by the machine that compiles the contract. So, in theory if someone uses `string18446744073709551616` (2**64) such contract will compile on a 64-bit machine but not on a 32-bit machine. So, maybe just parsing this into `u32` (or something even lower) would be better?

cburgdorf (approved):
What a massive effort! Great engineering! I didn't examined each and every line in full detail but I think I got the general idea. I'm looking forward to our planned chat to talk through it.

semantics/src/namespace/types.rs
g-r-a-n-t: I think a reasonable upper bound would be based on the how much memory you could allocate without surpassing the block gas limit. So I guess that would look something like this:

`3 * (size / 32) + ((size / 32)^2 / 256) = 10,000,000`

Plugging this into Wolfram, I get `size = 1,606,844`.

For comparison, `2^32` is `4,294,967,296`, so the limit based on gas is pretty far from that.

I think if we were to have a check in place, we should check the value against the number based on gas (whatever type it may be).

g-r-a-n-t: reference: [memory fee on page 26](https://ethereum.github.io/yellowpaper/paper.pdf)

I've just replace `a` with `size / 32`, since that's how many words a string will occupy.

cburgdorf: Oh yes, I'm absolutely for choosing a more realistic upper bound. My main concern was that if we choose `usize` as upper bound then we all types that compile on one machine but doesn't on another. I mean, realistically, we can choose a number that is much much lower than `1,606,844`, too. How about we just pick a round number constant such as `1 000 000` or `100 000` or even just `10000`. Realistically, I can not imagine anything near those numbers to have practical use in the near term anyway.

compiler/tests/fixtures/erc20_token.fe
```diff
@@ -18,7 +18,7 @@ contract ERC20:
         idx to: address
         value: u256
 
-    def __init__(name: string, symbol: string):
+    def __init__(name: string[100], symbol: string[100]):
```
cburgdorf, line 21: Ah, I guess these still need to be updated to `string100`

PR #123, opened 2020-11-12
Add support for while loops

cburgdorf:
### What was wrong?

We do not yet support while loops. Additionally, there were two problems regarding scoping rules:

1. Declarations within block scopes did not work because statements in if/else or while loops where not semantically checked which also means declarations where not respected.

2. After the previous problem was fixed it became obvious that declarations where not properly scoped with block scoping rules. Instead, they were attributed to the enclosing function scope. While this is inline with Python, we like to diverge from Python here to follow rust (or JS/TS `let` scoping rules for that matter) scoping rules which introduces a new scope for each block.

### How was it fixed?

**While Loop Support**

- This builds on top of #116 (because it has some needed helpers)
- Semantic pass
  - refactored repetitive code into `verify_is_boolean` helper function
  - verify `test` expression is boolean
  - raise `unimplemented` if `or_else` is used
- Mapper pass:
  - mapped to the appropriate YUL code
- Tests
  - Added a test that proves the `while` is working and the code block supports arbitrarily big statements 

**Block Scope**

1. `FunctionScope` was renamed to `BlockScope`. The rational here being that functions simply introduce the first level `BlockScope` and are no different from e.g. any other block scope that is introduced by another block within that function.

2. Turn `parent` on `BlockScope` into a `BlockScopeParent` which can either point to a contract or block scope.

3. Ensure `contract_scope` walks upwards to find the contract scope as it may not be the immediate parent

4. Introduce `function_scope` method that finds the block scope that is the top most block scope and hence the immediate child of the `contract_scope`. 

5. Ensure `def` on `BlockScope` performs lookups on parent block scopes effectively inheriting all defs of its (block scope) parents.

6. Ensure `if` / `else` and `while` all introduce new block scopes.

### DIFF
--- a/compiler/src/yul/mappers/functions.rs
+++ b/compiler/src/yul/mappers/functions.rs
@@ -106,7 +106,7 @@ fn func_stmt(
         fe::FuncStmt::Emit { .. } => emit(context, stmt),
         fe::FuncStmt::AugAssign { .. } => unimplemented!(),
         fe::FuncStmt::For { .. } => unimplemented!(),
-        fe::FuncStmt::While { .. } => unimplemented!(),
+        fe::FuncStmt::While { .. } => while_loop(context, stmt),
         fe::FuncStmt::If { .. } => if_statement(context, stmt),
         fe::FuncStmt::Assert { .. } => assert(context, stmt),
         fe::FuncStmt::Expr { .. } => expr(context, stmt),
@@ -215,6 +215,30 @@ fn func_return(
             }
             None => return Ok(statement! { leave }),
         }
+    }
+
+    unreachable!()
+}
+
+fn while_loop(
+    context: &Context,
+    stmt: &Spanned<fe::FuncStmt>,
+) -> Result<yul::Statement, CompileError> {
+    if let fe::FuncStmt::While {
+        test,
+        body,
+        or_else: _,
+    } = &stmt.node
+    {
+        let test = expressions::expr(context, test)?;
+        let yul_body = multiple_func_stmt(context, body)?;
+
+        return Ok(block_statement! {
+            (for {} ([test]) {}
+            {
+                [yul_body...]
+            })
+        });
     }
 
     unreachable!()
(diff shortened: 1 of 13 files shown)


cburgdorf:
@g-r-a-n-t This could probably use a few more tests but I first wanted to get your opinion on the overall design choice of not separating between function and block scope at all. My main rationale is that it makes the code easier because all statements and expressions can always safely assume to work with a `BlockScope`. I also tend to think that it would be reasonable to argue there is no such thing as function scope, a function simply introduces the first block scope which is no different from any other nested block scope. Does that sound fair to you?

compiler/src/yul/mappers/functions.rs
```diff
@@ -219,3 +219,27 @@ fn func_return(
...
+    if let fe::FuncStmt::While {
+        test,
+        body,
+        or_else: _,
+    } = &stmt.node
+    {
+        let test = expressions::expr(context, test)?;
+        let yul_body = multiple_func_stmt(context, body)?;
+
+        return Ok(yul::Statement::Block(block! {
```
g-r-a-n-t, line 236: I added a `block_statement!` macro to yultsur recently for situations like this

semantics/src/namespace/scopes.rs
```diff
@@ -34,7 +34,7 @@ pub enum ContractDef {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub enum FunctionDef {
+pub enum BlockDef {
     Base(Base),
```
g-r-a-n-t, line 38: It would probably be better to have a single `Var` definition variant here, instead of a variant for each variable type. I modified this in my strings PR, so there's no need to update it here. I just wanted to mention this.

g-r-a-n-t (approved):
looks great!

compiler/src/yul/mappers/functions.rs
cburgdorf: Ah, nice! I updated the PR to use that.

PR #124, opened 2020-11-12
Updates on Logo Design

daniniguez:
### What was wrong?

Just a copule of iterations and updates regarding different sizes of the logo's final version.

### How was it fixed?

We did some fine tuning so the logo might be equally appreciated while being presented in different sizes.

### To-Do

No further action as of now.

- [ ] Clean up commit history


cburgdorf:
Thanks. There are a couple of things we need to resolve before this can be merged:

1. The PR needs to be rebased on latest master

2. The dimensions of the favicon seem to be wrong. It needs to be a square.

image

Reading through [this](https://stackoverflow.com/questions/2268204/favicon-dimensions) I get the impression that adding it in 260 x 260 would be best (and then we can downscale on demand via https://realfavicongenerator.net/). That said, if there were no actual changes (other than downscaling) so that it is basically a downscaled version of `fe_source_small_size.svg ` then we don't have to duplicate it at all and just having `fe_source_small_size.svg` will be enough.

3. Remove all other images from this PR (the merchandise, the 2d variant). We try to keep binaries to a minimum here to not bloat the size of the repository too much.

So, the PR should only contain:

1. `fe_source_small_size.svg` / `fe_source_small_size.png`
2. `fe_favicon.svg` / `fe_favicon.svg` (Notice the rename, Also only add these if there are true differences to the former other than resizing)

daniniguez:
<blockquote>1. The PR needs to be rebased on latest master. </blockquote> 

I believe there should be no conflicts as of now. I did some clean up and just left the one commit.

<blockquote>2. The dimensions of the favicon seem to be wrong. It needs to be a square. </blockquote>

Updated the file as per your request to a 260 x 260 px.

<blockquote>3. Remove all other images from this PR (the merchandise, the 2d variant)... </blockquote>

Merchandise and 2d variant removed.

PR contains:
 
<ol>
 <li><code>fe_source_small_size.svg / fe_source_small_size.png</code> (&#8776; 70 x 70 px, <em>original rectangular-shape ratio</em>)</li>
 <li><code>fe_favicon.svg / fe_favicon.png</code> (260 x 260 px, <em>modified square-shape ratio</em>)</li>
</ol>

Left both files since the original logo (*twitter's regular and small size logo iterations*) is more of a *rectangular-shape ratio*, whereas the favicon shape is a *perfect square-shape ratio* which in turn actually changes the feel/dimensions of the logo, but since the favicon is so small it isn't perceived at all. For larger sizes turning the logo into a perfect square would totally affect the overall outcome of the design.

cburgdorf:
Sorry, this got lost in my inbox. Looks good :+1:

daniniguez:
> Sorry, this got lost in my inbox. Looks good 👍

No problem. Awesome! 👍🏻

PR #125, opened 2020-11-17
Implement support for break statement

cburgdorf:
### What was wrong?

We don't support `break` statements yet.

### How was it fixed?

1. Added a new `BlockScopeType` to `BlockScope` that can be one of `Function`, `IfElse` or `Loop`.
2. Added `pub fn inherits_type(&self, typ: BlockScopeType) -> bool` to `BlockScope` to be able to tell if certain statements (e.g. `break`, `continue`) can be used within the current scope or not.
3. Semantic pass checks with `inherits_type` if placement is ok and returns `SemanticError::BreakWithoutLoop` if not
4. Mapper pass maps `break` to YUL

### DIFF
--- a/semantics/src/namespace/scopes.rs
+++ b/semantics/src/namespace/scopes.rs
@@ -57,6 +57,7 @@ pub struct BlockScope {
     pub span: Span,
     pub parent: BlockScopeParent,
     pub defs: HashMap<String, BlockDef>,
+    pub typ: BlockScopeType,
 }
 
 #[allow(dead_code)]
@@ -70,6 +71,13 @@ pub enum Scope {
 pub enum BlockScopeParent {
     Contract(Shared<ContractScope>),
     Block(Shared<BlockScope>),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BlockScopeType {
+    Function,
+    IfElse,
+    Loop,
 }
 
 impl Scope {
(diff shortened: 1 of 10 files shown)


semantics/src/namespace/scopes.rs
```diff
@@ -287,10 +325,76 @@ mod tests {
         let contract_scope = ContractScope::new(module_scope);
         let block_scope_1 =
             BlockScope::from_contract_scope(Span::new(0, 0), contract_scope.clone());
-        let block_scope_2 = BlockScope::from_block_scope(Span::new(0, 0), block_scope_1.clone());
+        let block_scope_2 = BlockScope::from_block_scope(
+            Span::new(0, 0),
+            BlockScopeType::IfElse,
+            block_scope_1.clone(),
```
g-r-a-n-t, line 331: I had been using `Rc::clone(&scope)`, since that's what's used in the API docs for Rc: https://doc.rust-lang.org/std/rc/struct.Rc.html.

Both of these do the same thing, of course, and I don't have a strong opinion about which method we use to call `clone`, but I think we should be consistent.

g-r-a-n-t (approved):
:+1:

semantics/src/namespace/scopes.rs
cburgdorf: Thanks for bringing this up. I wasn't aware that there was a good reason for the the more verbose syntax. I looked it up and `RC::clone(&ptr)` is indeed the preferred syntax. The reason for that is to make it clear what the `clone` is about since `clone` is a very vaguely defined action in Rust. In this case it is just increasing the reference counter whereas e.g. cloning a `Vec<T>` can cause quite a bit of memory getting copied.

References:

https://github.com/nical/rfcs/blob/rc-newref-clone/text/0000-rc-newref-clone.md
https://github.com/rust-lang/rust/pull/42137#issuecomment-304908345

I'm going to merge this as is and send a follow up PR to rewrite all occurrences with the idiomatic syntax.

cburgdorf: Updated all usage in this PR https://github.com/ethereum/fe/pull/126

Issue #129, opened 2020-11-20
ABI pack/unpack operations


g-r-a-n-t:
### What is wrong?

We currently have [inline unpacking](https://github.com/ethereum/fe/blob/90ca889637a487f943ac3924dada538b0d475e0f/compiler/src/yul/abi/functions.rs#L141-L146) taking place in our encoding functions, but [do not support packing](https://github.com/ethereum/fe/blob/90ca889637a487f943ac3924dada538b0d475e0f/compiler/src/yul/abi/functions.rs#L217) within our decoding function.  Furthermore, we are unnecessarily using an unpacking loop statement on non-padded types within our encoding functions..

### How can it be fixed

We should create two new [operations](compiler/src/yul/abi/operations.rs) (`pack` and `unpack`) that take in a segment of memory and unpack or pack the content. These operations should then be used inside of our ABI encode/decode operations in cases where the `padded_size` and `data_size` of the inner type of an ABI array are not equal. In cases where they are equal (i.e. `u256` arrays and `bytes`) we just copy or return the entire segment of memory without modifying the content.

PR #131, opened 2020-11-23
Add support for more numeric types

cburgdorf:
### What was wrong?

We did not support other numeric types as `u256`.

### How was it fixed?

Added support for `u8`, `u16`, `u32`, `u64` and `u128`

### To-Do

- [x] use in storage
- [x] use as value
- [x] basic arithmetic
- [x] abi encoding

### DIFF
--- a/compiler/tests/evm_contracts.rs
+++ b/compiler/tests/evm_contracts.rs
@@ -178,7 +178,7 @@ fn deploy_contract(
     panic!("Failed to create contract")
 }
 
-fn u256_token(n: usize) -> ethabi::Token {
+fn uint_token(n: usize) -> ethabi::Token {
     ethabi::Token::Uint(U256::from(n))
 }
 
@@ -199,7 +199,7 @@ fn bytes_token(s: &str) -> ethabi::Token {
 }
 
 fn u256_array_token(v: Vec<usize>) -> ethabi::Token {
-    ethabi::Token::FixedArray(v.into_iter().map(|n| u256_token(n)).collect())
+    ethabi::Token::FixedArray(v.into_iter().map(|n| uint_token(n)).collect())
 }
 
 #[test]
@@ -232,14 +232,14 @@ fn test_assert() {
     with_executor(&|mut executor| {
         let harness = deploy_contract(&mut executor, "assert.fe", "Foo", vec![]);
 
-        let exit1 = harness.capture_call(&mut executor, "bar", &vec![u256_token(4)]);
+        let exit1 = harness.capture_call(&mut executor, "bar", &vec![uint_token(4)]);
 
         assert!(matches!(
             exit1,
             evm::Capture::Exit((evm::ExitReason::Revert(_), _))
         ));
 
-        let exit2 = harness.capture_call(&mut executor, "bar", &vec![u256_token(42)]);
+        let exit2 = harness.capture_call(&mut executor, "bar", &vec![uint_token(42)]);
 
         assert!(matches!(
             exit2,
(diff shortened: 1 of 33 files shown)


compiler/tests/evm_contracts.rs
```diff
@@ -266,6 +267,7 @@ fn test_assert() {
     case("return_u256_from_called_fn.fe", vec![], Some(u256_token(42))),
     case("return_u256.fe", vec![], Some(u256_token(42))),
     case("return_identity_u256.fe", vec![42], Some(u256_token(42))),
+    case("return_identity_u128.fe", vec![42], Some(u256_token(42))),
```
cburgdorf, line 270: @g-r-a-n-t 
This test passes but I'm not 100 % sure if I'm testing the right thing because the inputs (`vec![42]`) are converted to `u256token`

```rust
input
                .clone()
                .into_iter()
                .map(|val| u256_token(val))
                .collect(),
```

I thought I need to create a `u128_token` variant of that but `ethabi::Token::Uint` simply expects a `U256` so maybe it's just something I don't understand :man_shrugging:

g-r-a-n-t: The ABI encoding for uints is the same for every size, so it shouldn't really matter in practice how we encode uints for these tests. We might just want to rename `u256_token` to `uint_token`.

The difference in sizes of uints should only really be relevant internally when it comes to [packing](https://github.com/ethereum/fe/issues/129) values in memory and storage more efficiently.

compiler/tests/evm_contracts.rs
```diff
@@ -249,43 +249,51 @@ fn test_assert() {
...
     case("return_bool_true.fe", vec![], Some(bool_token(true))),
     case("return_bool_false.fe", vec![], Some(bool_token(false))),
-    case("return_u256_from_called_fn_with_args.fe", vec![], Some(u256_token(200))),
-    case("return_u256_from_called_fn.fe", vec![], Some(u256_token(42))),
-    case("return_u256.fe", vec![], Some(u256_token(42))),
-    case("return_identity_u256.fe", vec![42], Some(u256_token(42))),
+    case("return_u256_from_called_fn_with_args.fe", vec![], Some(uint_token(200))),
+    case("return_u256_from_called_fn.fe", vec![], Some(uint_token(42))),
+    case("return_u256.fe", vec![], Some(uint_token(42))),
+    // case("return_u128.fe", vec![], Some(uint_token(42))),
```
cburgdorf, line 270: This seems like a tricky case to me. We currently assume numeric literals to always be of type `U256`. In the case of something like...

```
contract Foo:
    pub def bar() -> u128:
        return 42
```

...it seems the compiler should be able to infer that `42` should be of type `u128`. But I'm not sure yet how I would do that. It seems impossible to assign a type directly in `expr_num` without knowing how it is being used. If it is being used as part of a return (e.g. `return 42`) it would be reasonable to lookup the return value of the function but I'm not sure if `expr_num` should try to figure out the context of the usage. So, I'm wondering if `expr_num` should instead just say that this is of numeric type (could be any one at that point) and leave it up to e.g. the `return` to type check it. E.g. we could introduce `Base::AnyNumeric` which would be used as a placeholder for situations like that.

cburgdorf: Another idea. Instead of introducing a variant such as `AnyNumeric`. We could do something like this:

```
pub enum Base {
    Numeric(Option<NumericSize>),
    Bool,
    Byte,
    Address,
}

#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
pub enum NumericSize {
    U256,
    I256,
    U128,
    I128,
    U64,
    I64
    U32,
    I32
    U16,
    I16,
    U8,
    I8
}
```

So, `expr_num` would resolve to `Base::Numeric(None)` which is to say it is a numeric type but it could be any of them at this point. In practice though, it wouldn't be *any* of them because e.g. if I write `return 40000` then it can't fit into `u8` or `u16` but it could still be anything from `u32` (or `i32`) upward. So, I'm still not sure how exactly this should be dealt with. It feels like what we would like to express in `expr_num` would be a list of allowed types and then leave it to other code to narrow it down.

Anyway, looking forward to discuss that with you after the holidays.

g-r-a-n-t:
```
contract Foo:
    pub def bar() -> u128:
        return 42
```

I think having `42` cast to a `u128` implicitly makes sense, but for now, it might be easier to just require the use of type constructors. So it could instead look like this:

```
contract Foo:
    pub def bar() -> u128:
        return u128(42)
```

In the example above, `42` would be given the type `u256` (as would be the case with any positive literals) and then the return of the constructor would be given type `u128`. With this approach, there's no mystery about the type of an expression, so we don't need to use an optional, which I think would simplify things using the numeric type.

When we get around to introducing implicit casts, we could do something similar to [`move_location`](https://github.com/ethereum/fe/blob/08fcdee7ecebf56d64ef6a104b2a40662c2b48cd/semantics/src/lib.rs#L66), where we would have an attribute like `cast_type: Optional<Type>` that could be set by the expression's context if it needs to be cast to a certain type (in this case `return 42` would cast the return expression to `u128`). We can worry about this later though.

cburgdorf: Ah, yes, requiring to go through a type constructor strikes me as a pragmatic solution to keep things simple for now :+1:

cburgdorf: I've put this into its own issue to keep this PR short and focused https://github.com/ethereum/fe/issues/136

semantics/src/namespace/types.rs
```diff
@@ -93,12 +93,24 @@ pub enum FixedSize {
...
 
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+pub enum Integer {
+    U256,
+    U128,
+    U64,
+    U32,
+    U16,
+    U8,
+}
```
cburgdorf, line 110: My idea is to let the `Integer` enum host everyting from `u8`..`u256`  to `i8`..`i256`. This is to keep the hierarchy shorter and hence simplify matching etc.

semantics/src/namespace/types.rs
```diff
@@ -93,12 +93,24 @@ pub enum FixedSize {
...
+
+pub const U256: Base = Base::Numeric(Integer::U256);
```
cburgdorf, line 112: I created this as a shortcut since it is used in quite a bunch of tests etc

cburgdorf:
I think this is ready for review @g-r-a-n-t

semantics/src/namespace/types.rs
```diff
@@ -192,7 +204,7 @@ impl FixedSize {
 impl FeSized for Base {
     fn size(&self) -> usize {
         match self {
-            Base::U256 => 32,
+            Base::Numeric(_) => 32,
```
g-r-a-n-t, line 207: We should have `FeSized` implemented for `Integer` and then this arm would look something like `Base::Numeric(integer) => integer.size()`. Each of the sizes would be the same as the ABI data sizes below.

If we don't, there will be issues reading and writing to arrays that have been decoded and packed in memory. It also results in memory savings (e.g. `u8[10]` would occupy 10 bytes as opposed to 320).

g-r-a-n-t (approved):
Looks good, just one comment about the FeSized impl.

semantics/src/namespace/types.rs
cburgdorf: Ah, makes sense! I just pushed the update.

Issue #130, opened 2020-11-20
Add support for ternary expressions


cburgdorf:
### What is wrong?

We do not yet support ternary expressions (even though they are already recognized by the parser)

Example that should be working but currently doesn't:

```python
contract Foo:

    pub def bar(input: u256) -> u256:
       return 1 if input > 5 else 0
```

### How can it be fixed

1. Take the above snippet, create a fixture file from it and invoke it here:

https://github.com/ethereum/fe/blob/90ca889637a487f943ac3924dada538b0d475e0f/compiler/tests/evm_contracts.rs#L256-L259

2. Run `cargo test --workspace --features solc-backend test_method_return` and notice the failing tests

3. As the errors should hint, we need to add type checking in the semantic pass

https://github.com/ethereum/fe/blob/90ca889637a487f943ac3924dada538b0d475e0f/semantics/src/traversal/expressions.rs#L40

4. We then need to map the expression to its YUL equivalent in the mapper pass

https://github.com/ethereum/fe/blob/90ca889637a487f943ac3924dada538b0d475e0f/compiler/src/yul/mappers/expressions.rs#L27

cburgdorf:
This landed with #132

PR #132, opened 2020-11-24
Add support for ternary expression

satyamakgec:
### What was wrong?
#130 

### To-Do

- [x] Improve  `yul::mappers::expressions::exp_ternary` so it can return the valid `yul::Expression`.

- [ ] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -277,18 +277,16 @@ fn expr_ternary(
     exp: &Spanned<fe::Expr>,
 ) -> Result<yul::Expression, CompileError> {
     if let fe::Expr::Ternary {
-        test,
         if_expr,
+        test,
         else_expr,
     } = &exp.node
     {
-        let yul_test_expr = expr(context, test)?;
-        let yul_if_expr = expr_comp_operation(context, if_expr)?;
-        let yul_else_exp = expr(context, else_expr)?;
+        let yul_test_expr = expr_comp_operation(context, test)?;
+        let yul_if_expr = expr(context, if_expr)?;
+        let yul_else_expr = expr(context, else_expr)?;
 
-        // TODO: Need to find a way to return an expression that comprises all 3 of
-        // above. Dummy return to make it compile.
-        return Ok(literal_expression! {(1)});
+        return Ok(expression! {ternary([yul_test_expr], [yul_if_expr], [yul_else_expr])});
     }
     unreachable!()
 }
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -282,7 +282,7 @@ fn expr_ternary(
         else_expr,
     } = &exp.node
     {
-        let yul_test_expr = expr_comp_operation(context, test)?;
+        let yul_test_expr = expr(context, test)?;
         let yul_if_expr = expr(context, if_expr)?;
         let yul_else_expr = expr(context, else_expr)?;
 
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -276,13 +276,18 @@ fn expr_ternary(
     context: &Context,
     exp: &Spanned<fe::Expr>,
 ) -> Result<yul::Expression, CompileError> {
-    if let fe::Expr::Ternary {test, if_expr, else_expr} = &exp.node {
+    if let fe::Expr::Ternary {
+        test,
+        if_expr,
+        else_expr,
+    } = &exp.node
+    {
         let yul_test_expr = expr(context, test)?;
         let yul_if_expr = expr_comp_operation(context, if_expr)?;
         let yul_else_exp = expr(context, else_expr)?;
 
-        // TODO: Need to find a way to return an expression that comprises all 3 of above.
-        // Dummy return to make it compile.
+        // TODO: Need to find a way to return an expression that comprises all 3 of
+        // above. Dummy return to make it compile.
         return Ok(literal_expression! {(1)});
     }
     unreachable!()
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -24,7 +24,7 @@ pub fn expr(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressio
         fe::Expr::Bool(_) => expr_bool(exp),
         fe::Expr::Subscript { .. } => expr_subscript(context, exp),
         fe::Expr::Attribute { .. } => expr_attribute(context, exp),
-        fe::Expr::Ternary { .. } => unimplemented!(),
+        fe::Expr::Ternary { .. } => expr_ternary(context, exp),
         fe::Expr::BoolOperation { .. } => unimplemented!(),
         fe::Expr::BinOperation { .. } => expr_bin_operation(context, exp),
         fe::Expr::UnaryOperation { .. } => unimplemented!(),
@@ -270,6 +270,22 @@ fn expr_attribute_self(
         }) => Ok(literal_expression! { (index) }),
         _ => unimplemented!(),
     }
+}
+
+fn expr_ternary(
+    context: &Context,
+    exp: &Spanned<fe::Expr>,
+) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::Ternary {test, if_expr, else_expr} = &exp.node {
+        let yul_test_expr = expr(context, test)?;
+        let yul_if_expr = expr_comp_operation(context, if_expr)?;
+        let yul_else_exp = expr(context, else_expr)?;
+
+        // TODO: Need to find a way to return an expression that comprises all 3 of above.
+        // Dummy return to make it compile.
+        return Ok(literal_expression! {(1)});
+    }
+    unreachable!()
 }
 
 #[cfg(test)]
--- a/semantics/src/traversal/expressions.rs
+++ b/semantics/src/traversal/expressions.rs
@@ -329,22 +329,24 @@ fn expr_ternary(
     exp: &Spanned<fe::Expr>,
 ) -> Result<ExpressionAttributes, SemanticError> {
     if let fe::Expr::Ternary {
-        test,
         if_expr,
+        test,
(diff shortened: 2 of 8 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -272,6 +272,27 @@ fn expr_attribute_self(
...
+        else_expr,
+    } = &exp.node
+    {
+        let yul_test_expr = expr(context, test)?;
+        let yul_if_expr = expr_comp_operation(context, if_expr)?;
+        let yul_else_exp = expr(context, else_expr)?;
+
+        // TODO: Need to find a way to return an expression that comprises all 3 of
+        // above. Dummy return to make it compile.
+        return Ok(literal_expression! {(1)});
```
cburgdorf, line 291: >My understanding is that we are missing the appropriate macro in the yulstur crate, As I can't find a macro or helper function that generates a valid expression from those 3 expressions, Or maybe I am missing something.

I'm going to answer this here because there's a chance that I'll end up giving you a wrong or misleading answer and so I want to page in @g-r-a-n-t to give him the chance to correct me if I'm talking nonsense :sweat_smile: 

So, what I'm thinking is that this would be best mapped to a `switch` statement in YUL. You can see this being used here:

https://github.com/ethereum/fe/blob/c2681b844bf441ee296356ea6ab41f86fbd1e125/compiler/src/yul/mappers/functions.rs#L107-L111

Now, the difference here would be the `switch` is a statement so it doesn't resolve to any value. So, what I'm thinking is that we would need to map this to a bigger code out to something like this:

Pseudo-YUL
```
switch test
case 0 { $result := <if-expression> }
case 1 { $result := <else-expression }
result
```

So, basically we would generate a switch *statement* that introduces a variable to store the result of each possible arm followed by an *expression* that simply resolves to whatever is behind `$result`.

Without getting my hand on the code to actually try it, this is what I think I would try first.

@g-r-a-n-t  Does this make sense?

semantics/src/traversal/expressions.rs
```diff
@@ -323,6 +323,32 @@ fn expr_comp_operation(
...
+    context: Shared<Context>,
+    exp: &Spanned<fe::Expr>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    if let fe::Expr::Ternary {
+        test,
+        if_expr,
+        else_expr,
+    } = &exp.node
+    {
+        let test_attributes = expr(Rc::clone(&scope), Rc::clone(&context), test);
```
cburgdorf, line 337: One thing we need to check here is that the `test` is actually a boolean expression. Similar to

https://github.com/ethereum/fe/blob/c2681b844bf441ee296356ea6ab41f86fbd1e125/semantics/src/traversal/functions.rs#L232

satyamakgec: ahh, the hardcoded boolean statement makes me confuse. No worries I will make that change.

FYI here we need to check the `if_attributes` as a boolean expression, not the `test` as the test consists of the successful return value. I think we need to work on the naming or make some re-arrangements if we are expecting to `test` as the boolean expression. 
Here is the declaration of the ternary expression https://github.com/ethereum/fe/blob/c2681b844bf441ee296356ea6ab41f86fbd1e125/parser/src/ast.rs#L196

compiler/src/yul/mappers/expressions.rs
satyamakgec: > So, basically we would generate a switch statement that introduces a variable to store the result of each possible arm followed by an expression that simply resolves to whatever is behind $result.

The problem I saw is that we `switch!` is a statement I can't use to generate the expression so do you suggest using `yul::Switch` to generate the expression as you suggest above directly in the mappers instead of using the macros. 

Why do we need to store the result isn't we recursively generate the cases for it as per the given expressions.

semantics/src/traversal/expressions.rs
cburgdorf: Not sure I follow, let's break it down:

- `test`: This is the expression that is being tested e.g. `foo` in `1 if foo else 0`. The thing that we need to check for is that `foo` would would be an expression that evaluates to a `boolean` because non-booleans are illegal as a `test` in a ternary operator. E.g. we can not allow `foo` if it is of type `u256`.

- `if_expr`: The expression that is evaluated if `test` is `true`
- `else_expr`: The expression that is evaluated if `test` is false.

The thing that we need to check is that `if_expr` and `else_expr` evaluate to the same type. E.g. we can not allow one side to evaluate to a `address` if the other side evaluates to a `u256`

compiler/src/yul/mappers/expressions.rs
cburgdorf: >The problem I saw is that we switch! is a statement I can't use to generate the expression

Correct. That's why I was thinking we could generate a switch that stores the result and then have that followed by an *expression* that just evaluates to whatever is behind that identifier. But as I'm looking over the various options, it's not immediately obvious to me how that could be done. @g-r-a-n-t can you help us out?

g-r-a-n-t: Ah, yeah this is tricky. In Yul we can't wrap statements within an expression like you can in Rust... so we fundamentally can't map this entirely on its own.

```rust
// an expression in Rust
{ 
  let a = 42;
  a
}
```

The work around for this should be to create a helper function that contains the terany statements and takes the relevant expressions as parameters. I'm thinking it should maybe look something like this:

```javascript
function teranary(test, if_expression, else_expression) -> result {
  switch test
  case 0 { result := <if-expression> }
  case 1 { result := <else-expression }
}
```

This ternary function would be included within the [standard](https://github.com/ethereum/fe/blob/c2681b844bf441ee296356ea6ab41f86fbd1e125/compiler/src/yul/runtime/functions.rs#L5-L20) set of functions made available during runtime, then when we map a ternary expression, it should looks something like this:

`1 if input > 5 else 0` -> `ternary(input > 5, 1, 0)` 

Hopefully this works and thank you for the commits. This was supposed to be an easy one :laughing:

g-r-a-n-t: I actually wouldn't really call it a work around. Languages that map to Yul are supposed to be human-readable, so it's good to keep expressions short.

semantics/src/traversal/expressions.rs
satyamakgec: I think this is the syntax we are trying to achieve as per Python grammar:
```
[on_true] if [expression] else [on_false]
```
So for our example

```
contract Foo:

    pub def bar(input: u256) -> u256:
        return 1 if input > 5 else 0
```

on_true = 1, expression = input > 5, on_false = 0

So by the parser, I got the Ternary expression values something like this - 

 `test` - Spanned { node: Num("1"), span: Span { start: 68, end: 69 } } which lead to `test_attribute` value is ExpressionAttributes { typ: Base(U256), location: Value }, So I don't think it will ever be boolean it will always be the evaluated value if `if_expression` is true.

`if_expr` - Spanned { node: CompOperation { left: Spanned { node: Name("input"), span: Span { start: 73, end: 78 } }, op: Spanned { node: Gt, span: Span { start: 79, end: 80 } }, right: Spanned { node: Num("5"), span: Span { start: 81, end: 82 } } }, span: Span { start: 73, end: 82 } } which lead to `if_expr_attribute` value is ExpressionAttributes { typ: Base(Bool), location: Value }. Which always be the boolean and it should be.

`else_expr` - Spanned { node: Num("0"), span: Span { start: 88, end: 89 } } which lead to `else_attribute` value is ExpressionAttributes { typ: Base(U256), location: Value } . This is expected result.

So what I mean above is as per your terminology `test` & `if_expr` value should be used interchangebly but as per the code, it is reverse or in other words, it is not as per your terminology you above mention. So what I suggest above either we change the Ternary declaration from 
```
Ternary {
        test: Box<Spanned<Expr<'a>>>,
        if_expr: Box<Spanned<Expr<'a>>>,
        else_expr: Box<Spanned<Expr<'a>>>,
    },  
to 
Ternary {
        if_expr: Box<Spanned<Expr<'a>>>,
        test: Box<Spanned<Expr<'a>>>,
        else_expr: Box<Spanned<Expr<'a>>>,
    },  
```
Or we can fix the parser.

BTW I already in place the check for the same type that is evaluated from the `if_expr` & `else_expr`.

cburgdorf: >to 
```
Ternary {
        if_expr: Box<Spanned<Expr<'a>>>,
        test: Box<Spanned<Expr<'a>>>,
        else_expr: Box<Spanned<Expr<'a>>>,
    },
```

Ah, you mean changing the order of the enum properties to match the syntax of the language. We can definitely do that :+1:

compiler/src/yul/mappers/expressions.rs
satyamakgec: Thanks, @g-r-a-n-t  & @cburgdorf it works.

semantics/src/traversal/expressions.rs
satyamakgec: Minor parser changes were needed to facilitate the ordering.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -272,6 +272,25 @@ fn expr_attribute_self(
...
+    context: &Context,
+    exp: &Spanned<fe::Expr>,
+) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::Ternary {
+        if_expr,
+        test,
+        else_expr,
+    } = &exp.node
+    {
+        let yul_test_expr = expr_comp_operation(context, test)?;
```
cburgdorf, line 285: The test needs to use `expr` as well. Otherwise cases like these would fail:

Boolean literal:

```
    pub def bar(input: u256) -> u256:
        return 1 if true else 0
```

Function call:

Boolean literal:

```
    pub def bar(input: u256) -> u256:
        return 1 if self.check(input) else 0
```

cburgdorf (approved):
Left one comment regarding the `expr_comp_operation` that needs to be changed to `expr` but otherwise looks good. Great work! :tada:

compiler/src/yul/mappers/expressions.rs
satyamakgec: Thanks for the pointer. I fixed it.

PR #133, opened 2020-11-25
fix lint

g-r-a-n-t:
### What was wrong?

Lint was failing due to clippy upgrade.

### How was it fixed?

Followed all the suggestions.

### To-Do

- [x] Clean up commit history


Lifestag:
Unsub

Issue #90, opened 2020-10-13
Ensure order of function in contract doesn't matter for semantic analyses


cburgdorf:
### What is wrong?

As mentioned in [this comment](https://github.com/ethereum/fe/pull/78#discussion_r503350830), analyzation  of functions currently happens one at a time which causes us not have all needed information available depending on the order in which functions are defined in a contract.

E.g. semantic analyzes works as intended on this snippet:

```
contract Foo:

    pub def foo():
        revert

    pub def bar() -> u256:
        return self.foo()
```

But doesn't when the order is the other way around.

```
contract Foo:

   pub def bar() -> u256:
        return self.foo()

    pub def foo():
        revert
```

### How can it be fixed

Citing @g-r-a-n-t from https://github.com/ethereum/fe/pull/78#discussion_r503350830

>At the moment, function definitions are [analyzed one at a time](https://github.com/ethereum/fe/blob/de034e05af021904be516ebd93cf9ed63640b1d5/semantics/src/traversal/contracts.rs#L39) from top to bottom. So, since `foo` is defined after `bar`, we can't see `foo` in `bar`s function body.

>We should modify the [contract analysis](https://github.com/ethereum/fe/blob/de034e05af021904be516ebd93cf9ed63640b1d5/semantics/src/traversal/contracts.rs#L29) function to instead do an initial sweep on all function definitions that adds their signatures to the scope.

PR #135, opened 2020-11-27
Split analysis of function definition and function body to ensure order of function definitions doesn't matter

cburgdorf:
### What was wrong?

As explained in #90 we do currently have a bug that causes us to fail to compile contracts with functions that call other functions if the called function isn't defined above the calling function. In plain english: The order of function definitions matters but it should not matter.

### How was it fixed?

1. `func_def` was split into two functions: `func_def` and `func_body`. The former is only concerned about checking the types of the function definition and creating a scope for the function. It does not perform any tasks on the `body`  of the function.

2. The `scope` of the function is now also preserved when calling `add_function` on the contract scope. This is to give `function_body` a clean way to resolve the (already existing) scope.

3. In `contract_def`, we add a second pass that iterates over all functions again (after they have been iterated to call `func_def`) to call `func_body`. This is effectively what causes the order of the function definitions to become insignificant.

4. `func_body` retrieves the function scope by looking up the function definition in the contract scope and performs all checks on the function body.

### DIFF
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -29,7 +29,7 @@ use fe_parser::span::Spanned;
 use std::rc::Rc;
 
 /// Gather context information for a function definition and check for type
-/// errors.
+/// errors. Does not inspect the function body.
 pub fn func_def(
     contract_scope: Shared<ContractScope>,
     context: Shared<Context>,
@@ -40,7 +40,7 @@ pub fn func_def(
         name,
         args,
         return_type,
-        body,
+        body: _,
     } = &def.node
     {
         let function_scope = BlockScope::from_contract_scope(Rc::clone(&contract_scope));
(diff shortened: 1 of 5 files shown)


semantics/src/traversal/functions.rs
```diff
@@ -28,8 +28,6 @@ use fe_parser::ast as fe;
 use fe_parser::span::Spanned;
 use std::rc::Rc;
 
-/// Gather context information for a function definition and check for type
```
g-r-a-n-t, line 31: We should keep doc comments on all public functions

semantics/src/traversal/functions.rs
```diff
@@ -82,7 +72,44 @@ pub fn func_def(
 
         context.borrow_mut().add_function(def, attributes);
 
-        traverse_statements(function_scope, context, body)?;
+        return Ok(());
+    }
+
+    unreachable!();
+}
+
```
g-r-a-n-t, line 80: same, doc comment here too

g-r-a-n-t (approved):
Looks good :+1: just a few comments with regard to documentation

semantics/src/traversal/functions.rs
cburgdorf: Good point. Removal was by accident.

PR #137, opened 2020-12-01
Type constructors

### LINKED ISSUE #136: Implement type constructor for all numeric types
### What is wrong?

Currently, code such as the following does not compile.

```
contract Foo:
    pub def bar() -> u128:
        return 42
```

The reason why this fails is because a numeric literal such as `42` is always treated as an `u256`. While it would make sense to have the `42` be automatically treated as a `u128`, we can take more pragmatic route before we get there.

We can continue to treat positive numeric integers as `u256` and instead use **type constructors** to explicitly return `42` as an `u128` in this case.

```
contract Foo:
    pub def bar() -> u128:
        return u128(42)
```

Originally from https://github.com/ethereum/fe/pull/131#discussion_r531729974

g-r-a-n-t:
### What was wrong?

closes #136 

We did not support type constructors for addresses and integers.

### How was it fixed?

Expressions wrapped within a type constructor are given the type defined by the constructor.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -8,6 +8,7 @@ use fe_semantics::namespace::types::{
     Type,
 };
 use fe_semantics::{
+    CallType,
     Context,
     Location,
 };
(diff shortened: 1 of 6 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -92,17 +92,21 @@ pub fn expr_call(
...
-            let yul_args: Vec<yul::Expression> = arguments
-                .iter()
-                .map(|val| call_arg(context, val))
-                .collect::<Result<_, _>>()?;
-
-            let func_name = identifier! { (attr.node) };
-
-            return Ok(expression! { [func_name]([yul_args...]) });
-        }
+        return match &func.node {
```
g-r-a-n-t, line 95: I dislike matching the actual AST node here, we should use a different data structure provided by semantic analysis...

semantics/src/traversal/expressions.rs
```diff
@@ -307,34 +308,86 @@ fn expr_call(
...
-
-                    return Ok(ExpressionAttributes::new(
-                        return_type.into(),
-                        Location::Value,
-                    ));
+        return match &func.node {
+            fe::Expr::Attribute { value, attr } => {
+                let value = expr_name_string(value)?;
+
+                if value == "self" {
```
cburgdorf, line 315: Maybe we should put `"self"` in a constant?

compiler/src/yul/mappers/expressions.rs
cburgdorf: Yeah, I think the compiler books also say that the semantic pass would provide a new abstraction to the mapper pass. So far, I haven't seen how that would diverge in a meaningful way but I'm open to the idea :)

cburgdorf (approved):
LGTM

PR #139, opened 2020-12-01
Refactoring of scopes

cburgdorf:
### What was wrong?

Multiple things:

1. `return_type` on `BlockScope` didn't feel right (because it carries no meaning for block scopes that originate in loops or if/else)

2. Using enums for `ContractDef`, `BlockDef` and `ModuleDef` made usage less ergonomic because one always had to match arms that would not make sense leaving to excessive use of `unreachable!()` or other anti patterns.

### How was it fixed?

1. Instead of putting `return_type` on `BlockScope` there's now a `name` on `BlockScope` that can tie a `BlockScope` to a function name. Frankly one can argue that `name` wouldn't be strictly necessary for block scopes originating from loops or if/else but one *can* think of the benefits of naming these (e.g. `block_scope_foo_1` -> `blockscope_foo_2`) e.g. when visualizing scopes in a hierachy.

Having the name, one can easily lookup the `ContractFunctionDef` from a `BlockScope` by calling `current_function_def`. The `ContractFunctionDef` then carries information such as the `return_type` and more.

2. Instead of using enums for `ContractDef`, `BlockDef` and `ModuleDef` we now have dedicated `struct` (where needed). In some cases, we don't use any wrapper because e.g. we store the bare `FixedType` for variables on the block scope. This also means that the generic lookup method `pub fn def(&self, name: String) -> Option<ContractDef>` was refactored into dedicated lookups that either return `Option<Event>`, `Option<ContractFieldDef>` or `Option<ContractFunctionDef>`. This makes pattern matching more straight forward simply because there are less cases that can possibly be matched. It also means that all the different components are kept in their own dedicated hashmaps. 

3. Calling code was optimized to get rid of unnecessary arms or ditch pattern matching altogether where applicable.

### DIFF
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -43,9 +43,10 @@ pub fn func_def(
         body: _,
     } = &def.node
     {
-        let function_scope = BlockScope::from_contract_scope(Rc::clone(&contract_scope));
-
         let name = name.node.to_string();
+        let function_scope =
+            BlockScope::from_contract_scope(name.clone(), Rc::clone(&contract_scope));
+
         let param_types = args
             .iter()
             .map(|arg| func_def_arg(Rc::clone(&function_scope), arg))
@@ -56,8 +57,6 @@ pub fn func_def(
             .map(|typ| types::type_desc_fixed_size(Scope::Block(Rc::clone(&function_scope)), &typ))
             .transpose()?
             .unwrap_or_else(|| Tuple::empty().to_fixed_size());
-
-        function_scope.borrow_mut().return_type = return_type.clone();
 
         let is_public = qual.is_some();
         contract_scope.borrow_mut().add_function(
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -315,7 +315,7 @@ fn emit(
     {
         let event_name = expressions::expr_name_string(func)?;
 
-        if let Some(ContractDef::Event(event)) = scope.borrow().contract_def(event_name) {
+        if let Some(event) = scope.borrow().contract_event_def(event_name) {
             context.borrow_mut().add_emit(stmt, event);
         }
 
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -98,10 +98,7 @@ pub fn func_body(
         let host_func_def = contract_scope
             .borrow()
             .function_def(func_name)
-            .expect(&format!(
-                "Failed to lookup function definition for {}",
-                &name.node
-            ));
+            .unwrap_or_else(|| panic!("Failed to lookup function definition for {}", &name.node));
 
         // If the return type is an empty tuple we do not have to validate any further
         // at this point because both returning (explicit) or not returning (implicit
--- a/semantics/src/traversal/functions.rs
+++ b/semantics/src/traversal/functions.rs
@@ -2,7 +2,6 @@ use crate::errors::SemanticError;
 use crate::namespace::scopes::{
     BlockScope,
     BlockScopeType,
-    ContractDef,
     ContractScope,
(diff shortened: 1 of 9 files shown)


semantics/src/traversal/functions.rs
```diff
@@ -57,8 +57,6 @@ pub fn func_def(
             .transpose()?
             .unwrap_or_else(|| Tuple::empty().to_fixed_size());
 
-        function_scope.borrow_mut().return_type = return_type.clone();
```
cburgdorf, line 60: I came across this change today (I missed this in my review...I'm a terrible reviewer btw). I'm not a big fan of this. In fact, when I was rebasing #135 on top of latest master today I accidentally dropped this line and I was baffled for a while why things why functions are reporting the wrong return type. Also having `return_type` on the `BlockScope` doesn't feel right to me as it is an attribute that doesn't mean anything for block scopes that originate in loops or other blocks. I do realize that it could become meaningful some day but currently it isn't.

So, what I'm proposing in this PR instead is to enrich the `BlockScopeType::Function`  to carry the name of the function ``BlockScopeType::Function { name: String }`. That way, we can lookup the defintion of the function in the contract definitions and get rid of `return_type` in `BlockScope`.

The PR has some ugly spots that would need to be cleaned up but I wanted to get a draft of this out to get your opinion.

g-r-a-n-t: "refactor scopes" is something that's been bouncing around in my head alot lately, so good to see this.

I agree that having `return_type` on the block scope is a bit out of place. In the PR where this was merged, I tried doing something similar to what you described, but ran into issues with the block scope inheritance functions, since they rely on concrete enum variants, so this refactor might be a bit more involved.

cburgdorf: > but ran into issues with the block scope inheritance functions

Yes, I worked around that by introducing a `inherits_function` method which pattern matches on `BlockScopeType::Function {..}`. That said, it's actually not that useful since every block scope inherits from a function which makes it so that we only actually use that for tests.

>"refactor scopes" is something that's been bouncing around in my head a lot lately

Good! Maybe we stick heads together in the next meeting. I've also thought about more radical changes including traits etc but I didn't arrive on anything I was happy with. Huge part of that is probably due to just missing some good rust design patterns.

semantics/src/traversal/functions.rs
```diff
@@ -96,26 +94,26 @@ pub fn func_body(
...
-
-            traverse_statements(Rc::clone(&scope), Rc::clone(&context), body)?;
+        let func_name = name.node.to_string();
+        let host_func_def = contract_scope
+            .borrow()
+            .function_def(func_name)
+            .expect(&format!(
+                "Failed to lookup function definition for {}",
+                &name.node
+            ));
```
cburgdorf, line 104: Calling `expect` here made more sense to me because the only reason why this could fail is a programming error on our end that needs to be fixed. And if that case happens, the message will give us a meaningful hint. It also seams that the old code would have returned `Ok(())` even if the `if let` would not have matched (because of an (unexpected) `None` case).

semantics/src/traversal/functions.rs
```diff
@@ -379,7 +377,11 @@ fn func_return(
         let attributes =
             expressions::expr_with_default_move(Rc::clone(&scope), Rc::clone(&context), value)?;
 
-        if attributes.typ != scope.borrow().func_return_type().into() {
+        let host_func_def = scope
+            .borrow()
+            .current_function_def()
+            .expect("Failed to get function definition");
+        if attributes.typ != host_func_def.return_type.into() {
```
cburgdorf, line 384: Same here. If we are checking a `return` and fail to lookup the function hosting the `return` then clearly we are in an unexpected situation, a bug that needs to be fixed.

g-r-a-n-t (approved):

PR #143, opened 2020-12-04
Indexed event params for base types

### LINKED ISSUE #138: Indexed event parameters
### What is wrong?

We do not support indexed event parameters.

### How can it be fixed

Attributing `idx` to an event parameter should affect the event's behavior as one would expect. See the Solidity docs for more info https://docs.soliditylang.org/en/latest/contracts.html#events.

g-r-a-n-t:
### What was wrong?

closes #138 

We were not handling indexed event parameters correctly. They were being included inside of the log data, when they should have been included as extra log topics.

This PR does not include support for indexed non-base type params.

### How was it fixed?

- added indexed fields to the ABI builder
- the `idx` qualifier now affects the way events are logged
- we return an error if there are more than 3 indexed event params (no `log5` operation available)

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/abi/builder.rs
+++ b/compiler/src/abi/builder.rs
@@ -102,7 +102,7 @@ fn event_field<'a>(
     Ok(EventField {
         name: String::from(field.name.node),
         typ: type_desc(&type_defs, &field.typ.node)?,
-        indexed: false,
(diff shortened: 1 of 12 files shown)


cburgdorf (approved):
LGTM

PR #146, opened 2020-12-04
Add support for signed integers

cburgdorf:
### What was wrong?

We do not have support for signed integers yet.

### How was it fixed?

1. Added support for `i8`, `i16`, `i32`, `i64`, `i128`, `i256`
2. Added rudimentary support for `unaryOperation` just in so far to be able to write out something like `-3` and turn it into an `I256`
3. Ensure `<=`, `>=`, `/`, `%` and `>>` operations all work with signed numbers
4. Added a bunch of tests


compiler/src/yul/mappers/expressions.rs
```diff
@@ -167,6 +167,22 @@ pub fn expr_bin_operation(
...
+pub fn expr_unary_operation(
+    context: &Context,
+    exp: &Spanned<fe::Expr>,
+) -> Result<yul::Expression, CompileError> {
+    if let fe::Expr::UnaryOperation { op, operand } = &exp.node {
+        let yul_operand = expr(context, operand)?;
+        if let fe::UnaryOperator::USub = &op.node {
+            let zero = literal_expression! {0};
+            // TODO: Is there no way to just negate yul_operand?
+            return Ok(expression! { sub([zero], [yul_operand]) });
```
cburgdorf, line 179: @g-r-a-n-t  Maybe I'm doing it wrong. I've fiddled quite a while with this trying to find a way to just negate `[yul_operand]`. At the end I settled with the `sub(0, <expr>)` operation but it feels like a hack to me. Am I doing it wrong?

compiler/tests/evm_contracts.rs
```diff
@@ -290,65 +322,72 @@ fn test_assert() {
...
+    case("return_subtraction_i256.fe", vec![int_token(-42), int_token(-42)], Some(int_token(0))),
+    case("return_subtraction_i256.fe", vec![int_token(-42), int_token(42)], Some(int_token(-84))),
+    case("return_multiplication_u256.fe", vec![uint_token(42), uint_token(42)], Some(uint_token(1764))),
+    case("return_multiplication_i256.fe", vec![int_token(-42), int_token(-42)], Some(int_token(1764))),
+    case("return_multiplication_i256.fe", vec![int_token(-42), int_token(42)], Some(int_token(-1764))),
+    case("return_division_u256.fe", vec![uint_token(42), uint_token(42)], Some(uint_token(1))),
+    case("return_division_i256.fe", vec![int_token(-42), int_token(-42)], Some(int_token(1))),
+    case("return_division_i256.fe", vec![int_token(-1), int_token(1)], Some(int_token(-1))),
+    // TODO: Investigate what's wrong with this test
+    // case("return_division_i256.fe", vec![int_token(-42), int_token(42)], Some(int_token(-1))),
```
cburgdorf, line 349: Surprisingly, this test fails and I'm not entirely sure why. As we can see `-1 / 1 = -1` works but doing the same with any other number gives me surprises. I wonder if this is somehow related to https://github.com/ethereum/fe/pull/146/files#r538425906

semantics/src/traversal/expressions.rs
```diff
@@ -263,6 +263,29 @@ fn expr_bin_operation(
...
+    context: Shared<Context>,
+    exp: &Spanned<fe::Expr>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    if let fe::Expr::UnaryOperation { op, operand } = &exp.node {
+        if let fe::UnaryOperator::USub = &op.node {
+            let _operand_attributes =
+                expr_with_value_move(Rc::clone(&scope), Rc::clone(&context), operand)?;
+            // No matter what type the operand was before, the minus symbol turns it into an
+            // I256 just like all positive values default to U256.
+            return Ok(ExpressionAttributes::new(
```
cburgdorf, line 277: I just want to highlight this part. As far as I understand this means that any literal such as `-1` will be treated as an `I256`  which seems inline with the fact that any positive number such as `1` defaults to `U256`. But it also means that expressions such as `-x` will default to `I256`  even if `x` was something like an `U8` before.

cburgdorf: Oh and I should probably reject using `-` on any operand that isn't of type `Numeric`?

cburgdorf:
@g-r-a-n-t This still has a few quirks or at least things that seem not right to me. Would be great to get your opinion on it.

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: I think this makes the most sense

compiler/tests/evm_contracts.rs
g-r-a-n-t: I think we'll need to write custom runtime functions for multiplication and division of signed integers. As I understand it, addition and subtraction works the same for signed and unsigned integers, but multiplication and division are different. `n / 1 = n` is the only case where the operations are the same.

It might be best to mark these operations as unimplemented and take care of them later.

semantics/src/traversal/expressions.rs
g-r-a-n-t: > But it also means that expressions such as -x will default to I256 even if x was something like an U8 before.

We could also make sure that `x` is a literal. Do we want to have expressions like `- (foo + 42)`?

> Oh and I should probably reject using - on any operand that isn't of type Numeric?

:+1:

compiler/src/yul/mappers/expressions.rs
cburgdorf: But doesn't this waste gas for the `sub` operation? Shouldn't we be able to just turn `6` into `-6` and `foo` into `-foo`? Is this a limitation of the shorthand macros or yul itself? I guess I could take a snippet of solidity and run it through `--experimental-via-ir` to find out. Same goes for the division case.

compiler/tests/evm_contracts.rs
cburgdorf: >but multiplication and division are different

Ah ok. I guess I'll take a solidity example and generate the YUL code from it to see what it compiles to.  

For the record, these multiplication cases work fine.
```
  case("return_multiplication_i256.fe", vec![int_token(-42), int_token(-42)], Some(int_token(1764))),
  case("return_multiplication_i256.fe", vec![int_token(-42), int_token(42)], Some(int_token(-1764))),
```

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: I don't believe Yul has a way to represent negative numbers in the language itself, so there isn't a way for us to write `-6` or `-foo` in Yul. This behavior needs to be implemented by the Fe compiler on top of Yul's `u256` type. 

If we were to assume that negative signs may only be applied to literals (`-6` and not `-foo`), then we could potentially map this value to its correct binary value using an unsigned value. So, `-6` would become `2475880078570760549798248447` in the Yul code. This would save a small amount of gas, but it makes the Yul code more difficult to read.

The other option is to do what you already have and map `-6` to `sub(0, 6)`, which I think makes the Yul code more readable.

compiler/tests/evm_contracts.rs
g-r-a-n-t: > For the record, these multiplication cases work fine.

Interesting.

> Ah ok. I guess I'll take a solidity example and generate the YUL code from it to see what it compiles to.

Yeah, I'd like to see what Solidity is doing.

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: I was just reading the Yul docs again and saw that they have a few built in functions for signed operations: `sdiv`, `smod`, `slt`, `sgt`, and `sar`.

https://docs.soliditylang.org/en/latest/yul.html

cburgdorf: Oh! Nice find. That should help :)

semantics/src/traversal/expressions.rs
cburgdorf: >Do we want to have expressions like `- (foo + 42)`?

It's legal in Rust and Python and I think disallowing it might make the language more awkward to use so I'd say we should allow it.

compiler/tests/evm_contracts.rs
cburgdorf: For the record:

Multiplication
```solidity
      function checked_mul_t_int256(x, y) -> product {

                // overflow, if x > 0, y > 0 and x > (maxValue / y)
                if and(and(sgt(x, 0), sgt(y, 0)), gt(x, div(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, y))) { revert(0, 0) }
                // underflow, if x > 0, y < 0 and y < (minValue / x)
                if and(and(sgt(x, 0), slt(y, 0)), slt(y, sdiv(0x8000000000000000000000000000000000000000000000000000000000000000, x))) { revert(0, 0) }
                // underflow, if x < 0, y > 0 and x < (minValue / y)
                if and(and(slt(x, 0), sgt(y, 0)), slt(x, sdiv(0x8000000000000000000000000000000000000000000000000000000000000000, y))) { revert(0, 0) }
                // overflow, if x < 0, y < 0 and x < (maxValue / y)
                if and(and(slt(x, 0), slt(y, 0)), slt(x, sdiv(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, y))) { revert(0, 0) }

                product := mul(x, y)
            }
```

Division:

```solidity
           function checked_div_t_int256(x, y) -> r {
                if iszero(y) { revert(0, 0) }

                // overflow for minVal / -1
                if and(
                    eq(x, 0x8000000000000000000000000000000000000000000000000000000000000000),
                    eq(y, sub(0, 1))
                ) { revert(0, 0) }

                r := sdiv(x, y)
            }
```

cburgdorf: And tests are passing now with the `sdiv`. Still need to do some final cleanups after lunch before I'll ping you for another round of review :)

compiler/src/yul/mappers/expressions.rs
cburgdorf: Oh and btw, solidity negates numbers the same way: `sub(0, _value)`. In other news `solc foo.sol  --ir` is becoming my new swiss army knife :switzerland:

semantics/src/namespace/types.rs
```diff
@@ -133,13 +139,34 @@ pub struct FeString {
...
 
+impl Integer {
+    pub fn is_signed(&self) -> bool {
+        match self {
+            Integer::I256 => true,
+            Integer::I128 => true,
+            Integer::I64 => true,
+            Integer::I32 => true,
+            Integer::I16 => true,
+            Integer::I8 => true,
```
cburgdorf, line 150: Having written :point_up: I could also see the case for splitting signed and unsigned integers into their own enums but it may lead to less nice code elsewhere...so, I'm fine to leave it as is but I would also be open to change it.

cburgdorf:
@g-r-a-n-t This can be given another review

g-r-a-n-t (approved):

PR #151, opened 2020-12-08
Safe user-defined identifier mapping. 

### LINKED ISSUE #102: Ensure compiler generated methods can not conflict with user defined methods
### What is wrong?

Whenever we the compiler generates helper functions, that is, functions that aren't defined by the user, we need to ensure that they can not conflict with functions defined by the user. E.g. if we generate a function such as `abi_encode_uint256_address` we risk that a user may also create a function named `abi_encode_uint256_address` which can lead to undefined runtime behavior. 

### How can it be fixed

A simple fix would be to ensure all generated, non-user defined functions are prefixed with a symbol that is legal in YUL but illegal in Fe. E.g. instead of generating `abi_encode_uint256_address` we could generate `$abi_encode_uint256_address` instead.

g-r-a-n-t:
### What was wrong?

closes #102 

Identifiers chosen by users could conflict with identifiers introduced by the compiler.

### How was it fixed?

User-defined variable names now have `$` added to the front and user defined function names now have `$$` added to the front.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/declarations.rs
+++ b/compiler/src/yul/mappers/declarations.rs
@@ -1,5 +1,6 @@
 use crate::errors::CompileError;
 use crate::yul::mappers::expressions;
+use crate::yul::utils;
 use fe_parser::ast as fe;
 use fe_parser::span::Spanned;
 use fe_semantics::namespace::types::{
@@ -37,7 +38,7 @@ fn var_decl_base(
         value,
     } = &decl.node
     {
-        let target = identifier! { (expressions::expr_name_string(target)?) };
+        let target = utils::var_name(expressions::expr_name_str(target)?);
 
         return Ok(if let Some(value) = value {
             let value = expressions::expr(context, &value)?;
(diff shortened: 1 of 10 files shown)


compiler/src/yul/mappers/declarations.rs
```diff
@@ -126,6 +127,9 @@ mod tests {
             }),
         );
 
-        assert_eq!(map(&harness.context, &harness.src), "let foo := alloc(200)");
+        assert_eq!(
+            map(&harness.context, &harness.src),
+            "let $foo := alloc(200)"
```
cburgdorf, line 132: Am I right to assume that all the runtime functions such as `alloc` should also turn into `$$alloc` (or maybe some scheme that sets them apart from user defined functions) to ensure those aren't callable by user code?

cburgdorf (approved):
Looks good. Just left a comment regarding runtime functions but we could move that into a separate issue.

compiler/src/yul/mappers/declarations.rs
g-r-a-n-t: Anytime someone writes a call expressions (e.g. `foo(42)`), the function identifier will include `$$` after being mapped (`$$foo(42)`), so it should be impossible for someone to write something that calls `alloc` or any other runtime function directly.

cburgdorf: Ah, sure! :see_no_evil:

Issue #134, opened 2020-11-26
Implement for in loop.


cburgdorf:
### What is wrong?

We haven't yet mapped for loops to YUL. 

Here's a piece of code that should work but doesn't.

```
my_array: u256[5] = [1,2,3,4,5]
sum: u256 = 0
for n in my_array:
  sum = sum + n
```

### How can it be fixed

1. We need to implement semantic analysis in: https://github.com/ethereum/fe/blob/7aeca1a88c2f7e11e958a3d64fdf221295336ea9/semantics/src/traversal/functions.rs#L157

2. We need to map to YUL in: https://github.com/ethereum/fe/blob/7aeca1a88c2f7e11e958a3d64fdf221295336ea9/compiler/src/yul/mappers/functions.rs#L78

Since we know the size of `my_array` at compile time, it should be possible to map to a for loop that increments an array index from zero up to the array's size.

Inside of the Yul loop, we would have an operation that loads the value and assigns it to the name given in the iterator
i.e. `let n := mloadn(add(my_array, mul(i, <inner_size>)), <inner_size>)`

We would also have to add `n` to the block scope.

cburgdorf:
This landed a while ago

PR #154, opened 2020-12-08
Initial Skeleton to `for_loop`

satyamakgec:
### What was wrong?

#134 

### To-Do

- [ ] Clean up commit history


satyamakgec:
Hey @cburgdorf, I did some changes and solve the `Identifier not found`  error for `n` i.e target but now I am getting the error `Identifier not found` for `my_array`. The cause of the error is that the `my_array` variable is declared in the block statement so it has its own scope and not allowing to access the same in the for loop block statement.

I am not sure how to return multiple statements without using `block_statement!` in the `compiler/src/yul/mappers/declaration::var_decl_array`.

Below is the generated yul

```
function $$bar() -> return_val {
                 {
                     let $my_array := avail() mstoren(add(avail(), mul(0, 32)), 1, 32) mstoren(add(avail(), mul(1, 32)), 2, 32) mstoren(add(avail(), mul(2, 32)), 3, 32) mstoren(add(avail(), mul(3, 32)), 4, 32) mstoren(add(avail(), mul(4, 32)), 5, 32) 
                }
                let $sum := 0 {
                     for {
                         let i := 0 
                    }
                    lt(i, 5) {
                         i := add(i, 1) 
                    }
                    {
                         let $n := mloadn(add($my_array, mul(i, 32)), 32) $sum := add($sum, $n) 
                    }
                    
                }
                {
                     return_val := $sum leave 
                }
                
            }
```
cc @g-r-a-n-t could you please also take a look to shed some light on it.

g-r-a-n-t:
Thanks for taking a shot at array literals @satyamakgec, it would be very cool to have this working.

So far everything we've been doing has worked with 1:1 statement mapping, but that doesn't look like it will hold up anymore. In order for this to work, we'll need to change the return type of our statement mapper functions from `Result<yul::Statement, CompileError>` to `Result<Vec<yul::Statement>, CompileError>` and then use a `fold` to accumulate all of our statements instead of a `map`.  Then instead of returning a block of statements for a declaration (which creates a new scope), we would just return a sequence of statements.

If you don't want to do this refactor in this PR, that's fine. Feel free to drop the array literal part and focus on the loop, which shouldn't have this problem.

g-r-a-n-t:
Thanks for taking a shot at array literals @satyamakgec, it would be very cool to have this working.

At some point we will need to refactor our statement mapping functions to return multiple statements, instead of one, so that we don't have the scoping issue you're running into.

In this case, though, the array literal on the LHS should be an expression of its own that would be assigned to the declared variable the same way any other expression would be, so there should be no need to include extra statements for populating the new array. At the moment, I don't have a clear idea of how array literal expressions should be handled, so it might be a bit too much to bite off.

I would suggest that we drop the array literal part here and focus on for loops, as I think array literals is the most challenging part.

satyamakgec:
Thanks @g-r-a-n-t, I think for the `for_loop` only we are already ready to go, I need to add some test cases to support that. Do you want to remove the extra code that is in favour of array literals? I think we can keep it as it doesn't affect the other parts and it is a working code. 

Do we also want to support the `range()` function right now, to have `for_loop` something like this - 
```
contract Foo:
  pub def bar(range: u256) -> u256:
      sum: u256 = 0;
      for i in range(0, range):
         sum = sum + i
      return sum
```
I think that will also need array literals to work.

g-r-a-n-t:
@satyamakgec the for loop code looks good, but I do think it would be better to pull the array literals out for now.

The function

```rust
pub fn expr_list(
    context: &Context,
    exp: &Spanned<fe::Expr>,
) -> Result<Vec<yul::Expression>, CompileError> { 
... 
}
```

is a bit off because it needs to map to a single Yul expression..

```rust
pub fn expr_list(
    context: &Context,
    exp: &Spanned<fe::Expr>,
) -> Result<yul::Expression, CompileError> {
...
}
```

That way it can be used like any other expression, which is what a user would expect. Also, from the looks of it, we would still have scoping issues since we're mapping to a Yul block.

If you're interested, there are a couple ways that we could map array literals to single expressions that I can think:

1.) Recursively write each value to memory.

`[1,2,3,4]` -> 
```
take1(
  alloc_mstoren(1, <elem_size>),
  take1(
    alloc_mstoren(2, <elem_size>),
    take1(
      alloc_mstoren(3, <elem_size>),
      alloc_mstoren(4, <elem_size>),
    )
  )
)
```

definition of `take1`:

```
function take1(val1, val2) -> return_val {
  return_val := val1
}
```

**pros:** 
- unbounded number of values
- might not be that hard to implement :crossed_fingers: 

**cons:** 
- looks ugly

2.) Add a special runtime function for each array literal.

`[1,2,3,4]` -> `array_literal_<elem_size>_4(1,2,3,4)`

**pros**: 
- looks clean

**cons**:
- more overhead (you have to add a function to the runtime for each array literal)
- bounded number of values (`DUP15` is the highest dup operation, which limits the number of variables allowed in scope to 16)

satyamakgec:
Thanks @g-r-a-n-t for providing the tips.

Some followup questions - 

1. Do you really think recursion will work, I don't think so because for recursion to work we need predictable values in an array, If the user provides the array literal something like this [1, 5,7,8,4] then I don't think the above methodology works as I am not able to pass the vector of yul expression in the `function_definition!`.

2. For the explicit function types, For clarity do you proposing to have different runtime functions as per the array literal type & size both? If yes then I think it will be bloat of code and not a worthy effort to solve the problem. I think we have to find a clever way to achieve it either supports an array of expressions in the `function_definition!` to solve the problem easily.

What I mean to support the `Vec<yul::Expression>` in the `function_definition!` is to directly pass that as a single param. then maybe iteratively assign array literals to the memory and assign it to the target as an expression.

Maybe my understanding is wrong somewhere please let me know your thoughts.

satyamakgec:
Or we can close this task and remove the `expr_call()` from the yul/mapper.rs and create another issue to discuss the same.

g-r-a-n-t:
sorry for the late reply @satyamakgec, was on break.

> Do you really think recursion will work, I don't think so because for recursion to work we need predictable values in an array, If the user provides the array literal something like this [1, 5,7,8,4] then I don't think the above methodology works as I am not able to pass the vector of yul expression in the function_definition!.

If we were to go this route, we would basically just be mapping `[1,4,7,8,4]` to an independent Yul expression. This expression would need to do two things: 1.) return a pointer and 2.) store each value in memory. The approach suggested above does this by using a special function `take1` that that takes two parameters, returning the first and ignoring the second. This allows us to retain the pointer (evaluated in the first argument) and allows us to store the remaining values (done recursively in the second argument).

The Yul expression:

```
take1(
  alloc_mstoren(1, <elem_size>),
  take1(
    alloc_mstoren(2, <elem_size>),
    take1(
      alloc_mstoren(3, <elem_size>),
      alloc_mstoren(4, <elem_size>),
    )
  )
)
```

would be formed by looping over each mapped sub-expression in reverse and wrapping them in a new `take1` call.

I'm pretty sure this should be possible with the existing macros by doing something like:

```
accum = expression! { 
  take1(
    alloc_mstoren(
      [curr_elem],
      <elem_size>
    ),
    [accum]
};
```

but again, it's not the most human-readable mapping. The main thing this gets us is a consolidated mapping that does not bloat the runtime.

> For the explicit function types, For clarity do you proposing to have different runtime functions as per the array literal type & size both? If yes then I think it will be bloat of code and not a worthy effort to solve the problem. I think we have to find a clever way to achieve it either supports an array of expressions in the function_definition! to solve the problem easily.

I'd definitely be happy to find a solution better than the two I've proposed :) It would probably be better if we merged the loop code and started a separate issue for literal array expression.

satyamakgec:
@g-r-a-n-t I ripped out the literal support from the PR I will make a new issue to support array literals. Where we can discuss more thorughly.

g-r-a-n-t:
Great, thank you for the commits. It's nice to have for loops working now.

If we can find a time between our timezones, I'd be happy to have a call discussing array literals.

satyamakgec:
Yeah sure, how about 12:00 PM EST Tomorrow or Friday?

PR #155, opened 2020-12-09
Memory and storage copying

### LINKED ISSUE #152: Memory and storage copying
### What is wrong?

We do not have operations for copying segments of memory to memory or segments of storage to storage. This makes it so that assignments like `my_array = my_array` or `self.my_array = self.my_array` trigger an *unimplemented* panic during compilation. We should also be using memory-to-memory copying when passing values into other functions.

### How can it be fixed

Implement said copy operations and use them for assignments, declarations, and call arguments.

g-r-a-n-t:
### What was wrong?

closes #152 

- we didn't have operations for copying storage -> storage or memory -> memory
- the existing naming scheme for these operations was a bit confusing (e.g. `scopy` for storage -> memory copying)
- copy operations were going byte-by-byte, which is extremely inefficient

### How was it fixed?

- implemented said copying operations and used them in the relevant places (assignments and expression moves)
- updated the existing operations with a cleaner naming scheme
- copying now happens word-by-word

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/runtime/functions.rs
+++ b/compiler/src/yul/runtime/functions.rs
@@ -7,9 +7,11 @@ pub fn std() -> Vec<yul::Statement> {
         alloc(),
         alloc_mstoren(),
         free(),
-        ccopy(),
-        mcopy(),
-        scopy(),
+        ccopym(),
+        mcopys(),
+        scopym(),
+        mcopym(),
+        scopys(),
         mloadn(),
         sloadn(),
         cloadn(),
@@ -62,9 +64,9 @@ pub fn free() -> yul::Statement {
 }
 
 /// Copy calldata to memory a newly allocated segment of memory.
-pub fn ccopy() -> yul::Statement {
+pub fn ccopym() -> yul::Statement {
     function_definition! {
-        function ccopy(cptr, size) -> mptr {
+        function ccopym(cptr, size) -> mptr {
             (mptr := alloc(size))
             (calldatacopy(mptr, cptr, size))
         }
@@ -72,29 +74,94 @@ pub fn ccopy() -> yul::Statement {
 }
 
 /// Copy memory to a given segment of storage.
-pub fn mcopy() -> yul::Statement {
+pub fn mcopys() -> yul::Statement {
     function_definition! {
-        function mcopy(mptr, sptr, size) {
-            (for {(let i := 0)} (lt(i, size)) {(i := add(i, 1))}
+        function mcopys(mptr, sptr, size) {
+            (let offset := 0)
+            (for { } (lt((add(offset, 32)), size)) { }
             {
-                (let _mptr := add(mptr, i))
-                (let _sptr := add(sptr, i))
-                (sstoren(_sptr, (mloadn(_mptr, 1)), 1))
+                (let _mptr := add(mptr, offset))
+                (let _sptr := add(sptr, offset))
+                (sstore(_sptr, (mload(_mptr))))
+                (offset := add(offset, 32))
+            })
+
+            (let rem := sub(size, offset))
+            (if (gt(rem, 0)) {
+                (let _mptr := add(mptr, offset))
+                (let _sptr := add(sptr, offset))
+                (sstoren(_sptr, (mloadn(_mptr, rem)), rem))
             })
         }
     }
 }
 
 /// Copy storage to a newly allocated segment of memory.
-pub fn scopy() -> yul::Statement {
+pub fn scopym() -> yul::Statement {
     function_definition! {
-        function scopy(sptr, size) -> mptr {
+        function scopym(sptr, size) -> mptr {
             (mptr := alloc(size))
-            (for {(let i := 0)} (lt(i, size)) {(i := add(i, 1))}
+            (let offset := 0)
+            (for { } (lt((add(offset, 32)), size)) { }
             {
-                (let _mptr := add(mptr, i))
-                (let _sptr := add(sptr, i))
-                (mstoren(_mptr, (sloadn(_sptr, 1)), 1))
+                (let _mptr := add(mptr, offset))
+                (let _sptr := add(sptr, offset))
+                (mstore(_mptr, (sload(_sptr))))
+                (offset := add(offset, 32))
+            })
+
+            (let rem := sub(size, offset))
+            (if (gt(rem, 0)) {
+                (let _mptr := add(mptr, offset))
+                (let _sptr := add(sptr, offset))
+                (mstoren(_mptr, (sloadn(_sptr, rem)), rem))
+            })
+        }
+    }
+}
+
+/// Copies a segment of storage to another segment of storage.
+pub fn scopys() -> yul::Statement {
+    function_definition! {
+        function scopys(ptr1, ptr2, size) {
+            (let offset := 0)
+            (for { } (lt((add(offset, 32)), size)) { }
+            {
+                (let _ptr1 := add(ptr1, offset))
+                (let _ptr2 := add(ptr2, offset))
+                (sstore(_ptr2, (sload(_ptr1))))
+                (offset := add(offset, 32))
+            })
+
+            (let rem := sub(size, offset))
+            (if (gt(rem, 0)) {
+                (let _ptr1 := add(ptr1, offset))
+                (let _ptr2 := add(ptr2, offset))
+                (sstoren(_ptr2, (sloadn(_ptr1, rem)), rem))
+            })
+        }
+    }
+}
+
+/// Copies a segment of memory to another segment of memory.
+pub fn mcopym() -> yul::Statement {
+    function_definition! {
+        function mcopym(ptr1, size) -> ptr2 {
+            (ptr2 := alloc(size))
+            (let offset := 0)
+            (for { } (lt((add(offset, 32)), size)) { }
+            {
+                (let _ptr1 := add(ptr1, offset))
+                (let _ptr2 := add(ptr2, offset))
+                (mstore(_ptr2, (mload(_ptr1))))
+                (offset := add(offset, 32))
+            })
+
+            (let rem := sub(size, offset))
+            (if (gt(rem, 0)) {
+                (let _ptr1 := add(ptr1, offset))
+                (let _ptr2 := add(ptr2, offset))
+                (mstoren(_ptr2, (mloadn(_ptr1, rem)), rem))
             })
         }
     }
--- a/compiler/tests/fixtures/address_bytes10_map.fe
+++ b/compiler/tests/fixtures/address_bytes10_map.fe
@@ -2,7 +2,7 @@ contract Foo:
     pub bar: map<address, bytes[10]>
 
     pub def read_bar(key: address) -> bytes[10]:
-        return self.bar[key]
+        return self.bar[key].to_mem()
 
     pub def write_bar(key: address, value: bytes[10]):
         self.bar[key] = value
--- a/compiler/tests/fixtures/data_copying_stress.fe
+++ b/compiler/tests/fixtures/data_copying_stress.fe
@@ -1,0 +1,57 @@
+contract Foo:
+    my_string: string42
+    my_other_string: string42
+
+    my_u256: u256
+    my_other_u256: u256
+
+    my_nums: u256[5]
+
+    event MyEvent:
+        my_string: string42
+        my_u256: u256
+
+    pub def set_my_vals(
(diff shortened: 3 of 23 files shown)


compiler/src/yul/runtime/functions.rs
```diff
@@ -7,9 +7,11 @@ pub fn std() -> Vec<yul::Statement> {
         alloc(),
         alloc_mstoren(),
         free(),
-        ccopy(),
-        mcopy(),
-        scopy(),
+        calldata_to_mem(),
+        mem_to_sto(),
```
satyamakgec, line 11: can't we have naming like this `mcopys()`, `scopym()`, `mcopym()`, `scopys()`. It is more aligned with other runtime functions we have. We can think of a proper naming scheme that can be `location from + operation + location to`

compiler/src/yul/runtime/functions.rs
```diff
@@ -62,39 +64,104 @@ pub fn free() -> yul::Statement {
...
+                (let _sptr := add(sptr, offset))
+                (sstore(_sptr, (mload(_mptr))))
+                (offset := add(offset, 32))
+            })
+
+            (let rem := sub(size, offset))
+            (if (gt(rem, 0)) {
+                (let _mptr := add(mptr, offset))
+                (let _sptr := add(sptr, offset))
+                (sstoren(_sptr, (mloadn(_mptr, rem)), rem))
```
satyamakgec, line 93: Suggestion - All these functions have a similar definition the only difference I can see is the load operation & store operation. Can't we use macros to reduce the code dedup?

semantics/src/traversal/expressions.rs
```diff
@@ -296,97 +289,157 @@ fn expr_call(
...
+        let value_attributes = expr(scope, context, value)?;
+
+        // for now all of these function expect 0 arguments
+        if !args.node.is_empty() {
+            return Err(SemanticError::wrong_number_of_params());
+        }
+
+        return match attr.node {
+            builtins::CLONE => value_attributes.into_cloned(),
+            builtins::TO_MEM => value_attributes.into_cloned_from_sto(),
```
cburgdorf, line 374: So, `.clone()` would be used to do a memory to memory copy and `to_mem()` to make a storage to memory copy? We might have to fine tune the naming.

compiler/tests/fixtures/address_bytes10_map.fe
```diff
@@ -2,7 +2,7 @@ contract Foo:
     pub bar: map<address, bytes[10]>
 
     pub def read_bar(key: address) -> bytes[10]:
-        return self.bar[key]
```
cburgdorf, line 5: I think it would be good to add cases like this to our `compile_errors` tests to ensure they don't accidentally compile to something with undefined behavior. 

Am I right to assume that an assignment without `to_mem()` would now also be a compile error?

```
pub def failed_assignment(key: address):
    won_work: bytes[10 = self.bar[key]
```

I assume the above would fail and this would be needed to fix it.

```
pub def working_assignment(key: address):
    should_work: bytes[10 = self.bar[key].to_mem()
```

If that's the case, then I think we should add these to the compile errors.

compiler/tests/fixtures/data_copying_stress.fe
```diff
@@ -0,0 +1,45 @@
...
+        my_u256: u256,
+        my_other_u256: u256
+    ):
+        self.my_string = my_string
+        self.my_other_string = my_other_string
+        self.my_u256 = my_u256
+        self.my_other_u256 = my_other_u256
+
+    pub def set_to_my_other_vals():
+        self.my_string = self.my_other_string
```
cburgdorf, line 24: I'm confused about what is happening here. Are we *copying* because strings are currently copied by default? Or are we setting `self.my_string` to another storage reference?

My first reaction was: Ah, we are probably copying because strings are set to implicitly copy but then further down I see this code:

```
pub def emit_my_event():
        self.emit_my_event_internal(
            self.my_string.to_mem(),
            self.my_u256.to_mem()
        )
```

...which makes me think they aren't implicitly copied and need `.to_mem()` to explicitly copy them. Can you clarify what happens here?

To be honest, the more I think about it, the more I'm leaning towards **not implicitly copying** for any variable that is in storage no matter its type. This may feel a little awkward for things like numerics where people are just so used to having them copied implicitly but at the same time it  makes the semantics very very clear.

Because strings in Fe are currently not much different from arrays. Both have a fixed max length but can occupy any size between 0 and the max length. So you can have a `string10000` and it would be hard to explain why we implicitly copy it but demand `to_nem` for an `u256[10000]`.

compiler/tests/fixtures/address_bytes10_map.fe
cburgdorf: Ok, I pulled down the source and to my surprise, both of these compile:

Without `to_mem()`
```
    pub def some(key: address):
        x: bytes[10]
        x = self.bar[key]
```

with `to_mem()`

```
    pub def some(key: address):
        x: bytes[10]
        x = self.bar[key].to_mem()
```

I would have assumed that the version without `to_mem()` should fail, no?

cburgdorf (approved):
This is starting to look very nice. I left a few comments. There are still some things that confuse me. I'd also like to see some added tests for the compile errors we would expect from code that doesn't use `to_mem()` when it should. I left a comment with an example code that compiles without `to_mem()` which I *think* might be wrong. Would be great to walk through this PR in a call.

semantics/src/traversal/expressions.rs
g-r-a-n-t: What is a good alternative?

compiler/tests/fixtures/address_bytes10_map.fe
g-r-a-n-t: > I would have assumed that the version without to_mem() should fail, no?

Yes, I need to add an error here:

https://github.com/ethereum/fe/blob/07c09e516e13bd49ad65da936930a9353688b728/compiler/src/yul/mappers/assignments.rs#L61-L65

right now `to_mem` is implicit.

g-r-a-n-t: here actually: 

https://github.com/ethereum/fe/blob/07c09e516e13bd49ad65da936930a9353688b728/compiler/src/yul/mappers/assignments.rs#L40-L42

compiler/tests/fixtures/data_copying_stress.fe
g-r-a-n-t: > ...which makes me think they aren't implicitly copied and need .to_mem() to explicitly copy them. Can you clarify what happens here?

This is a memory to storage copy: `self.my_string = self.my_other_string.to_mem()`

This is a storage to storage copy: `self.my_string = self.my_other_string`

g-r-a-n-t: > Are we copying because strings are currently copied by default?

Strings, arrays, and any other reference types will be treated the same in terms of assignments.

semantics/src/traversal/expressions.rs
g-r-a-n-t: > So, .clone() would be used to do a memory to memory copy and to_mem() to make a storage to memory copy? 

And yeah, this is the case.

compiler/tests/fixtures/data_copying_stress.fe
g-r-a-n-t: > To be honest, the more I think about it, the more I'm leaning towards not implicitly copying for any variable that is in storage no matter its type. This may feel a little awkward for things like numerics where people are just so used to having them copied implicitly but at the same time it makes the semantics very very clear.

This is to say that something like

```
my_num: u256 = self.my_nums[42]
```

would not automatically load the value stored in the pointer at index 42 onto the stack? So we would instead need another attribute function to load this onto the stack?


```
my_num: u256 = self.my_nums[42].to_stack()
```

I think we should stick with the conventional behavior of loading onto the stack when indexing a value type.

compiler/src/yul/runtime/functions.rs
g-r-a-n-t: I think this looks alot better. thanks for the suggestion

compiler/src/yul/runtime/functions.rs
g-r-a-n-t: just tried my hand at it, but realized there's the difficulty of dealing with the different signatures on `sstore` and `mstore`

semantics/src/traversal/expressions.rs
cburgdorf: One alternative to consider would be to treat the difference as an implementation detail and just use `clone` for both. If I understand correctly, there isn't really a case where one has to choose between `clone()` or `to_mem()` because each does only apply in one specific scenario where the other one doesn't apply. But we can argue that eventually both methods allocate a new segment of memory. It is just the origin that is different. So maybe we could just branch into two different implementations internally but only expose `clone()`?

compiler/tests/fixtures/data_copying_stress.fe
cburgdorf: >I think we should stick with the conventional behavior of loading onto the stack when assigning a value type.

Ok, but isn't it confusing then that I have to call `to_mem()` on `self.my_u256` here. 

```
pub def emit_my_event():
        self.emit_my_event_internal(
            self.my_string.to_mem(),
            self.my_u256.to_mem()
        )
```

Why wouldn't that cause an implicit copy directly?

compiler/tests/fixtures/data_copying_stress.fe
```diff
@@ -0,0 +1,57 @@
...
+
+    pub def assign_my_nums_and_return() -> u256[5]:
+        my_nums_mem: u256[5]
+        self.my_nums[0] = 42
+        self.my_nums[1] = 26
+        self.my_nums[2] = 0
+        self.my_nums[3] = 1
+        self.my_nums[4] = 255
+        my_nums_mem = self.my_nums.to_mem()
+        return my_nums_mem
```
cburgdorf, line 48: Could the same thing have been achieved with?

```
pub def assign_my_nums_and_return() -> u256[5]:
        my_nums_mem: u256[5]
        my_nums_mem[0] = 42
        my_nums_mem[1] = 26
        my_nums_mem[2] = 0
        my_nums_mem[3] = 1
        my_nums_mem[4] = 255
        self.my_nums = my_nums_mem.clone()
        return my_nums_mem
```

So, all the individual assignments would be made in memory and then a copy of the whole memory array would be put in storage in one go. This is not about changing this, just for my understanding.

compiler/tests/fixtures/data_copying_stress.fe
```diff
@@ -0,0 +1,57 @@
...
+
+    pub def emit_my_event():
+        self.emit_my_event_internal(
+            self.my_string.to_mem(),
+            self.my_u256.to_mem()
```
cburgdorf, line 53: So, now spotted that further down in `sized_vals_in_sto.fe` we have this:

```
 emit MyEvent(
            self.num,
            self.nums.to_mem(),
            self.str.to_mem()
        ) 
```

Here we don't call `.to_mem()` on `self.num` which is what I would expect. So, I'm assuming that maybe this is a bug?

compiler/tests/fixtures/data_copying_stress.fe
cburgdorf: Btw, this is my mental model of how I assume things to work now:


|           |                      Mem  |                               Storage |
|---------|---------------------------|-------------------------------------|
| Mem  | `some = val.clone()`  | `some = self.val.to_mem()`  |
| Storage  |  `self.some = val.clone()` |  `self.some = self.val` |

For numeric

|           |          Mem  |            Storage   |
|---------|-----------------|-----------------------|
| Mem  | `some = val`  | `some = self.val`  |
| Storage  |  `self.some = val` |  `self.some = self.val` |

satyamakgec: I agree with the concern @cburgdorf in terms of semantics but for the user-friendliness, I will vote for implicit conversion to memory as this is something every code used to, Almost every other language does that. We may need to communicate our message more strongly to our developer community for some exceptions like a string in this case.

For the event case, I do agree with you @cburgdorf that it should be implicit conversion as with standard semantics. I haven't checked the code thoroughly to find gotchas in it but I am assuming @g-r-a-n-t is very well aware of this behavior and if we want we can log it in our backlog.

Issue #145, opened 2020-12-04
Enforce capacity of integer types


cburgdorf:
### What is wrong?

It's currently possible to use a type constructor of e.g. `u8` and feed it a number that is to high such as `u8(1000)`.

### How can it be fixed

1. When using type constructors with literals it should be enforced that the literal fits into the given type
2. We will probably also need to specify the type in the generated YUL code. E.g. I would expect that this code:

```
x = u8(1)
```

should translate into:

```
x := 1:u8
```

[Related YUL docs on integer literals](https://docs.soliditylang.org/en/v0.7.1/yul.html#literals)

PR #163, opened 2020-12-14
Ensure integer type constructor reject anything but numeric literals

cburgdorf:
### What was wrong?

Currently, our integer type constructor are too permissive. E.g. the following is allowed today even though it isn't guaranteed that `val` will fit into the `u16)`

```
    pub def bar(val: u8) -> u16:
        return u16(val)

```

I believe that our integer type constructor should simply not allow being used with anything other than a numeric literal such as `1` or `-3`.

Numeric literals can be checked on compile time whereas arbitrary variables can't. 

### How was it fixed?

- Added a function `validate_is_numeric_literal` and used it inside `expr_call` to validate
- Added a new error type `NumericLiteralExpected`
- Added a test to prove bad code is rejected.

Notice that this does not yet entail #145 but it is paving the way for it.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/compiler/tests/compile_errors.rs
+++ b/compiler/tests/compile_errors.rs
@@ -24,7 +24,8 @@ use std::fs;
     case("return_addition_with_mixed_types.fe", "TypeError"),
     case("return_lt_mixed_types.fe", "TypeError"),
     case("indexed_event.fe", "MoreThanThreeIndexedParams"),
-    case("unary_minus_on_bool.fe", "TypeError")
+    case("unary_minus_on_bool.fe", "TypeError"),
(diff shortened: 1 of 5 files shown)


g-r-a-n-t (approved):
Makes sense. We can perform compile time checks on literals like this and introduce methods like `my_num.as_i128()` for runtime casting.

cburgdorf:
>introduce methods like my_num.as_i128() for runtime casting.

Good idea. I'm writing up an issue for this. Eventually we might open up something like `u32(x)` in scenarios where `x` is <= `32` bit.

cburgdorf:
I created an issue for the runtime conversions https://github.com/ethereum/fe/issues/165

PR #167, opened 2020-12-15
Integer capacity checks

cburgdorf:
### What was wrong?

Previously it was possible to use numeric type constructors with literals that would not fit into the type. E.g. `u8(1000)` should error but currently doesn't.

### How was it fixed?

- Added bound checks for all numeric types
- Added tests for all expected compile errors
- Added tests that retrieve the min and max values for each numeric type.

### To-Do

- [x] Clean up commit history


semantics/src/namespace/types.rs
```diff
@@ -151,6 +169,43 @@ impl Integer {
...
+        fn handle_parse_error<T>(result: Result<T, ParseIntError>) -> bool {
+            if let Err(error) = result {
+                return match error.kind() {
+                    IntErrorKind::PosOverflow => false,
+                    IntErrorKind::NegOverflow => false,
+                    // If we try to parse a negative value for an unsigned type
+                    IntErrorKind::InvalidDigit => false,
+                    // We don't expect this but it would be tragic if we would map this to `false`
+                    // incase it happens because it would mean we sweep a bug under the rug.
+                    other => panic!("Unexpected ParseIntError: {:?}", other),
```
cburgdorf, line 185: @g-r-a-n-t  This is a case where I think we should panic because we never expect it to be hit but if we just map this to `false` and for some reason that branch is hit we would have a hard time searching for a very obscure bug.

cburgdorf:
I probably need to update the tarpaulin setup. If I install it via `cargo install cargo-tarpaulin` and then run `cargo tarpaulin --all --verbose` it works. If I run it through `./coverage.sh` I get the same failure locally. Will further investigate tomorrow.

g-r-a-n-t (approved):

PR #169, opened 2020-12-18
Updated solc to v0.8.0.

g-r-a-n-t:
### What was wrong?

We were using an older version of solc for Yul compilation.

### How was it fixed?

Went into solc-rust and updated the solidity submodule. There are still some things that I'd like to look into in solc-rust before opening a PR in that repo, so we're just using my personal fork for now.

Surprisingly, the solc-backend feature of the Fe compiler still works fine. I thought there would be atleast a couple breaking changes in libyul.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,6 +46,8 @@ jobs:
     - name: Install system dependencies
       run: |
         sudo apt-get install -y libboost-all-dev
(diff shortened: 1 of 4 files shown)


g-r-a-n-t:
looks like I need to get g++ 0.8 to install

PR #170, opened 2020-12-18
WIP

cburgdorf:
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/compiler/tests/fixtures/data_copying_stress.fe
+++ b/compiler/tests/fixtures/data_copying_stress.fe
@@ -26,6 +26,23 @@ contract Foo:
         self.my_string = self.my_other_string
         self.my_u256 = self.my_other_u256
 
+    pub def multiple_references_shared_memory(my_array: u256[10]):
+        my_2nd_array: u256[10]
+        my_3rd_array: u256[10]
+        my_2nd_array = my_array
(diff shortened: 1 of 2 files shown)


compiler/tests/fixtures/data_copying_stress.fe
```diff
@@ -26,6 +26,23 @@ contract Foo:
...
+    pub def multiple_references_shared_memory(my_array: u256[10]):
+        my_2nd_array: u256[10]
+        my_3rd_array: u256[10]
+        my_2nd_array = my_array
+        my_3rd_array = my_2nd_array
+
+        assert my_array[3] != 5
+        my_array[3] = 5
+        assert my_array[3] == 5
+        assert my_2nd_array[3] == 5
```
cburgdorf, line 38: @g-r-a-n-t based on our conversation I would expect `my_2nd_array` and `my_array` to point to the same location in memory and hence this assert to be true (as well as all the commented asserts further down).

But that isn't the case. My first reaction was, oh ok, maybe this is because `my_array` is coming from `CALLDATA`  so I tried the same with an array that I created locally but that didn't change anything.

Did I misunderstood the current state of how arrays are supposed to behave or is this a bug?

g-r-a-n-t:
closed in #171

Issue #176, opened 2021-01-11
Remove the mentions of Vyper in the README


fubuloubu:
### What is wrong?
Some people are still confused if Fe is related to Vyper, I think a potential reason why is because the README mentions Vyper a lot in defining what Fe is. Politics aside, it just reads poorly to have a bunch of statements that basically say "it's like this thing, but different", and it would be more useful to the reader to understand what Fe is without having to understand the context under how the project got created (although we do appreciate the free marketing!)

### How can it be fixed
Remove references to "Vyper" (or `rust-vyper`) and define Fe by itself: what it does, what are its principles, what makes it a good choice, etc. This will be better for users to read through rather than to see it as a derivative of another project, which will only confuse them more.

For context, Vyper had a lot of negative references to Solidity at one point early in it's existence, but we decided to remove many of those references and focus on what the language does well vs. comparing it to something else. In the long run, this helped the language differentiate itself from Solidity, and was friendlier to newcomers who may not have understood the references or where Vyper was their first experience with Ethereum smart contract development.

cburgdorf:
Yeah, our Readme feels outdated at this point. And as we are aiming for our first release this months this is something we should fix as soon as possible. Thanks for pointing it out. We'll make sure to remove references to Vyper. As you said, they aren't really helpful to understand what the language is about.

cburgdorf:
@fubuloubu the new Readme landed. If you find anything else that you think isn't quite right please let me know.

fubuloubu:
Reads much better now!

PR #177, opened 2021-01-11
Overhaul Readme

cburgdorf:
### What was wrong?

In the recent months there were several critics brought up regarding our Readme:

1. People had a hard time finding example and generally requested to see a piece of Fe code featured in the Readme
2. The information was found to be overcrowded
3.  In #176 it was brought up that Fe should distance itself more clearly from Vyper

The Readme hasn't had much overhaul since the project was originally born as *rust-vyper*. When it was written back then, a big motivation of the Readme was to compare and explain what was back then an alternative Vyper compiler. Naturally, the Readme has a lot of phrases that can be understood as "It's like the original Vyper but xyz". This doesn't reflect the current state of the project anymore which has since moved on to become Fe.

### How was it fixed?

A lot of the information was just deleted because frankly it doesn't need to be in the Readme at this point. Things like recently achieved goals or our development methodology.

The length of the Readme was roughly cut in half and reorganized to have fewer headlines. Things that were added are:

1. A two-liner about the language
2. A link to the Roadmap, our next big goal and an explanation that we will kick off monthly releases starting this month
3. An embedded code example 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,36 +6,12 @@ Fe is an emerging smart contract language for the Ethereum blockchain.
 [![Build Status](https://github.com/ethereum/fe/workflows/CI/badge.svg)](https://github.com/ethereum/fe/actions)
 [![Coverage](https://codecov.io/gh/ethereum/fe/branch/master/graph/badge.svg)](https://codecov.io/gh/ethereum/fe)
 
-## Getting started
-
-- [Build the compiler](https://github.com/ethereum/fe/blob/master/docs/build.md)
 
 ## Overview
 
-Fe is inspired by Vyper and aims to achieve the goals of the [existing Vyper
-project](https://github.com/vyperlang/vyper) along with the following
-additional goals:
+Fe is a statically typed language for the Ethereum Virtual Machine (EVM). It is inspired by Python and Rust and is easy to learn even for developers who are new to the Ethereum ecosystem.
 
-* Emphasis on correct implementation via use of better tooling, application of
-  formal methods, and stricter development practices
-* A complete language specification to aid in verification of compiler
-  correctness
-* Support for both eWASM and EVM code generation via use of the [YUL IR
-  format](https://solidity.readthedocs.io/en/v0.5.13/yul.html)
-* Implementation in a powerful, systems-oriented language (Rust) with strong
-  safety guarantees
-* Syntactic improvements and additional code re-use features
-* Statically built compiler binaries for smoother installation process
-* WASM compiler binaries for enhanced portability and in-browser compilation of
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,152 +21,53 @@ though we include a summary of them here:
 * Binary fixed-point math
 * Restrictions on reentrancy
 * Static looping
+* Module imports
+* Usage of [YUL](https://docs.soliditylang.org/en/latest/yul.html) IR to target both EVM and eWASM
+* WASM compiler binaries for enhanced portability and in-browser compilation of
+  Fe contracts
+* Implementation in a powerful, systems-oriented language (Rust) with strong safety guarantees to reduce risk of compiler bugs
```
cburgdorf, line 28: @g-r-a-n-t  I unified the project goals in this section. Let me know if you think I left out something important. I left out thinks that we aren't yet fully decided on such as inheritance or modifier

README.md
```diff
@@ -45,152 +21,53 @@ though we include a summary of them here:
...
+
+Additional information about design goals and background can be found in the [official announcement](https://snakecharmers.ethereum.org/fe-a-new-language-for-the-ethereum-ecosystem/).
+
```
cburgdorf, line 31: A bit of backstory is conserved in the announcement post so I think its enough to link to it here

README.md
```diff
@@ -45,152 +21,53 @@ though we include a summary of them here:
...
-### Short-term goals
-
-* [ ] Implement a functional Uniswap Demo in Fe
-
-### Long-term goals
-
-* Implement compilation routines for the full set of current Fe features
-* Implement a YUL compiler in Rust to eliminate the need to interact with the
-  Solidity compiler
-* Formally verify the Fe compiler using the language specification
```
cburgdorf, line 188: There's a lot of deleted stuff but I felt most is just clutter that will make it harder for people to get to know the language. If you think some of this should be preserved then maybe it could be moved to another document that we would link to.

README.md
```diff
@@ -45,152 +21,53 @@ though we include a summary of them here:
 * Binary fixed-point math
 * Restrictions on reentrancy
 * Static looping
+* Module imports
```
g-r-a-n-t, line 24:
```suggestion
* Module imports and a standard library
```

g-r-a-n-t (approved):

README.md
cburgdorf: Good point! Maybe even as separate items?

g-r-a-n-t: sounds good :+1:

Issue #153, opened 2020-12-08
Checked arithmetic by default.


cburgdorf:
### What is wrong?

We currently do not support any protection against over/underflows. We should add SafeMath-like behavior in that we perform these checks and revert in cases where we would today wrap around.

I can imagine that we eventually might want to add a way to explicitly allow for wrapping behavior for with different paths would exist e.g.

- Solidity uses [`unchecked`](https://github.com/ethereum/solidity/pull/9465/files) blocks
- Rust uses a special [`Wrapping`](https://doc.rust-lang.org/std/num/struct.Wrapping.html) type

I think that we can introduce checked arithmetic by default now and worry about opting-out at some later point in the future.

PR #179, opened 2021-01-12
Uniswap demo

g-r-a-n-t:
### What was wrong?

We need some contracts for the Uniswap demo.

### How was it fixed?

- copied the Solidity implementation and translated the contracts to Fe
- added testing
- documented remaining features needed for full support

### To-Do

- [x] Clean up commit history


g-r-a-n-t:
Still a few small things I want to clean up and features that need to be documented, but I thought it would be nice to get some feedback before the weekend.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
+contract UniswapV2Factory:
+    fee_to: address
+    fee_to_setter: address
+
+    pairs: map<address, map<address, address>>
+
+    # todo: discuss unbounded storage arrays
+    all_pairs: address[0]
```
cburgdorf, line 8: Good point. I've been thinking about this every now and then

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
...
+        self.fee_to_setter = fee_to_setter
+
+
+# This needs to be merged with UniswapV2Pair, since we do not support inheritance yet.
+# For the sake of readability though, we can leave it separate until we're ready to
+# deploy and test.
+contract UniswapV2ERC20:
+    # `const` is not supported by the parser yet.
+    # const name: string10 = "Uniswap V2"
+    # const symbol: string6 = "UNI-V2"
```
cburgdorf, line 61: This seems like a low hanging fruit that we could plan for soon. We should create an issue for it.

cburgdorf (commented):
Nice! Looks like we aren't too far off from getting it compiled. Also :+1: on moving these things into `/demo/`.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
+contract UniswapV2Factory:
+    fee_to: address
+    fee_to_setter: address
```
satyamakgec, line 3: Hmm, I did not notice that we don't support visibility specifiers. So by default, every variable is private?

g-r-a-n-t: Yeah, for this I was thinking that we should just manually implement `fee_to` and `fee_to_setter` public getter functions. It might be better, though, to add visibility modifiers for fields to the list of "nice to have features" and if we can't finish support for it by April, we would just use the public functions.

On a side note, we could add this to the list of things that could be implemented using a desugaring pass. 

i.e.

```
pub fee_to: address
```

becomes

```
fee_to: address

pub def fee_to() -> address:
  return `self.fee_to`
```

cburgdorf: >On a side note, we could add this to the list of things that could be implemented using a desugaring pass.

I like that idea :+1:

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
...
+
+    # called once by the factory at time of deployment
+    pub def initialize(token0: address, token1: address):
+        assert msg.sender == factory, "UniswapV2: FORBIDDEN"
+        self.token0 = token0
+        self.token1 = token1
+
+    # update reserves and, on the first call per block, price accumulators
+    def _update(balance0: u256, balance1: u256, reserve0: u112, reserve1: u112):
+        assert balance0 <= u112(-1) and balance1 <= u112(-1), "UniswapV2: OVERFLOW"
```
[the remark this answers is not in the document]
cburgdorf: Yeah, that's a good point. It feels hacky to me that some languages even allow this. We already have a check in place that something like `u8(-1)` doesn't compile (nothing of type `uxxx` allows passing `-1`). I think we want something like `u8.max()`, `u8.min()` to expose these.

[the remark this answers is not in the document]
satyamakgec: Yeah I like the idea of having `uxxx.max()` & `uxxx.min()`.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
...
+contract UniswapV2Pair:
+    # const MINIMUM_LIQUIDITY: u256 = 1000
+    # todo set the value below to keccak256('transfer(address,uint256)')
+    # const SELECTOR: u32 = 0
+
+    factory: address
+    token0: address
+    token1: address
+
+    reserve0: u112
```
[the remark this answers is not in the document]
satyamakgec: I haven't seen the code that does packing of the variables or maybe I did but missed it. I am more inclined to use every functionality that we have if support smart packing by the compiler itself then let's use it.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,359 @@
...
+
+        if fee_on:
+            self.k_last = reserve0 * reserve1 # reserve0 and reserve1 are up-to-date
+
+        emit Mint(msg.sender, amount0, amount1)
+
+    # this low-level function should be called from a contract which performs important safety checks
+    pub def burn(to: address): # -> (u256, u256):
+        # (reserve0: u112, reserve1: u112) = self.get_reserves() # gas savings
+        token0: ERC20 = ERC20.load(self.token0)
```
satyamakgec, line 279: `load` is the extra keyword needed to typecast explicitly for the contracts?

compiler/tests/fixtures/demos/uniswap.fe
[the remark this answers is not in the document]
g-r-a-n-t: The `u112` part is a direct carryover from the solidity implementation. I left it here for the time being just so we don't forget about something that's used to cut gas costs. It should be removed, though, since we don't support integers of size 112. Right now all base types occupy a whole word in storage, so there is no packing going on.

compiler/tests/fixtures/demos/uniswap.fe
g-r-a-n-t: I was thinking in terms of rust's `From` trait here, but used the term `load` instead of `from`. Personally, I think having typecasts of the form `Type(expression)` scattered around makes code somewhat difficult to read.

@cburgdorf, what do you think? You've done most of the work on typecasting so far and usually have good thoughts on syntax in general.

cburgdorf: I don't have a strong opinion here but I would mildly prefer to go with `ERC20(<address>)` instead of the `ERC20.load(<address>)` syntax.

g-r-a-n-t: I'll change it to `ERC20(address)` for now and we can see how it works.

compiler/tests/fixtures/demos/uniswap.fe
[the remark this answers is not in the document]
satyamakgec: That make sense now.

g-r-a-n-t:
@cburgdorf , core functionality seems to work. There are a few more things that we should try to get done before April, tho. These things are documented in TODO comments that link to issues.

compiler/tests/evm_contracts.rs
```diff
@@ -1238,3 +1281,241 @@ fn self_address() {
         );
     });
 }
+
+#[test]
+fn uniswap_contracts() {
```
cburgdorf, line 1286: The file is growing so large that I think we should start to split it up some. I think the Uniswap tests could all go into one dedicated uniswap test file.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,365 @@
...
+#        return ThreeNums(
+#            num1=self.reserve0,
+#            num2=self.reserve1,
+#            num3=self.block_time_stamp
+#        )
+        pass
+
+    # called once by the factory at time of deployment
+    pub def initialize(token0: address, token1: address):
+        assert msg.sender == self.factory, "UniswapV2: FORBIDDEN"
```
cburgdorf, line 136: If I'm not mistaken, passing a string to assert such as `"UniswapV2: FORBIDDEN"` is recognized by the parser but doesn't have any effect yet, right? If so, we should put a comment and issue on it as well.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,365 @@
...
+        self.token0 = token0
+        self.token1 = token1
+
+    # update reserves and, on the first call per block, price accumulators
+    def _update(balance0: u256, balance1: u256, reserve0: u256, reserve1: u256):
+        #assert balance0 <= u112(-1) and balance1 <= u112(-1), "UniswapV2: OVERFLOW"
```
cburgdorf, line 142: I guess `u112(-1)` is a solidity hack to mean `u112.max_value`. But do you know why this uses an `u112`? I guess we could also just initialize up a `u256` with the whatever the max value of a `u112` is and put that in a constant so that these would become: `assert balance0 <= SOME_MAXIMUM and balance1 <= SOME_MAXIMUM, "UniswapV2: OVERFLOW"`

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,365 @@
...
+        # changed from u32s
+        block_timestamp: u256 = block.timestamp % 2**32
+        time_elapsed: u256 = block_timestamp - self.block_timestamp_last # overflow is desired
```
cburgdorf, line 145: >`# overflow is desired`

If the overflow is desired than I think this would eventually have to use something like `wrapping(block_timestamp) - wrapping(self.block_timestamp_last)` since Fe will not allow checked arithmetic by default. This is mentioned here https://github.com/ethereum/fe/issues/153 and is inspired by Rust's [Wrapping](https://doc.rust-lang.org/std/num/struct.Wrapping.html) type. I think it shouldn't be too hard to implement.

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,365 @@
...
+        if time_elapsed > 0 and reserve0 != 0 and reserve1 != 0:
+            # `*` never overflows, and + overflow is desired
```
cburgdorf, line 147: Same here with the overflows

compiler/tests/fixtures/demos/uniswap.fe
```diff
@@ -0,0 +1,365 @@
...
+    pub def swap(amount0_out: u256, amount1_out: u256, to: address):
+        assert amount0_out > 0 or amount1_out > 0, "UniswapV2: INSUFFICIENT_OUTPUT_AMOUNT"
+        reserve0: u256 = self.reserve0
+        reserve1: u256 = self.reserve1
+        assert amount0_out < reserve0 and amount1_out < reserve1, "UniswapV2: INSUFFICIENT_LIQUIDITY"
+
+        token0: ERC20 = ERC20(self.token0)
+        token1: ERC20 = ERC20(self.token1)
+
+        assert to != address(token0) and to != address(token1), "UniswapV2: INVALID_TO"
```
cburgdorf, line 251: We talked about that instead of obtaining the address via `address(token0)` we could also obtain it via `token.address`. I think that would be inline with `self.address`, too. `address` would be a reserved word in the context of a contract properties. Since the ABI only defines functions it would still be allowed for a contract to implement an `address()` function though. We should probably create an issue and put a comment in for that, too.

cburgdorf (approved):
Really really cool! I left a few comments regarding things that I consider worth to create issues for and link to them in the example. I also think we should put the testing code into a separate file because our all-tests-in-one-file approach is starting to look unsustainable :sweat_smile: 

Other than that, looks amazing :+1:

compiler/tests/fixtures/demos/uniswap.fe
g-r-a-n-t: yeah, good point

compiler/tests/fixtures/demos/uniswap.fe
g-r-a-n-t: `u112` is used for more efficient storage packing of reserves. I think since we're using `u256` instead of `u112` for reserves, we could just delete this assert entirely.

compiler/tests/fixtures/demos/uniswap.fe
g-r-a-n-t: it seems like there is some nuanced behavior here. Im not quite sure why these overflows are desired..

for the time being, we could also consider just using `%` to reproduce the "desired" overflow

cburgdorf: But we will soon have checks for all arithmetic operations. So, we need to think about how that would work in the future. If you like the [Wrapping](https://doc.rust-lang.org/std/num/struct.Wrapping.html)  approach, I guess we could implement that fairly quickly. I would imagine we simply add another type `WrappedInteger` similar to `Integer` and a type constructor `wrapped` that e.g. produced as `WrappedInteger::U16` when you call `wrapped(some_u16_val)`. And the mapper would simply map these to the unchecked arithmetic opcodes. How does that sound to you?

g-r-a-n-t: Im not opposed to adding a wrapped type https://github.com/ethereum/fe/issues/286

PR #195, opened 2021-01-18
Add descriptions of to_mem and clone to spec.

g-r-a-n-t:
### What was wrong?

We were missing descriptions of these two builtin functions in the spec.

### How was it fixed?

Added descriptions in the data layout section.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -605,13 +605,13 @@ An *event type* is the type denoted by the name of an [`event` item].
 
 There are three places where data can be stored on the EVM:
 
-- **stack**: 256-bit values placed on the stack that are loaded using `DUP` operations. 
+- **stack**: 256-bit values placed on the stack that are loaded using `DUP` operations.
 - **storage**: 256-bit address space where 256-bit values can be stored. Accessing higher
 storage slots does not increase gas cost.
 - **memory**: 256-bit address space where 256-bit values can be stored. Accessing higher
 memory slots increases gas cost.
 
-Each data type described in section 5 can be stored in these locations. How data is stored is 
+Each data type described in section 5 can be stored in these locations. How data is stored is
 described in this section.
 
 ### 6.1. Stack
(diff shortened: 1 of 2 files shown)


spec/index.md
```diff
@@ -667,35 +667,55 @@ bar: map<address, u256> # bar is assigned a static nonce by the compiler
...
 given pointer.
 
+#### 6.2.3. The `clone` function
+
+Reference type values can be cloned using the `clone` function.
+
+Example:
+
+```python
+foo: u256[10] = bar.clone()
```
cburgdorf, line 716: Maybe this could be expanded to use some `assert`s to make the behavior more clear.

```
foo: u256[10] = bar.clone()
assert foo[1] == bar[1]
foo[1] = 1
assert foo[1] != bar[1]
```

We could even add in another section that explains what would happen if we don't use `clone`  e.g.

```
foo: u256[10] = bar
assert foo[1] == bar[1]
foo[1] = 1
assert foo[1] == bar[1]
```

cburgdorf (approved):
Nice. I left some comments inline where I think we might be able to make it even clearer. Also, this could use a `195.doc.md` file to go into the release notes (I will just regenerate the release notes in #194 which is why I put my manual edits into a second commit that I can just re-apply)

spec/index.md
satyamakgec: As per the example, it seems the cloning only copies the value instead of the pointer reference?

satyamakgec (approved):

spec/index.md
cburgdorf: Correct. Use `clone` to get an independent copy, omit clone and you actually copy the reference instead. We have tests that demonstrate it but I think it would make sense to expand the example right here.

satyamakgec: Yeah, makes sense @cburgdorf.

spec/index.md
```diff
@@ -667,35 +667,55 @@ bar: map<address, u256> # bar is assigned a static nonce by the compiler
...
-foo: u256[100] # foo is a pointer that references 100 * 256 bits in memory. 
+foo: u256[100] # foo is a pointer that references 100 * 256 bits in memory.
 ```
 
 To find an element inside of a sequence type, the relative location of the element is added to the
 given pointer.
 
+#### 6.2.3. The `clone` function
+
+Reference type values can be cloned using the `clone` function.
```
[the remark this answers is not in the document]
g-r-a-n-t: > But I think this would not be the case and that instead clone() can only be used to make a memory to memory clone, right?

yup

PR #197, opened 2021-01-20
Compile multiple contracts in single module.

g-r-a-n-t:
### What was wrong?

We were only able to compile the first contract in a module.

### How was it fixed?

- Updated the compiler module internals and API to handle multiple contracts.
- Updated the contract testing harness and added a test for a module with two separate contracts.
- Updated the CLI. We now write the following to the output directory:

```
output
|-- Bar
|   |-- Bar.bin
|   |-- Bar_abi.json
|   `-- Bar_ir.yul
|-- Foo
|   |-- Foo.bin
|   |-- Foo_abi.json
|   `-- Foo_ir.yul
|-- module.ast
`-- module.tokens
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/abi/elements.rs
+++ b/compiler/src/abi/elements.rs
@@ -9,32 +9,8 @@ use serde::{
 };
 use std::collections::HashMap;
 
-/// Wrapper around a map of contract names to their ABIs.
-#[derive(Serialize, Debug, PartialEq, Clone)]
-pub struct ModuleABIs {
-    #[serde(flatten)]
-    pub contracts: HashMap<String, Contract>,
-}
-
-impl Default for ModuleABIs {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl ModuleABIs {
-    pub fn new() -> Self {
-        Self {
-            contracts: HashMap::new(),
-        }
-    }
-
-    /// Serialize the module into a JSON object that maps each contract in the
-    /// module to its ABI.
-    pub fn json(&self) -> Result<String, CompileError> {
-        Ok(serde_json::to_string(self)?)
-    }
-}
+/// The ABIs for each contract in a Fe module.
+pub type ModuleAbis = HashMap<String, Contract>;
 
 /// All public interfaces of a Fe contract.
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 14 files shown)


compiler/src/abi/elements.rs
```diff
@@ -243,36 +219,8 @@ mod tests {
         FuncOutput,
         FuncType,
         Function,
-        ModuleABIs,
         VarType,
     };
-    use std::collections::HashMap;
-
-    #[test]
-    fn module_abis_json() {
```
g-r-a-n-t, line 252: I don't think there is any reason to serialize an entire module, so I removed that functionality and the test validating it.

newsfragments/197.feature.md
```diff
@@ -0,0 +1 @@
+The CLI now compiles every contract in a module, not just the first one.
```
cburgdorf, line 1: Maybe worth hinting about the compiled output with the example from the PR description.

```
output
|-- Bar
|   |-- Bar.bin
|   |-- Bar_abi.json
|   `-- Bar_ir.yul
|-- Foo
|   |-- Foo.bin
|   |-- Foo_abi.json
|   `-- Foo_ir.yul
|-- module.ast
`-- module.tokens
```

src/main_full.rs
```diff
@@ -66,96 +69,82 @@ pub fn main() {
...
+    output_dir: &str,
+    targets: &[CompilationTarget],
+) -> Result<(), String> {
+    let output_dir = Path::new(output_dir);
     verify_nonexistent_or_empty(output_dir)?;
     std::fs::create_dir_all(output_dir).map_err(ioerr_to_string)?;
 
-    let contents = fs::read_to_string(src_file).map_err(ioerr_to_string)?;
-
-    let output = fe_compiler::evm::compile(&contents, to_compile_stage(&targets))
```
cburgdorf, line 95: Just double checking since the compile stages seem to be gone. Can we still compile code that produces invalid YUL? E.g. with the current compiler one can compile `fe foo.fe --emit yul` even if `fe foo.fe` wouldn't work because the yul -> binary compilation errors. This is helpful to generate and inspect the faulty YUL code that fails to compile to binary.

cburgdorf (approved):
Looks good. One question I have is whether this removes support for generating YUL code when the YUL code is faulty which would be a bit sad since I find it quite helpful during development.

E.g. with the current compiler one can compile `fe foo.fe --emit yul` even if `fe foo.fe` wouldn't work because the yul -> binary compilation errors. This is helpful to generate and inspect the faulty YUL code that leads to the compilation failure.

src/main_full.rs
g-r-a-n-t: Good point :+1:

PR #202, opened 2021-01-22
Ensure event emits are type checked

cburgdorf:
### What was wrong?

On master, the following would not raise an error even though it should:

```
contract Foo:
    event MyEvent:
        val_1: string100
        val_2: u8

    pub def foo():
        emit MyEvent("foo", 1000)

```

### How was it fixed?

Basically reuse the `call_arg` that is defined in `expression` and a little bit of refactoring the event to expose all field types.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/tests/fixtures/compile_errors/call_event_with_wrong_types.fe
+++ b/compiler/tests/fixtures/compile_errors/call_event_with_wrong_types.fe
@@ -1,0 +1,7 @@
+contract Foo:
+    event MyEvent:
+        val_1: string100
+        idx val_2: u8
+
+    pub def foo():
+        emit MyEvent("foo", 1000)
--- a/compiler/tests/fixtures/strings.fe
+++ b/compiler/tests/fixtures/strings.fe
@@ -9,7 +9,7 @@ contract Foo:
         s5: string100
 
     pub def __init__(s1: string42, a: address, s2: string26, u: u256, s3: string100):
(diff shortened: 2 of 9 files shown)


compiler/tests/fixtures/strings.fe
```diff
@@ -9,7 +9,7 @@ contract Foo:
         s5: string100
 
     pub def __init__(s1: string42, a: address, s2: string26, u: u256, s3: string100):
-        emit MyEvent(s2, u, s1, s3, a, "static string", "foo")
+        emit MyEvent(s2, u, s1, s3, a, "static string", string100("foo"))
```
cburgdorf, line 12: This came to my attention after I merged #201 so it sparked this PR to tighten our type checks.

g-r-a-n-t (approved):

compiler/tests/fixtures/compile_errors/call_event_with_wrong_types.fe
```diff
@@ -0,0 +1,7 @@
+contract Foo:
+    event MyEvent:
+        val_1: string100
+        val_2: u8
```
g-r-a-n-t, line 4: for sanity it might be good to mark one of these as indexed

cburgdorf: Good point :+1:

Issue #168, opened 2020-12-16
Compiler bug panics


g-r-a-n-t:
### What is wrong?

Currently, the mapper module in the compiler library uses a `Result<_, CompileError>` return pattern for invalid attributes/ASTs. See [expressions](https://github.com/ethereum/fe/blob/220185f83544720967c6172075e1645e104bed7c/compiler/src/yul/mappers/expressions.rs#L23) as an example.

An error result here is indicative of a programmer mistake and cannot be recovered from in any meaningful way, so we should panic instead. See: https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html.

This is not to say that panicking is a good pattern. Generally, we should eliminate situations in the compiler where panics are used by defining attribute data structures that do not permit fallible mappings. This is not always practical, though, so the programmer should be able to leave panics in code branches that they assume will not be entered.

### How can it be fixed

- remove all `Result` returns in the mappers
- create a custom panic that would help a programmer identify the bug

See rustc's compiler panic as an example:

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', compiler/rustc_middle/src/ty/query/mod.rs:235:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust-clippy/issues/new

note: Clippy version: clippy 0.0.212 (0f6f2d6 2020-12-06)

query stack during panic:
end of query stack
error: could not compile `fe-compiler`

To learn more, run the command again with --verbose.
```

PR #204, opened 2021-01-26
Contract type and external calls.

g-r-a-n-t:
### What was wrong?

We were unable to make calls to external contracts.

### How was it fixed?

- Added a new `Contract` type. The contract type is treated like an address, but carries extra information about the contact's public interface.
- When compiling a contract, the module scope is checked for any external contract type definitions. If an external contract type is found, a set of functions is added to the contract's runtime for calling to the external contract's public interface. For now, contract type definitions are only visible to subsequently defined contracts, this can be changed later.
- Calls to external contracts in Fe (e.g. `Foo(foo_address).bar(42, my_string)`) are mapped to calls to the runtime functions mentioned above. So for example, the call mentioned before is mapped to `Foo_bar($foo_address, 42, $my_string)`. This makes the Yul code pretty easy to read.

Examples of external contract call runtime functions:

```javascript
// with return data
function Foo_build_array(addr, val_0, val_1) -> return_val {
  let instart := alloc_mstoren(0x649de2ee, 4) 
  let insize := add(4, add(64, 0)) 
  pop(abi_encode_uint256_uint256(val_0, val_1)) 
  let outsize := 96 let outstart := alloc(outsize) 
  pop(call(gas(), addr, 0, instart, insize, outstart, outsize)) 
  return_val := abi_decode_uint2563_mem(outstart, 0) 
}
            
// no return data
function Foo_emit_event(addr, val_0, val_1, val_2) {
  let instart := alloc_mstoren(0x16bc6ad2, 4) 
  let insize := add(4, add(256, ceil32(mul(mload(val_2), 1)))) 
  pop(abi_encode_uint256_address5_string11(val_0, val_1, val_2)) 
  pop(call(gas(), addr, 0, instart, insize, 0, 0)) 
}
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/namespace/scopes.rs
+++ b/analyzer/src/namespace/scopes.rs
@@ -106,6 +106,11 @@ impl ContractScope {
         }))
     }
 
+    /// Return the module scope that the contract scope inherits from
+    pub fn module_scope(&self) -> Shared<ModuleScope> {
+        Rc::clone(&self.parent)
+    }
+
     /// Lookup contract event definition by its name.
     pub fn event_def(&self, name: String) -> Option<Event> {
         self.event_defs.get(&name).map(|def| (*def).clone())
--- a/analyzer/src/namespace/types.rs
+++ b/analyzer/src/namespace/types.rs
@@ -7,7 +7,10 @@ use std::num::{
     ParseIntError,
 };
 
+use crate::FunctionAttributes;
 use num_bigint::BigInt;
+
+const ADDRESS_BYTE_LENGTH: usize = 20;
 
 pub fn u256_max() -> BigInt {
     BigInt::from(2).pow(256) - 1
@@ -99,6 +102,7 @@ pub enum Type {
     Map(Map),
     Tuple(Tuple),
     String(FeString),
+    Contract(Contract),
 }
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
@@ -107,6 +111,7 @@ pub enum FixedSize {
     Array(Array),
     Tuple(Tuple),
     String(FeString),
+    Contract(Contract),
 }
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
@@ -155,6 +160,12 @@ pub struct Tuple {
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
 pub struct FeString {
     pub max_size: usize,
+}
+
+#[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
+pub struct Contract {
+    pub name: String,
+    pub functions: Vec<FunctionAttributes>,
 }
 
 impl TryFrom<&str> for FeString {
@@ -246,6 +257,7 @@ impl From<FixedSize> for Type {
             FixedSize::Base(base) => Type::Base(base),
             FixedSize::Tuple(tuple) => Type::Tuple(tuple),
             FixedSize::String(string) => Type::String(string),
+            FixedSize::Contract(contract) => Type::Contract(contract),
         }
     }
 }
@@ -257,6 +269,7 @@ impl FeSized for FixedSize {
             FixedSize::Array(array) => array.size(),
             FixedSize::Tuple(tuple) => tuple.size(),
             FixedSize::String(string) => string.size(),
+            FixedSize::Contract(contract) => contract.size(),
         }
     }
 }
@@ -287,6 +300,7 @@ impl AbiEncoding for FixedSize {
             FixedSize::Base(base) => base.abi_name(),
             FixedSize::Tuple(tuple) => tuple.abi_name(),
             FixedSize::String(string) => string.abi_name(),
+            FixedSize::Contract(contract) => contract.abi_name(),
         }
     }
 
@@ -296,6 +310,7 @@ impl AbiEncoding for FixedSize {
             FixedSize::Base(base) => base.abi_safe_name(),
             FixedSize::Tuple(tuple) => tuple.abi_safe_name(),
             FixedSize::String(string) => string.abi_safe_name(),
+            FixedSize::Contract(contract) => contract.abi_safe_name(),
         }
     }
 
@@ -305,6 +320,7 @@ impl AbiEncoding for FixedSize {
             FixedSize::Array(array) => array.abi_type(),
             FixedSize::Tuple(tuple) => tuple.abi_type(),
             FixedSize::String(string) => string.abi_type(),
+            FixedSize::Contract(contract) => contract.abi_type(),
         }
     }
 }
@@ -333,6 +349,7 @@ impl TryFrom<Type> for FixedSize {
             Type::Tuple(tuple) => Ok(FixedSize::Tuple(tuple)),
             Type::String(string) => Ok(FixedSize::String(string)),
             Type::Map(_) => Err(SemanticError::type_error()),
+            Type::Contract(contract) => Ok(FixedSize::Contract(contract)),
         }
     }
 }
@@ -343,7 +360,7 @@ impl FeSized for Base {
             Base::Numeric(integer) => integer.size(),
             Base::Bool => 1,
             Base::Byte => 1,
-            Base::Address => 20,
+            Base::Address => ADDRESS_BYTE_LENGTH,
         }
     }
 }
@@ -457,7 +474,7 @@ impl AbiEncoding for Base {
             },
             Base::Address => AbiType::Uint {
                 size: AbiUintSize {
-                    data_size: 20,
+                    data_size: ADDRESS_BYTE_LENGTH,
                     padded_size: 32,
                 },
             },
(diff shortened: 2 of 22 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -105,12 +106,23 @@ pub fn expr_call(
 
                     Ok(expression! { [func_name]([yul_args...]) })
                 }
-                CallType::ValueAttribute { .. } => {
-                    if let fe::Expr::Attribute { value, .. } = &func.node {
-                        expr(context, value)
-                    } else {
-                        unreachable!()
+                CallType::ValueAttribute => {
+                    if let fe::Expr::Attribute { value, attr } = &func.node {
```
g-r-a-n-t, line 110: clean up

compiler/src/yul/runtime/functions/calls.rs
```diff
@@ -0,0 +1,77 @@
...
+                function_definition! {
+                    function [call_function_name](a, [param_idents...]) {
+                        (let instart := alloc_mstoren([selector], 4))
+                        (let insize := add(4, [encoding_size]))
+                        (pop([encoding_operation]))
+                        (pop((call((gas()), a, 0, instart, insize, 0, 0))))
+                    }
+                }
+            } else {
+                // todo: should prly use a separate function for static sizes
```
g-r-a-n-t, line 53: should be able to determine the max encoded size at compile time

compiler/src/yul/runtime/functions/calls.rs
```diff
@@ -0,0 +1,77 @@
+use crate::abi::utils as abi_utils;
+use crate::yul::names;
+use crate::yul::operations::abi as abi_operations;
+use fe_semantics::namespace::types::{AbiEncoding, AbiDecodeLocation};
+use fe_semantics::namespace::types::Contract;
+use yultsur::*;
+
+pub fn contract_calls(contract: Contract) -> Vec<yul::Statement> {
```
g-r-a-n-t, line 8: cleanup and add comments

compiler/src/yul/operations/calls.rs
```diff
@@ -0,0 +1,13 @@
+use crate::yul::names;
+use fe_semantics::namespace::types::Contract;
+use yultsur::*;
+
+pub fn contract_call(
```
g-r-a-n-t, line 5: docs

compiler/src/yul/names.rs
```diff
@@ -39,6 +39,11 @@ pub fn decode_name<T: AbiEncoding>(typ: &T, location: AbiDecodeLocation) -> yul:
     identifier! { (full_name) }
 }
 
+pub fn contract_call(contract_name: String, func_name: String) -> yul::Identifier {
```
g-r-a-n-t, line 42: docs

semantics/src/traversal/contracts.rs
```diff
@@ -46,9 +53,34 @@ pub fn contract_def(
             };
         }
 
-        context
+        let mut contract_attributes = ContractAttributes::from(contract_scope.borrow());
```
g-r-a-n-t, line 56: cleanup

semantics/src/traversal/expressions.rs
```diff
@@ -536,6 +537,34 @@ fn expr_call_value_attribute(
...
+    args: &Spanned<Vec<Spanned<fe::CallArg>>>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    if let Some(function) = contract
+        .functions
+        .iter()
+        .find(|function| function.name == attr.node)
+    {
+        let return_type = function.return_type.to_owned();
+
+        // todo: add type checking on args
```
g-r-a-n-t, line 554: add type checking and a compiler errors test

semantics/src/traversal/expressions.rs
```diff
@@ -601,6 +630,18 @@ fn expr_name_call_type(
                 TryFrom::try_from(value).map_err(|_| SemanticError::undefined_value())?,
             ),
         }),
+        value
```
g-r-a-n-t, line 633: clean up and make sure contract types can be assigned

analyzer/src/namespace/scopes.rs
```diff
@@ -106,6 +106,11 @@ impl ContractScope {
         }))
     }
 
+    /// Return the module scope that the block scope inherits from
```
g-r-a-n-t, line 109: fix docs

analyzer/src/traversal/expressions.rs
```diff
@@ -563,6 +589,51 @@ fn expr_call_value_attribute(
...
+    if let Some(function) = contract
+        .functions
+        .iter()
+        .find(|function| function.name == func_name)
+    {
+        let return_type = function.return_type.to_owned();
+
+        if matches!(return_type, FixedSize::String(_)) {
+            // we need figure out how to deal with dynamically sized returns
+            // for now, this only affects string
```
g-r-a-n-t, line 608: strings

compiler/src/yul/operations/abi.rs
```diff
@@ -53,6 +53,38 @@ pub fn encode_size<T: AbiEncoding>(types: Vec<T>, vals: Vec<yul::Expression>) ->
     return expression! { add([static_size], [data_operations::sum(dyn_size)]) };
 }
 
+/// Returns an expression that gives the max size of the encoded values.
+pub fn static_encode_size<T: AbiEncoding>(types: Vec<T>) -> Result<yul::Expression, String> {
```
g-r-a-n-t, line 57: add tests for this

compiler/src/yul/names.rs
```diff
@@ -39,6 +39,12 @@ pub fn decode_name<T: AbiEncoding>(typ: &T, location: AbiDecodeLocation) -> yul:
     identifier! { (full_name) }
 }
 
+/// Generates an external call function name for a given type and location.
+pub fn contract_call(contract_name: String, func_name: String) -> yul::Identifier {
```
cburgdorf, line 43: It may be slightly more idiomatic (and efficient) to take these as `&str`. But our code base isn't consistent with this so far so I don't actually care that much. At some point when the project gets more mature we'll probably tackle these things. That said, in cases where taking `&str` doesn't complicate anything (e.g. doesn't depend on introducing life time syntax) I'd advocate for always preferring taking a `&str` not a `String`

compiler/src/yul/mappers/expressions.rs
```diff
@@ -103,15 +101,40 @@ pub fn expr_call(
...
+                                    builtins::Method::Clone => expr(context, value),
+                                    builtins::Method::AbiEncode => todo!(),
+                                    builtins::Method::AbiEncodePacked => todo!(),
+                                    builtins::Method::Keccak256 => todo!(),
+                                }
+                            }
+                        };
                     }
+
+                    panic!("invalid attributes")
```
cburgdorf, line 137: Was this deliberate?

analyzer/src/namespace/types.rs
```diff
@@ -568,17 +583,31 @@ impl AbiEncoding for FeString {
     }
 }
 
+impl FeSized for Contract {
+    fn size(&self) -> usize {
+        20
```
cburgdorf, line 588: Could go into a constant such as `ADDRESS_BYTE_LENGTH`

cburgdorf (approved):
Amazing work :tada: 
Left a few comments inline.

newsfragments/204.feature.md
```diff
@@ -0,0 +1,24 @@
...
+
+contract FooProxy:
+    pub def call_build_array(
+        foo_address: address,
+        a: u256,
+        b: u256,
+    ) -> u256[3]:
+        foo: Foo = Foo(foo_address)
+        return foo.build_array(a, b)
+```
```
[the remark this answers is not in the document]
g-r-a-n-t: Yes, we definitely want this in the future.

compiler/src/yul/mappers/expressions.rs
g-r-a-n-t: If the call type is `ValueAttribute` and the function expression doesn't actually consist of an attribute expression, then I think it makes sense to panic with "invalid attributes".

compiler/src/yul/names.rs
g-r-a-n-t: Agreed, too much value passing going on in the codebase.

compiler/src/yul/mappers/expressions.rs
cburgdorf: Sorry, should have been more clear. What I meant was, why not `return CompileError::static_str("invalid attributes")` like we do at these other places. Now this is the only place in the mapper that does a straight panic (not counting `unimplemented!()`)

g-r-a-n-t: I think we need to do some refactoring with error handling in the compiler library. In general, I think result returns should only be used when it's useful to the user. An error here means that the there is a bug in the compiler, so I think we should panic.

I wrote up #168 for this refactor.

PR #206, opened 2021-01-27
Make cli work without solc-backend, add --overwrite option

sbillig:
### What was wrong?

- CLI refused to run without solc-backend feature, despite still being useful (to me, at least).
- CLI lacked --overwrite option, so subsequent runs required an rm -r.

### How was it fixed?

Made the `cfg(feature = solc-backend)` guards in src/main.rs much more fine-grained. If the user tries to emit bytecode without the feature enabled, a warning is printed and the other selected targets are emitted.

Added a simple --overwrite flag. Note that this doesn't blow away the directory or its contents, it just truncates the existing file (if it exists) when writing a new file with the same name.

### To-Do

- [x] Clean up commit history


cburgdorf (commented):
Hey Sean,

thank you very much for the PR. I like those changes! 

Regarding the [failed build](https://github.com/ethereum/fe/pull/206/checks?check_run_id=1774958974) it just looks like you need to run `cargo fmt` to get your code reformatted and then it should pass.

>Added a simple --overwrite flag. Note that this doesn't blow away the directory or its contents, it just truncates the existing file (if it exists) when writing a new file with the same name.

What's the rational for not replacing the files when `--overwrite` is set? If we just append to existing files we'll probably end up with that aren't that useful anymore. E.g. a bytecode file with duplicate bytecode won't be deployable and a YUL file with outputs of multiple runs might be hard to navigate.

Unless you have a good argument in favor I would probably recommend that `--overwrite` should *replace* conflicting files rather than *append* to them.

sbillig:
> Regarding the [failed build](https://github.com/ethereum/fe/pull/206/checks?check_run_id=1774958974) it just looks like you need to run `cargo fmt` to get your code reformatted and then it should pass.

Whoops, my fmt-on-save was using stable rustfmt, which doesn't support some of the settings in rustfmt.toml. Fixed.
 
> What's the rational for not replacing the files when `--overwrite` is set? If we just append to existing files we'll probably end up with that aren't that useful anymore. E.g. a bytecode file with duplicate bytecode won't be deployable and a YUL file with outputs of multiple runs might be hard to navigate.
> 
> Unless you have a good argument in favor I would probably recommend that `--overwrite` should _replace_ conflicting files rather than _append_ to them.

Sorry, that wasn't quite clear. It does replace conflicting files (by truncating them to 0 bytes before writing). The caveat is just that if you, say, run `fe foo.fe` then `fe bar.fe`, the artifacts from the foo compile (without conflicting names) will still exist in the output directory. So it's just that cruft may accumulate, which is a behavior difference from what would happen previously with a manual `rm -r` between runs.

g-r-a-n-t:
> The caveat is just that if you, say, run fe foo.fe then fe bar.fe, the artifacts from the foo compile (without conflicting names) will still exist in the output directory.

We could also have something like: *run `fe --emit abi foo.fe`*, *modify the interface of foo.fe*, and then *run `fe --emit bytecode foo.fe`*, which would result in incompatible files. Not a big issue, but it could warrant overwriting the contract artifacts dir.

g-r-a-n-t (approved):

cburgdorf (approved):

sbillig:
> We could also have something like: _run `fe --emit abi foo.fe`_, _modify the interface of foo.fe_, and then _run `fe --emit bytecode foo.fe`_, which would result in an ABI that is not compatible with the bytecode. Not a big issue, but it could warrant overwriting the entire contract artifacts dir instead of the individual files.

@g-r-a-n-t I considered that, but I'm hesitant to delete a directory.

Eg a contrived dangerous example: `fe home.fe -o /` (where home.fe contains a contract called 'home'). Not sure what the best solution is. Maybe fe could warn the user if there are outdated files for the current contract in the output directory or something.

PR #207, opened 2021-01-28
Renamed the semantics lib to analyzer.

g-r-a-n-t:
This renames the fe-semantics library to fe-analyzer. 

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
Not that I think it is needed for this cahnge but these kind of changes *could* go into a `207.internal.md` to get covered in the *internal* section of the release notes (which is for developers working on the Fe codebase...which isn't a large group yet so feel free to ignore :sweat_smile: )

PR #208, opened 2021-01-28
Add support for chain.id, block.timestamp, other builtin attributes

### LINKED ISSUE #190: chain id and block timestamp
### What is wrong?

There isn't a way for users to get the chain id or block timestamp.

### How can it be fixed

Users should be able to get these values with the expressions `chain.id` and `block.timestamp`.

sbillig:
### What was wrong?

There are some missing builtins needed for the uniswap demo.

### How was it fixed?

~It hasn't been yet :)
I'll probably do block.timestamp, maybe more depending on whether they raise tricky issues.~

Edit: I've added support for:
block.coinbase
block.difficulty
block.number
block.timestamp
chain.id
msg.value
tx.origin
tx.gas_price (aside: do we want to standardize on snake_case like .to_mem()?)

resolves #190

I changed the builtins to use enums, so the compiler can tell us when we aren't handling a case. I also stubbed out some probable builtins in those enums (based on the existing demo fe code, and vyper/solidity). I'm sure there are some things yet to be decided (eg is keccak256 a method on every value, or a standalone builtin function, or imported from the std library, or part of a builtin trait, or...?). I'm happy to change whatever as decisions are made, of course.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/compiler/tests/evm_contracts.rs
+++ b/compiler/tests/evm_contracts.rs
@@ -1,7 +1,4 @@
 #![cfg(feature = "solc-backend")]
-use ethabi;
-use evm;
-
 use evm_runtime::{
     ExitReason,
     Handler,
@@ -45,29 +42,29 @@ impl ContractHarness {
         &self,
         executor: &mut Executor,
         name: &str,
-        input: &Vec<ethabi::Token>,
+        input: &[ethabi::Token],
     ) -> evm::Capture<(evm::ExitReason, Vec<u8>), std::convert::Infallible> {
         let function = &self.abi.functions[name][0];
 
         let context = evm::Context {
-            address: self.address.clone(),
-            caller: self.caller.clone(),
+            address: self.address,
+            caller: self.caller,
             apparent_value: self.value,
         };
 
         let input = function
-            .encode_input(input.as_slice())
+            .encode_input(input)
             .expect("Unable to encode input");
 
-        executor.call(self.address.clone(), None, input, None, false, context)
+        executor.call(self.address, None, input, None, false, context)
     }
 
     pub fn test_function(
         &self,
         executor: &mut Executor,
         name: &str,
-        input: Vec<ethabi::Token>,
-        output: Option<ethabi::Token>,
+        input: &[ethabi::Token],
+        output: Option<&ethabi::Token>,
     ) {
         let function = &self.abi.functions[name][0];
 
@@ -78,7 +75,7 @@ impl ContractHarness {
                         .decode_output(&actual_output)
                         .expect(&format!("unable to decode output: {:?}", &actual_output))[0];
 
-                    assert_eq!(&output, actual_output)
+                    assert_eq!(output, actual_output)
                 }
             }
             evm::Capture::Exit((reason, _)) => panic!("failed to run \"{}\": {:?}", name, reason),
@@ -90,16 +87,16 @@ impl ContractHarness {
         &self,
         executor: &mut Executor,
         name: &str,
-        input: Vec<ethabi::Token>,
+        input: &[ethabi::Token],
     ) {
-        match self.capture_call(executor, name, &input) {
+        match self.capture_call(executor, name, input) {
             evm::Capture::Exit((ExitReason::Revert(_), _)) => {}
             _ => panic!("function did not revert"),
         }
     }
 
     // Executor must be passed by value to get emitted events.
-    pub fn events_emitted(&self, executor: Executor, events: Vec<(&str, Vec<ethabi::Token>)>) {
+    pub fn events_emitted(&self, executor: Executor, events: &[(&str, &[ethabi::Token])]) {
         let raw_logs = executor
             .deconstruct()
             .1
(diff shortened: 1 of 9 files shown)


g-r-a-n-t:
The refactoring looks good.

> is keccak256 a method on every value, or a standalone builtin function, or imported from the std library, or part of a builtin trait, or...?

For now I'm thinking keccak256 and the ABI encoding functions will be builtin to certain types as attribute functions, similar to `to_mem` and `clone`. Ideally these functions will be implemented in the standard library later on. 

Also, we probably want to limit the use of keccak256 to `bytes[n]`. This way, it's what you're what hashing. So anytime `keccak256` is used, you would first encode the value and then hash (e.g. `(42, "hello world").abiEncode().keccak256()`).

g-r-a-n-t:
Here's the [Yul reference](https://docs.soliditylang.org/en/latest/yul.html#evm-dialect) btw.

sbillig:
Ok, this is probably enough for a first pass. msg.data and msg.sig aren't simple yul function calls, so I'm going to wait on those and maybe start looking at something more pressing.

The 'use refs in evm_contracts tests' changes are mostly superfluous. I hit a minor ownership issue and thought the nicest solution was for `test_function`s output arg to be `Option<&Token>` instead of `Option<Token>`, and ended up replacing all the vecs with `&[]` too. Happy to revert if such changes are annoying to anyone.

semantics/src/traversal/expressions.rs
```diff
@@ -611,12 +665,22 @@ fn expr_attribute_call_type(
...
-                func_name: attr.node.to_string(),
-            }),
-            _ => Ok(CallType::ValueAttribute),
+        if let fe::Expr::Name(name) = value.node {
+            use builtins::Object;
+            match Object::from_str(name) {
+                Ok(Object::Block) => todo!(),
+                Ok(Object::Chain) => todo!(),
+                Ok(Object::Msg) => todo!(), // TODO: error because msg has no methods?
+                Ok(Object::Tx) => todo!(),
```
cburgdorf, line 674: TIL: https://doc.rust-lang.org/std/macro.todo.html

semantics/src/builtins.rs
```diff
@@ -1,5 +1,53 @@
-pub const SELF: &str = "self";
-pub const SENDER: &str = "sender";
-pub const MSG: &str = "msg";
-pub const CLONE: &str = "clone";
-pub const TO_MEM: &str = "to_mem";
+use strum::EnumString;
```
cburgdorf, line 1: Oh, that's cool! I didn't know about `EnumString` and yeah it makes total sense to use enums here for added guidance from the compiler :tada:

semantics/src/traversal/expressions.rs
sbillig: One of my favorite recent additions!

semantics/src/traversal/expressions.rs
```diff
@@ -611,12 +665,22 @@ fn expr_attribute_call_type(
...
-            fe::Expr::Name(builtins::SELF) => Ok(CallType::SelfAttribute {
-                func_name: attr.node.to_string(),
-            }),
-            _ => Ok(CallType::ValueAttribute),
+        if let fe::Expr::Name(name) = value.node {
+            use builtins::Object;
+            match Object::from_str(name) {
+                Ok(Object::Block) => todo!(),
+                Ok(Object::Chain) => todo!(),
+                Ok(Object::Msg) => todo!(), // TODO: error because msg has no methods?
```
g-r-a-n-t, line 673: Yeah, we could use the `UndefinedValue` error here or create something more specific like `UndefinedFunction`.

sbillig: I made these UndefinedValues for now.

sbillig:
Rebased on master; should be good now. If I add `.abi_encode()`, etc I'll do it in a new PR.

compiler/tests/evm_contracts.rs
```diff
@@ -124,7 +123,7 @@ impl ContractHarness {
                 })
                 .collect::<Vec<_>>();
 
-            if !outputs_for_event.contains(&expected_output) {
+            if !outputs_for_event.iter().any(|v| &v == expected_output) {
```
cburgdorf, line 126: No big deal but we generally try to avoid single letter variables. This could be named `val` instead.

compiler/tests/evm_contracts.rs
```diff
@@ -224,12 +227,12 @@ fn bytes_token(s: &str) -> ethabi::Token {
     ethabi::Token::FixedBytes(ethabi::FixedBytes::from(s))
 }
 
-fn u256_array_token(v: Vec<usize>) -> ethabi::Token {
-    ethabi::Token::FixedArray(v.into_iter().map(|n| uint_token(n)).collect())
+fn u256_array_token(v: &[usize]) -> ethabi::Token {
```
cburgdorf, line 230: But we still have some leftover single letter variable names today...we should adjust all of them (but doesn't have to be in this PR)

cburgdorf (approved):
This looks great! All the `Vec<T>` to &[T]` changes seem to indicate that we aren't running clippy against the tests. Something that we should probably do.

g-r-a-n-t:
very nice!

g-r-a-n-t (approved):

Issue #213, opened 2021-01-29
byte index 6 is out of bounds of `ir56` at compiler/src/abi/builder.rs:181:41


agroce:
Compiling this file:

```
contract Foo:
    pub def bar(x: i256, y: ir56) -> i256:
        return x - y
```

```
Warning: bytecode output requires 'solc-backend' feature. Try `cargo build --release --features solc-backend`. Skipping.
thread 'main' panicked at 'byte index 6 is out of bounds of `ir56`', compiler/src/abi/builder.rs:181:41
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/core/src/panicking.rs:92:14
   2: core::str::slice_error_fail
   3: fe_compiler::abi::builder::type_desc
   4: fe_compiler::abi::builder::func_def_arg
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted
```

On latest github master, built with

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer (we've found more than 50 bugs (that were worth fixing, and have been fixed) in solc using this afl variant) + afl.rs.

This might be a more interesting type of bug, if you think fe is ready for fuzzing at all.

cburgdorf:
Fixed by #217

PR #217, opened 2021-01-30
Fix crash in abi builder on unknown type with short name

sbillig:
### What was wrong?

The abi builder crashed on an unknown type (with a short name) in a function signature. Eg:
```
contract F:
  pub def f(x: foo) -> bool:
    return true
```

This presumably should have been caught by the analyzer; I haven't looked into how it slipped through the cracks.

### How was it fixed?

changed `if &base[..6] == "string"` to `if base.starts_with("string")`. Other similar tests all use `starts_with` already, btw.

### To-Do

- [ ] Clean up commit history


sbillig:
Oh, this resolves #213. I happened upon it when I misspelled u256 uint.

cburgdorf (approved):

PR #231, opened 2021-02-03
Add get_type_defs/get_module_type_defs to scope

cburgdorf:
### What was wrong?

Nothing wrong but a fall out from #203. This adds convenience helper to the contract and module scopes to filter for a list of type defintions.

### How was it fixed?

Basically took some existing code and put it in a function + wrangling 20 minutes about the right syntax for the closure.


g-r-a-n-t (approved):
looks great

PR #239, opened 2021-02-08
Basic support for `create` and `create2`

g-r-a-n-t:
### What was wrong?

There was no support for contract creation.

### How was it fixed?

- Built out support for type attribute calls and added `create` functions to the contract type.
- The analyzer now records contract types that are created and the compiler adds their Yul objects to the contract's runtime.
- Added a compiler flag for Yul optimization.

### To-Do

- [x] Update [Spec](https://github.com/ethereum/fe/blob/master/spec/index.md) if applicable

- [x] Clean up commit history


compiler/src/yul/runtime/functions/contract.rs
```diff
@@ -82,3 +82,13 @@ pub fn contract_calls(contract: Contract) -> Vec<yul::Statement> {
         })
         .collect()
 }
+
+pub fn create2_contract() -> yul::Statement {
+    function_definition! {
+        function create2_contract(code_start, code_size, value, salt) -> return_address {
```
g-r-a-n-t, line 88: better param names

compiler/src/yul/runtime/mod.rs
```diff
@@ -81,7 +81,7 @@ pub fn build(context: &Context, contract: &Spanned<fe::ModuleStmt>) -> Vec<yul::
             attributes
                 .external_contracts
                 .iter()
-                .map(|contract| functions::calls::contract_calls(contract.to_owned()))
+                .map(|contract| functions::contract::contract_calls(contract.to_owned()))
```
g-r-a-n-t, line 84: rename this function

compiler/tests/fixtures/create2_contract.fe
```diff
@@ -0,0 +1,9 @@
+contract Foo:
```
g-r-a-n-t, line 1: add a test for create

analyzer/src/namespace/scopes.rs
```diff
@@ -295,6 +299,15 @@ impl BlockScope {
         }
         true
     }
+
+    /// Filter module scope for type definitions that match the given predicate
+    pub fn get_module_type_defs<B, F: FnMut(&Type) -> Option<B>>(&self, predicate: F) -> Vec<B> {
+        self.module_scope().borrow().get_type_defs(predicate)
+    }
+
+    pub fn get_module_type_def(&self, name: &str) -> Option<Type> {
```
g-r-a-n-t, line 328: docs

compiler/src/yul/mappers/contracts.rs
```diff
@@ -55,24 +59,28 @@ pub fn contract_def(
...
+                        (return(0, 0))
                     },
                 },
-                objects: vec![],
-                // We can't reach to data objects in the "contract" hierachy so in order to have
-                // the data objects available in both places we have to put them in both places.
-                data: data.clone(),
-            }],
+            },
+            objects: external_contracts,
```
g-r-a-n-t, line 74: these contract objects are not optimized out by the yul compiler, so they should only be added if `create` or `create2` are used

compiler/tests/evm_contracts.rs
```diff
@@ -1256,3 +1279,21 @@ fn external_contract() {
         foo_harness.events_emitted(executor, &[("MyEvent", &[my_num, my_addrs, my_string])]);
     })
 }
+
+#[test]
```
g-r-a-n-t, line 1291: add a test to make sure using different salts results in different addresses

analyzer/src/traversal/expressions.rs
```diff
@@ -660,6 +632,63 @@ fn expr_call_value_attribute(
...
+    scope: Shared<BlockScope>,
+    context: Shared<Context>,
+    typ: Type,
+    func_name: String,
+    args: &Spanned<Vec<Spanned<fe::CallArg>>>,
+) -> Result<ExpressionAttributes, SemanticError> {
+    let arg_attributes = expr_call_args(Rc::clone(&scope), context, args)?;
+
+    match (typ, func_name.as_str()) {
+        (Type::Contract(contract), "create2") => {
```
g-r-a-n-t, line 645: use builtins

analyzer/src/traversal/expressions.rs
```diff
@@ -660,6 +632,63 @@ fn expr_call_value_attribute(
...
+
+                Ok(ExpressionAttributes::new(
+                    Type::Contract(contract),
+                    Location::Value,
+                ))
+            } else {
+                Err(SemanticError::type_error())
+            }
+        }
+        (Type::Contract(contract), "create") => {
```
g-r-a-n-t, line 668: use builtins

compiler/src/yul/mappers/expressions.rs
```diff
@@ -145,6 +143,18 @@ fn expr_call(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressi
 
                     panic!("invalid attributes")
                 }
+                CallType::TypeAttribute { typ, func_name } => match (typ, func_name.as_str()) {
+                    (Type::Contract(contract), "create2") => Ok(contract_operations::create2(
```
g-r-a-n-t, line 147: builtins

compiler/src/yul/mappers/expressions.rs
```diff
@@ -145,6 +143,18 @@ fn expr_call(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressi
...
+                        &contract,
+                        yul_args[0].to_owned(),
+                        yul_args[1].to_owned(),
+                    )),
+                    (Type::Contract(contract), "create") => Ok(contract_operations::create(
```
g-r-a-n-t, line 152: builtins

analyzer/src/builtins.rs
```diff
@@ -2,14 +2,21 @@ use strum::EnumString;
...
     Clone,
     ToMem,
     Keccak256,
     AbiEncode,
     AbiEncodePacked,
 }
 
+#[derive(Debug, PartialEq, EnumString)]
+#[strum(serialize_all = "snake_case")]
+pub enum TypeMethod {
```
cburgdorf, line 15: Maybe even `ContractTypeMethod` as those would only be used there.

compiler/tests/fixtures/create2_contract.fe
```diff
@@ -0,0 +1,9 @@
...
+    pub def get_my_num() -> u256:
+        return 42
+
+contract FooFactory:
+    pub def create2_foo() -> address:
+        # value and salt
+        foo: Foo = Foo.create2(0, 52)
```
cburgdorf, line 8: I created an issue for the proposed shift to double colon syntax (`Foo::create2(0, 52)`) https://github.com/ethereum/fe/issues/241

spec/index.md
```diff
@@ -526,7 +526,26 @@ x = true
 
 ### 5.1.1.2. Contract type
 
-An *contract type* is the type denotated by the name of an [`contract` item].
+An *contract type* is the type denoted by the name of an [`contract` item].
+
+A value of a given contract type caries the contract's public interface as
```
cburgdorf, line 531: caries -> carries

spec/index.md
```diff
@@ -526,7 +526,26 @@ x = true
...
+```python 
+ contract Foo:
+    pub def get_my_num() -> u256:
+        return 42
+
+contract FooFactory:
+    pub def create2_foo() -> address:
+        # `0` is the value being sent and `52` is the address salt
+        foo: Foo = Foo.create2(0, 52)
+        return address(foo)
```
cburgdorf, line 547: Should we consider giving contracts a special attribute to obtain the address such so that the above could be written as:

```
pub def create2_foo() -> address:
        # `0` is the value being sent and `52` is the address salt
        foo: Foo = Foo.create2(0, 52)
        return foo.__address__
```

With the cast, a bug such as the following would be possible that would result in returning the wrong address.

```
pub def create2_foo() -> address:
        # `0` is the value being sent and `52` is the address salt
        foo: Foo = Foo.create2(0, 52)
        bar: Bar = Bar.create2(0, 11)
        return address(bar)
```

In the example above, we *mean* to return the address of the newly created `foo` but we accidentally return the address of `bar` instead. Sure, one can also accidentially return `bar.__address__` but it feels to me that having the address accessible as a property removes one layer of error potential. That said, I really dislike the `__address__` syntax but wouldn't now a good way to get around the potential for naming conflicts. At least not until we have something like traits that we could bring into scope conditionally to expose special functions/attributes.

cburgdorf (approved):
Looks good! Left some comments inline.

compiler/tests/fixtures/create2_contract.fe
g-r-a-n-t: great! thank you

spec/index.md
g-r-a-n-t: Yeah, I wasn't sure how to deal with this.

We can discuss this more in #179 and if we come up with a better idea, it can be changed. For now, I'm inclined to just merge this as is.

cburgdorf: Yeah, should be good for now. :+1:

PR #242, opened 2021-02-10
Use Into<String> instead of String for scope APIs

cburgdorf:
@g-r-a-n-t I've been thinking about this change and wanted to get your opinion on it. Basically, for all the scope APIs that currently take `String` as a parameter I changed them to take `Into<String>`. The benefit is that callsites get slightly cleaner because both `&str` and `String` will be accepted. The downside is that the generic syntax makes the function definition slightly more verbose.

Another alternative to consider would be to change all the function parameters to `&str` instead which might actually be a good balance because it looks less bloaty compared to the generic syntax and still gets rid of all the `to_string()`, `clone()`, `to_owned()` at the callsites (while sometimes demanding an additional `&` if we are coming from a `String`. This is also what's  [recommended here](https://users.rust-lang.org/t/idiomatic-string-parmeter-types-str-vs-asref-str-vs-into-string/7934/7).

Interested to hear your opinion about this.

### DIFF
--- a/analyzer/src/namespace/scopes.rs
+++ b/analyzer/src/namespace/scopes.rs
@@ -91,8 +91,8 @@ impl ModuleScope {
     }
 
     /// Add a type definiton to the scope
-    pub fn add_type_def(&mut self, name: String, typ: Type) {
-        self.type_defs.insert(name, typ);
+    pub fn add_type_def<T: Into<String>>(&mut self, name: T, typ: Type) {
+        self.type_defs.insert(name.into(), typ);
     }
 
     /// Filter module scope for type definitions that match the given predicate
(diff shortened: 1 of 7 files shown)


analyzer/src/namespace/scopes.rs
```diff
@@ -101,8 +101,8 @@ impl ModuleScope {
     }
 
     /// Gets a type definition by name.
-    pub fn get_type_def(&self, name: &str) -> Option<Type> {
```
cburgdorf, line 104: This was just changed for consistency in the PR but the alternative that I proposed in the PR description would actually be exactly this, taking`&str` everywhere. Maybe that would be better after all?

g-r-a-n-t: I think this is the way we should be doing it. What you already have is an improvement, but what you're suggesting here is probably the most rusty approach.

cburgdorf: Agreed! Will open up a new PR

PR #243, opened 2021-02-10
Runtime testing

### LINKED ISSUE #214: Contract runtime testing.
### What is wrong?

We're currently limited to only testing the public interface of a contract. This can make debugging features that rely on multiple new runtime functions difficult. We should be able to test the runtime directly within our contracts test suite.

### How can it be fixed

Create a new testing function that takes the following: an executor, contract fixture path, contract name, and a set of Yul statements to inject into the runtime.

It should look something like this:

```rust
test_runtime(
  &executor,
  "my_module.fe",
  "MyContract",
  statements! {
    (if (eq(add(1, 1), 2)) {
    (revert(0, 0))
  }
)
```

What we ultimately want to do is build a Yul object similar to an actual contract, but it only contains the runtime (no constructor) and the test statements. We would then compile this object to bytecode and execute it. If we hit a revert, the test has failed.

We should be able to create a function similar to [`runtime::build_with_abi_dispatcher`](https://github.com/ethereum/fe/blob/4750ab803e513255a7cbb702e3795a554ee66481/compiler/src/yul/runtime/mod.rs#L69), but named `build_with_test_statements` that simply inserts test statements into the runtime object. We then compile and execute.

g-r-a-n-t:
### What was wrong?

closes #214 

### How was it fixed?

The solution I used was a bit different than the one proposed in the linked issue. Instead of testing functions in the runtime object of a given contract, the functions are brought in directly from the runtime builder and tested. This seemed like a much more straight forward way of doing things.

- reorganized the test directory
- add a new testing utility function `test_runtime_functions` that takes a vector of functions to be tested and a vector of statements to test those functions
- added a handy `assert_eq` macro that builds an if statement that enters and reverts if the assertion is false
- used the utilities described above to test some standard runtime functions

### To-Do

- [ ] Get to the bottom of the apparent `sloadn` bug

- [x] Clean up commit history


compiler/src/yul/runtime/mod.rs
```diff
@@ -125,3 +125,14 @@ pub fn build_with_abi_dispatcher(
 
     panic!("missing contract attributes")
 }
+
+pub fn build_with_test_statements(
```
g-r-a-n-t, line 129: delete this

compiler/src/evm/mod.rs
```diff
@@ -16,7 +16,7 @@ pub fn compile(mut contracts: NamedYulContracts) -> Result<NamedBytecodeContract
         .collect::<Result<NamedBytecodeContracts, _>>()
 }
 
-fn compile_single_contract(yul_src: YulIr) -> Result<Bytecode, CompileError> {
+pub fn compile_single_contract(yul_src: YulIr) -> Result<Bytecode, CompileError> {
```
g-r-a-n-t, line 19: docs

compiler/tests/evm_runtime.rs
```diff
@@ -0,0 +1,170 @@
...
+        test_runtime_functions(
+            &mut executor,
+            functions::std(),
+            statements! {
+                (let a := 0x4200000000000000000000000000000000000000000000000000000000420026)
+                (let b := 0x0000000000000000000000000000000000000000000000000000000000000026)
+                (sstore(100, a))
+
+                // this fails for some reason.. possible bug?
+                // [assert_eq!(b, (sloadn(130, 2)))]
```
g-r-a-n-t, line 111: it looks like this might be a bug @cburgdorf. I could be missing something tho

g-r-a-n-t:
still a few todo items to check off, but this should be good for a review

compiler/tests/evm_runtime.rs
cburgdorf: This is interesting. You can prove `[assert_eq!(0x42, (sloadn(100, 1)))]` but that's as far as I got. I tried changing the second byte to anything but `00` and then use `sloadn(101, 1)` but that already fails. So, maybe a base assumption about how storage works is wrong? :sweat_smile:

cburgdorf (approved):
Really cool! I played a bit with the code and tried to find an answer to your inline comment..it taught me a bit but I had no luck getting to the bottom of it.

compiler/tests/evm_runtime.rs
```diff
@@ -0,0 +1,170 @@
...
+            statements! {
+                (let a := 0x4200000000000000000000000000000000000000000000000000000000420026)
+                (let b := 0x0000000000000000000000000000000000000000000000000000000000000026)
+                (mstore(100, a))
+
+                [assert_eq!(b, (mloadn(130, 2)))]
+            },
+        );
+    })
+}
```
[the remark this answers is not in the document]
g-r-a-n-t: added a few more assertions like this

compiler/tests/evm_runtime.rs
g-r-a-n-t: I'm going to merge this and continue digging into the storage stuff. Pretty cool that we already found an issue with this testing.

PR #244, opened 2021-02-11
Refactor string usage to be aligned across code base and make fewer allocations

cburgdorf:
### What was wrong?

1. We had a mix of using `String` or `&str` for function parameters with no clear guidance on when to use which
2. We were using `String` for function parameters resulting in:
    a.) Lots of verbose `to_string()`, `clone()` usage
    b.) More memory allocations than necessary

### How was it fixed?

This PR changes the code base to follow a simple rule which may not always result in fewest allocations but has seems to have reduced the overall amount of allocations while at the same time cleaning up a lot of verbosity and being very simple. The rule that was followed is to simple always use `&str` as function parameters and use `to_owned()` wherever we need to convert from a string slice to an allocated `String`.

### DIFF
--- a/analyzer/src/namespace/events.rs
+++ b/analyzer/src/namespace/events.rs
@@ -12,7 +12,7 @@ pub struct Event {
 }
 
 impl Event {
-    pub fn new(name: String, fields: Vec<FixedSize>, indexed_fields: Vec<usize>) -> Self {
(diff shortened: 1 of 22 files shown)


g-r-a-n-t (approved):
great refactor :smile:

PR #246, opened 2021-02-13
Support updating struct fields in storage

cburgdorf:
### What was wrong?

We did not support updating individual struct fields in storage (e.g. `self.my_struct.some_field = 1`)

### How was it fixed?

1. Make the traversal recognize nested attribute expressions
2. Map them to carry the location of the entire struct
3. Map field of storage struct to a pointer of `ptr_of_storage_struct + 32 * field_index`
4. Add test

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -369,6 +369,26 @@ fn expr_attribute(
             TxField,
         };
 
+        if let fe::Expr::Attribute { .. } = &value.node {
+            match context.get_expression(value) {
+                Some(ExpressionAttributes {
+                    location,
+                    typ: Type::Struct(val),
+                    ..
+                }) => match val.get_field_index(attr.node) {
+                    Some(index) => {
+                        if let Location::Storage { nonce: Some(nonce) } = location {
+                            return Ok(nonce_with_offset_to_ptr(*nonce, index * 32));
+                        } else {
+                            return Err(CompileError::static_str("invalid attributes"));
+                        };
+                    }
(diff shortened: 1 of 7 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -444,6 +464,14 @@ pub fn nonce_to_ptr(nonce: usize) -> yul::Expression {
     literal_expression! { (ptr) }
 }
 
+/// Converts a storage nonce into a pointer based on the keccak256 hash
+pub fn nonce_with_offset_to_ptr(nonce: usize, offset: usize) -> yul::Expression {
+    let ptr = get_full_signature(nonce.to_string().as_bytes());
+    let ptr = literal_expression! { (ptr) };
+    let offset = literal_expression! { (offset) };
+    expression! { (add([ptr], [offset])) }
```
cburgdorf, line 472: @g-r-a-n-t Not sure if this is still the right thing to do after your recent findings about how to deal with storage.  

But if it is, we should probably calculate the `ptr` manually to get rid of the `add(ptr, offset)`. I thought that this is something that the optimizer would do for us but I ran with `--optimize` and it isn't optimizing the `add` away.

g-r-a-n-t: this is fine for now

g-r-a-n-t (approved):

compiler/src/yul/mappers/expressions.rs
cburgdorf: :+1: I will create an issue to track this optimization though

Issue #219, opened 2021-01-31
Yul compilation failed: ParserError: Switch statement without any cases.


agroce:
Compiling this file:

```
contract Foo:
    my_u128: u128
```

```
thread 'main' panicked at 'Yul compilation failed: {"errors":[{"component":"general","formattedMessage":"ParserError: Switch statement without any cases.\n --> input.yul:1:3990:\n  |\n1 |  ... ta_size)) } } switch cloadn(0, 4)  }   }  }\n  |                                         ^\n\n","message":"Switch statement without any cases.","severity":"error","sourceLocation":{"end":3990,"file":"input.yul","start":3989},"type":"ParserError"},{"component":"general","formattedMessage":"ParserError: Expected end of source but got '}'\n --> input.yul:1:3997:\n  |\n1 |  ... )) } } switch cloadn(0, 4)  }   }  }\n  |                                         ^\n\n","message":"Expected end of source but got '}'","severity":"error","sourceLocation":{"end":3997,"file":"input.yul","start":3996},"type":"ParserError"}]}', compiler/src/lib.rs:45:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/std/src/panicking.rs:435:5
   2: fe_compiler::compile
   3: fe::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted
```

On latest github master, built with

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer (we've found more than 50 bugs (that were worth fixing, and have been fixed) in solc using this afl variant) + afl.rs.

The PR works nicely!

cburgdorf:
Ah yeah, that's a good one. I ran into it manually recently. Contracts without any methods trigger that error.

PR #248, opened 2021-02-15
Fix bug where compilation of contracts without public functions would result in malformed YUL

cburgdorf:
### What was wrong?

As identified by the fuzzer in #219 contracts without any public functions compile to faulty YUL.

### How was it fixed?

Compile to `pop(0)` instead of `switch (..)` when there are no public functions.

### To-Do


g-r-a-n-t (approved):

output/Empty/Empty_ir.yul
```diff
@@ -0,0 +1,168 @@
+object \"Empty\" {
```
g-r-a-n-t, line 1: :wastebasket:

cburgdorf: oops

Issue #188, opened 2021-01-15
keccak256


g-r-a-n-t:
### What is wrong?

Users cannot perform `keccak256` on values (e.g. `my_num.keccak256()`).

### How can it be fixed

An attribute function `keccak256` should be implemented on all data types.

cburgdorf:
I can see the convenience of being able to call `keccak256()` on any variable. But I think I'd prefer having it as a standalone function because:

1. Planning ahead, this may not be the only cryptographic function that we need (e.g. [see](https://docs.soliditylang.org/en/v0.4.21/units-and-global-variables.html#mathematical-and-cryptographic-functions))
2. In the future `keccak256` usage my start to decline leaving us with "API pollution"
3. We can start having `keccak256` globally available and eventually move it behind an import when we support them (e.g. `from crypto import keccak256`). This will be a less harsh breaking change compared to removing a builtin on all types.

In general, my stance is to be extra cautious with builtin methods that we provide on a broad set of types. I think that importing is generally more future proof and provides a cleaner way to evolve the language. I think that's also a general stance that the rust team uses as guidance.

cburgdorf:
This landed with #255

PR #255, opened 2021-02-16
Implement keccak256 as builtin method

cburgdorf:
### What was wrong?

As discussed in #188. Fe needs to expose the `keccak256` hash function to be usable from user code.

### How was it fixed?

1. Implemented `keccak256` as a global function
2. Analyzer ensures it can only be called with `bytes[n]` parameter
3. Mapper is pretty straight forward
4. keccak utils were slightly refactored to expose a bare `get_keccak256` method
5. Added contract tests to calculate various keccaks
6. Added test to prove `keccak256` can not be called with arrays of other types
7. Added bare runtime tests for the YUL `keccak256` function

### To-Do

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -136,7 +136,6 @@ fn expr_call(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressi
                                     builtins::ValueMethod::Clone => expr(context, value),
                                     builtins::ValueMethod::AbiEncode => todo!(),
                                     builtins::ValueMethod::AbiEncodePacked => todo!(),
-                                    builtins::ValueMethod::Keccak256 => todo!(),
                                 }
                             }
                         };
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -6,8 +6,12 @@ use crate::yul::operations::{
     structs as struct_operations,
 };
 use crate::yul::utils;
-use fe_analyzer::builtins::ContractTypeMethod;
+use fe_analyzer::builtins::{
+    ContractTypeMethod,
+    GlobalMethod,
+};
 use fe_analyzer::namespace::types::{
+    FeSized,
     FixedSize,
     Type,
 };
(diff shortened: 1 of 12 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -105,6 +112,20 @@ fn expr_call(context: &Context, exp: &Spanned<fe::Expr>) -> Result<yul::Expressi
...
+                        let first_arg = args.node.first().expect("Missing argument");
+                        let arg_expr = context
+                            .get_expression(first_arg)
+                            .expect("invalid attributes");
+                        let size = to_fixed_size(&arg_expr.typ)?;
+                        let func_name: &str = func.into();
+
+                        let func_name = identifier! { (func_name) };
+                        let size = identifier_expression! { (size.size()) };
+                        Ok(expression! { [func_name]([yul_args[0].to_owned()], [size]) })
```
cburgdorf, line 126: https://docs.soliditylang.org/en/latest/yul.html#simple-example

IIUC `keccak256(p, n)` is expecting a pointer to memory (`keccak(mem[p…(p+n)))`) so calling it with a stack value like `keccak256(1)` doesn't do what it is supposed to. It would return the keccak over the 32 bytes starting at memory address `1` which is absolutely unrelated to the value `1`.

I think what we should probably do is to validate that thergument to `keccak256` is always of location `Memory`. So valid examples would be `keccak256(mem_var)` or `keccak256(storage_var.to_mem())` and probably `keccak256(abi_encode_packed(1))` (because the latter would turn `1` into a memory pointer where the value is encoded.

compiler/src/yul/mappers/expressions.rs
```diff
@@ -67,8 +71,7 @@ fn move_expression(
     from: Location,
     to: Location,
 ) -> Result<yul::Expression, CompileError> {
-    let typ =
-        FixedSize::try_from(typ).map_err(|_| CompileError::static_str("invalid attributes"))?;
+    let typ = to_fixed_size(&typ)?;
```
g-r-a-n-t, line 74: I think we should keep the original line with an `expect` instead of `map_err`, so we crash in this situation. All the results should be removed in this module.

g-r-a-n-t (approved):
looks good

PR #259, opened 2021-02-18
Add test that demos sqrt and min function needed for the Uniswap demo

cburgdorf:
### What was wrong?

We need a Fe implementation of `sqrt` and `min` for the Uniswap demo.

### How was it fixed?

Added a fixture `math.fe` that implements those.


g-r-a-n-t (approved):
LGTM

Issue #189, opened 2021-01-15
Support for my_val.abi_encode() and my_val.abi_encode_packed()


g-r-a-n-t:
### What is wrong?

Users cannot encode values.

### How can it be fixed

We should support the attribute functions `abi_encode_packed` and `abi_encode`, which would return bytes.

The size of these bytes might be a bit tricky, since we don't support dynamically sized byte arrays.

g-r-a-n-t:
abi packed stuff can now be found in #281

PR #260, opened 2021-02-19
Require KWargs for struct initialization

cburgdorf:
### What was wrong?

Example:

```
struct House:
    vacant: bool
    price: u256
```

Currently, `House` can be instantiated as `House(true, 1000000)`.

This means:

1. Property assignment can easily get mixed up leading to bugs
2. struct initialization is visually very close to regular function invocations

### How was it fixed?

With this change it is required to be instantiated like `House(vacant=true, price=1000000)`

While working on this it got me thinking if this is something that should be handled on the parser stage instead but I don't think that would be straight forward because technically there is nothing the syntax is identical with method invocation syntax.

We **could** move this into the parser stage **if** we used a distinct syntax such as `House(vacant: true, price: 100)`. But I'm personally ok to keep it as with the check being handled by the analyzer.

### DIFF
--- a/analyzer/src/errors.rs
+++ b/analyzer/src/errors.rs
@@ -8,6 +8,7 @@ use fe_parser::span::Span;
 pub enum ErrorKind {
     BreakWithoutLoop,
     ContinueWithoutLoop,
+    KeyWordArgsRequired,
     MissingReturn,
     NotSubscriptable,
     NumericCapacityMismatch,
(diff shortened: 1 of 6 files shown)


g-r-a-n-t (approved):

g-r-a-n-t:
Something I noticed while working on #189 is that initialization doesn't work much of the time, since the values are in an unpredictable order upon retrieval.

The following will fail to compile for me:

```
struct House:
    price: u256
    size: u256
    rooms: u8
    vacant: bool

contract Foo:
    pub def create_house():
        self.my_house = House(1, 2, u8(5), false)
``` 

as the attribute types just so happen to be retrieved in the the order `u256, u8, u256, bool`.

I'll fix this in my PR.

g-r-a-n-t:
fixed. I just added another attribute for key order to the struct struct

PR #263, opened 2021-02-22
Ordered struct fields.

g-r-a-n-t:
### What was wrong?

The expanded test was failing with a type error on struct initialization since the values are retrieved in an effectively random order.

### How was it fixed?

Added a vector to keep track of field order.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/namespace/types.rs
+++ b/analyzer/src/namespace/types.rs
@@ -39,7 +39,6 @@ pub trait FeSized {
 /// needed by our encoding/decoding functions to properly read and write data.
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
 pub struct AbiUintSize {
-    ///
     pub data_size: usize,
     pub padded_size: usize,
 }
@@ -164,6 +163,7 @@ pub struct Tuple {
 pub struct Struct {
     pub name: String,
     fields: BTreeMap<String, Base>,
+    order: Vec<String>,
 }
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Ord, Eq)]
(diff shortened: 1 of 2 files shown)


compiler/tests/fixtures/structs.fe
```diff
@@ -1,40 +1,55 @@
...
+        assert self.my_house.price == 1000
+        assert self.my_house.rooms == u8(100)
+        assert self.my_house.vacant
 
     pub def bar() -> u256:
-        building: House = House(price=300, size=500, vacant=true)
+        building: House = House(price=300, size=500, rooms=u8(20), vacant=true)
         assert building.size == 500
         assert building.price == 300
+        #assert building.rooms == u8(20)
```
g-r-a-n-t, line 43: this fails when uncommented. looking into it as well

g-r-a-n-t:
I wanted to open this early, so we could get into the release.

newsfragments/263.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Struct initialization would often fail with a type error since the data structure 
+responsible for managing struct fields does not enforce order.
+
+Order is now accounted for separately.
```
cburgdorf, line 4: We don't need to put a release notes entry for this bugfix since the bug didn't make it into a release

analyzer/src/namespace/types.rs
```diff
@@ -182,37 +182,48 @@ impl Struct {
         Struct {
             name: name.to_string(),
             fields: BTreeMap::new(),
+            order: vec![],
```
cburgdorf, line 185: I'm surprised this is needed. The reason I chose `BTreeMap` was because I thought it is ordered. I guess if that doesn't hold we could also change it back to a regular `HashMap`.

cburgdorf (approved):

analyzer/src/namespace/types.rs
g-r-a-n-t: They are ordered, as in, they implement `Ord`. However, this only means that the map objects themselves can be compared and sorted. Regular hashmaps don't implement `Ord` and therefore can't be compared or sorted. Being able to sort types is actually required at the moment for dispatching ABI runtime functions, since we want to remove any duplicates. Removal of duplicates is accomplished by first sorting the a collection of parameter types and the de-duplicating the collection.

PR #264, opened 2021-02-22
Support boolean not operator

cburgdorf:
### What was wrong?

Fe doesn't yet support the boolean `not` operator. E.g. the following code doesn't compile:

```
if not covid_test.is_positive(person):
    allow_boarding(person)
```

### How was it fixed?

1. Analyzer validates `UnaryOperator::Not` is used with `bool` only
2. Mapper maps it to `iszero([operand])`
3. Tests

### DIFF
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -425,21 +425,32 @@ fn expr_unary_operation(
     exp: &Spanned<fe::Expr>,
 ) -> Result<ExpressionAttributes, SemanticError> {
     if let fe::Expr::UnaryOperation { op, operand } = &exp.node {
-        if let fe::UnaryOperator::USub = &op.node {
-            let operand_attributes = value_expr(Rc::clone(&scope), Rc::clone(&context), operand)?;
+        let operand_attributes = value_expr(Rc::clone(&scope), Rc::clone(&context), operand)?;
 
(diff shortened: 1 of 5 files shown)


g-r-a-n-t (approved):

analyzer/src/traversal/expressions.rs
```diff
@@ -425,21 +425,31 @@ fn expr_unary_operation(
...
+                ));
             }
-            // No matter what numeric type the operand was before, the minus symbol turns it
-            // into an I256 just like all positive values default to U256.
-            return Ok(ExpressionAttributes::new(
-                Type::Base(Base::Numeric(Integer::I256)),
-                Location::Value,
-            ));
+            fe::UnaryOperator::Not => {
+                if !matches!(operand_attributes.typ, Type::Base(Base::Bool)) {
```
g-r-a-n-t, line 443: this could also be expressed as

```rust
return if !matches!(operand_attributes.typ, Type::Base(Base::Bool)) {
  Err(SemanticError::type_error())
} else {
  Ok(ExpressionAttributes::new(
    Type::Base(Base::Bool),
    Location::Value,
  ))
};
```

which I personally find to be more idiomatic

cburgdorf: Good point :+1:

PR #265, opened 2021-02-22
Do over/underflow checks for addition (SafeMath)

cburgdorf:
### What was wrong?

As explained in #153 artihmetic operations should check for under/overflows by default and revert if such are detected.

### How was it fixed?

This adds checks for all additions of both signed and unsigned integers. 

1. Added `checked_add_u256` and `checked_add_i256` functions to runtime
2. Used them for `fe::BinOperator::Add` in mapper

### DIFF
--- a/compiler/src/yul/mappers/expressions.rs
+++ b/compiler/src/yul/mappers/expressions.rs
@@ -11,8 +11,10 @@ use fe_analyzer::builtins::{
     GlobalMethod,
 };
 use fe_analyzer::namespace::types::{
+    Base,
     FeSized,
     FixedSize,
+    Integer,
     Type,
 };
 use fe_analyzer::{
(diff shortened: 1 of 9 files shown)


compiler/src/yul/mappers/expressions.rs
```diff
@@ -239,7 +239,10 @@ pub fn expr_bin_operation(
             .typ;
 
         return match op.node {
-            fe::BinOperator::Add => Ok(expression! { add([yul_left], [yul_right]) }),
+            fe::BinOperator::Add => match typ.is_signed_integer() {
```
g-r-a-n-t, line 242: i imagine at some point we'll add more checking functions for smaller integers?

g-r-a-n-t (approved):

compiler/src/yul/mappers/expressions.rs
cburgdorf: Ah, good point. I will expand this PR to include checks for all integer sizes for additions.

cburgdorf: Ok, done. Additions for all supported sizes should now be checked.

g-r-a-n-t:
looks great :+1:

PR #267, opened 2021-02-24
Overflow/Underflow protection for subtraction

cburgdorf:
### What was wrong?

We need subtraction to perform over/underflow checks by default (#153)

### How was it fixed?

1. Refactored existing code based on `add` to move all min/max values into a shared map
2. Added code to generically generate names for these functions with also cleans up dispatching
3. Refactored code to generate functions based on all min/max values and the generic function name generation
4. Then added `sub` protection based on the refactoring
5. Added tests.

### DIFF
--- a/compiler/src/yul/constants.rs
+++ b/compiler/src/yul/constants.rs
@@ -1,0 +1,59 @@
+use fe_analyzer::namespace::types::Integer;
+use maplit::hashmap;
+use std::collections::HashMap;
+use yultsur::*;
+
+/// Return a hashmap containing min/max YUL literals for each supported integer
+/// size
+pub fn numeric_min_max() -> HashMap<Integer, (yul::Expression, yul::Expression)> {
+    hashmap! {
+        Integer::U256 => (
+            literal_expression! {0x0},
+            literal_expression! {0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff},
+        ),
+        Integer::U128 => (
+            literal_expression! {0x0},
+            literal_expression! {0xffffffffffffffffffffffffffffffff},
+        ),
+        Integer::U64 => (
+            literal_expression! {0x0},
+            literal_expression! {0xffffffffffffffff},
+        ),
+        Integer::U32 => (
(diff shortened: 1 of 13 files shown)


compiler/src/yul/constants.rs
```diff
@@ -0,0 +1,57 @@
+use fe_analyzer::namespace::types::Integer;
+use maplit::hashmap;
+use std::collections::HashMap;
+use yultsur::*;
+
+pub fn numeric_min_max() -> HashMap<Integer, (yul::Expression, yul::Expression)> {
```
cburgdorf, line 6: I wanted to have this as an actual constant using [phf](https://docs.rs/phf/0.8.0/phf ) but that would have required `Integer` to to implement `PhfHash` as it doesn't work with the builtin `Hash` trait. But requiring to depend on `phf` across multiple crates just for this didn't feel quite right to me. I gave up and left this as a method.

g-r-a-n-t: Its fine. Could we get some docs on this?

compiler/src/yul/runtime/functions/mod.rs
```diff
@@ -36,17 +36,6 @@ pub fn std() -> Vec<yul::Statement> {
...
-        math::checked_add_u8(),
-        math::checked_add_i256(),
-        math::checked_add_i128(),
-        math::checked_add_i64(),
-        math::checked_add_i32(),
-        math::checked_add_i16(),
-        math::checked_add_i8(),
-    ]
+    ];
+    [fns, math::all()].concat()
```
g-r-a-n-t, line 40: Now that we have alot more functions, we should probably organize all of the functions this way. Not something that needs to happen in this PR though.

g-r-a-n-t (approved):
looks good

compiler/src/yul/runtime/functions/mod.rs
cburgdorf: Oh, this is perfect. Just yesterday I was asked if we had some *good first timer* issues in the code base. I'll make this one.

PR #270, opened 2021-02-26
Support for self.address and bool operations

g-r-a-n-t:
### What was wrong?

We didn't support the two small features mentioned above. They are both needed for the Uniswap demo.

### How was it fixed?

- analyzing the expressions
- mapping the expressions
- testing

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/270.feature.md
+++ b/newsfragments/270.feature.md
@@ -1,0 +1,29 @@
+Support for the boolean operations `and` and `or`.
+
+Examples:
+
+```
+contract Foo:
+    pub def bar(x: bool, y: bool) -> bool:
+        return x and y
+```
+
+```
+contract Foo:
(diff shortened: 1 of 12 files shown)


newsfragments/270.feature.md
```diff
@@ -0,0 +1,29 @@
...
+Support for `self.address`.
+
+This expression returns the address of the current contract.
+
+Example:
+
+```
+contract Foo:
+    pub def bar() -> address:
+        return self.address
```
cburgdorf, line 26: We should put `address` into the keywords section of the spec

https://github.com/ethereum/fe/blob/5142f7e9c68aa6b5b3df6e2a85c3ab212d8badbc/spec/index.md#21-keywords

cburgdorf (approved):

PR #271, opened 2021-02-26
Add overflow/underflow checks for multiplicatons

cburgdorf:
### What was wrong?

We do not yet perform overflow/underflow checks for multiplications.

### How was it fixed?

Pretty straight forward based on the previous implementations for add and sub. Except for that pesty rusty-evm bug that I created an issue for: https://github.com/ethereum/fe/issues/285

### DIFF
--- a/compiler/src/yul/runtime/functions/math.rs
+++ b/compiler/src/yul/runtime/functions/math.rs
@@ -22,6 +22,25 @@ pub fn checked_add_fns() -> Vec<yul::Statement> {
     ]
 }
 
+/// Return a vector of runtime functions for multiplications with
+/// over-/underflow protection
+pub fn checked_mul_fns() -> Vec<yul::Statement> {
+    vec![
+        checked_mul_unsigned(Integer::U256),
+        checked_mul_unsigned(Integer::U128),
+        checked_mul_unsigned(Integer::U64),
+        checked_mul_unsigned(Integer::U32),
+        checked_mul_unsigned(Integer::U16),
+        checked_mul_unsigned(Integer::U8),
+        checked_mul_signed(Integer::I256),
+        checked_mul_signed(Integer::I128),
+        checked_mul_signed(Integer::I64),
+        checked_mul_signed(Integer::I32),
+        checked_mul_signed(Integer::I16),
+        checked_mul_signed(Integer::I8),
+    ]
+}
+
 /// Return a vector of runtime functions for subtraction with over-/underflow
 /// protection
 pub fn checked_sub_fns() -> Vec<yul::Statement> {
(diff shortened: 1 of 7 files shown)


compiler/tests/fixtures/checked_arithmetic.fe
```diff
@@ -70,4 +70,45 @@ contract CheckedArithmetic:
...
+    pub def mul_u16(left: u16, right: u16) -> u16:
+        return left * right
+
+    pub def mul_u8(left: u8, right: u8) -> u8:
+        return left * right
+
+    pub def weird_bug() -> i256:
+        val1: i256 = i256(-57896044618658097711785492504343953926634992332820282019728792003956564819968)
+        val2: i256 = i256(1)
+        return val1 * val2
```
cburgdorf, line 96: This is `min_val_i256 * 1` and it fails and I have no idea why. The same thing works in solidity.

compiler/src/yul/runtime/functions/math.rs
```diff
@@ -38,7 +57,47 @@ pub fn checked_sub_fns() -> Vec<yul::Statement> {
...
+
+    function_definition! {
+        function [fn_name](val1, val2) -> product {
+            // overflow, if val1 > 0, val2 > 0 and val1 > (max_value / val2)
+            (if (and((and((sgt(val1, 0)), (sgt(val2, 0)))), (gt(val1, (div([max_value.clone()], val2)))))) { (revert(0, 0)) })
+            // underflow, if val1 > 0, val2 < 0 and val2 < (min_value / val1)
+            (if (and((and((sgt(val1, 0)), (slt(val2, 0)))), (slt(val2, (sdiv([min_value.clone()], val1)))))) { (revert(0, 0)) })
+            // underflow, if val1 < 0, val2 > 0 and val1 < (min_value / val2)
+            // THIS IS THE CHECK THAT CAUSES THE REVERT
+            (if (and((and((slt(val1, 0)), (sgt(val2, 0)))), (slt(val1, (sdiv([min_value], val2)))))) { (revert(0, 0)) })
```
cburgdorf, line 95: This is the check that causes us to revert. We detect an underflow. The check is 100 % identical to solidity. Only the variables are renamed but I double checked the generated yul functions and logic is exactly the same what my `solc 0.6.7+commit.b8d736ae.Linux.g++` generates.

compiler/tests/evm_contracts.rs
```diff
@@ -648,6 +648,15 @@ fn checked_arithmetic() {
             &[],
         );
 
+
+        // The I256 bug in isolation
+        harness.test_function(
```
cburgdorf, line 653: This is testing the bug in isolation. Basiscally `i256_min_value * 1` doesn't work and I can't figure out why.

compiler/tests/evm_runtime.rs
```diff
@@ -187,6 +187,22 @@ fn test_runtime_sloadn() {
...
+fn test_weird_bug() {
+    with_executor(&|mut executor| {
+        test_runtime_functions(
+            &mut executor,
+            functions::std(),
+            statements! {
+                (let min_val := 0x8000000000000000000000000000000000000000000000000000000000000000)
+                (let val1 := 0x8000000000000000000000000000000000000000000000000000000000000000)
+                (let val2 := 0x01)
+                [assert_eq!(0, (and((and((slt(val1, 0)), (sgt(val2, 0)))), (slt(val1, (sdiv(min_val, val2)))))))]
```
g-r-a-n-t, line 200: So this is returning true and we expect false? Could be a solidity bug, I don't quite understand the logic here.

cburgdorf: Yes, this is the exact line (from the entire check) that causes us to revert. The actual check is [here](https://github.com/ethereum/fe/pull/271#discussion_r583583219)

Computing `i256_min_value * 1` in solidity with remix works just fine. And when I take a snippet of solidity that has that code and transpile it to yul with `solc snippet.sol --ir` then this check is what I get. And this is how I created all the checks so far and all of them except this one worked.

One thing that I could think of is that there's a bug with this check in the generated YUL that I get from `solc` and that it works in remix because it doesn't go through the YUL step and actually uses a different computation than the one that I get from `solc snippet.sol --ir`. I'll give it another shot today.

compiler/tests/evm_runtime.rs
```diff
@@ -187,6 +187,34 @@ fn test_runtime_sloadn() {
...
+                // (let val1 := 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff80)
+
+                (let val2 := 0x1)
+                // underflow, if val1 < 0, val2 > 0 and val1 < (min_value / val2)
+                //[assert_eq!(0, (and((and((slt(val1, 0)), (sgt(val2, 0)))), (slt(val1, (sdiv(min_val, val2)))))))]
+
+                // i256_min < (i256_min / 1)
+                // e.g. for i8 that would be: -128 < (-128 / 1)
+                // Why does this return 1? rust_evm bug?
+                [assert_eq!(0x1, (slt(val1, (sdiv(min_val, val2)))))]
```
cburgdorf, line 211: @g-r-a-n-t This seems to be the offending part. E.g. if you compute `i8_min < (i8_min / 1)` aka `-128 < (-128 / 1)` the result is `false` / `0x0`. But if you do the same with `i256` it returns `true` / `0x1` which doesn't seem to make sense. I wonder if this is an actual `rust-evm` bug.

I noticed there's a newer rust-evm version released already but nothing suspicious in the changes stood out to me. 

Another idea would be that there's a bug in the conversion from YUL to bytecode but I'm not yet sure how to best validate that.

cburgdorf:
@g-r-a-n-t I skipped the failing `i256` test for now and created an issue. I'll try to get this fixed upstream. This is ready for review now.

g-r-a-n-t (approved):

PR #276, opened 2021-03-01
Full support for empty tuples

cburgdorf:
### What was wrong?

Even though functions do always implicitly return empy tuples it isn't yet supported to actually use the empty tuple syntax (`()`) in code.

E.g. the following doesn't work:

```
  pub def explicit_return():
    return ()
```

This doesn't work either:

```
  pub def explicit_return() -> ():
    pass
```

With this change, all of these are treated equal:

```
contract Foo:

  pub def explicit_return_a1():
    return

  pub def explicit_return_a2():
    return ()

  pub def explicit_return_b1() ->():
    return

  pub def explicit_return_b2() ->():
    return ()

  pub def implicit_a1():
    pass

  pub def implicit_a2() ->():
    pass
```

### How was it fixed?

1. Taught the Parser how to parse tuples via the `tuple_type` function
2. Added tests for tuple parsing
3. Added support for (empty!) tuples in the analyzer and mapper pass
4. Taught the mapper to treat `return ()` as if there is no return at all (`leave`)
5. A few more tests

### To-Do

- [x] Clean up commit history

### DIFF
--- a/compiler/src/abi/builder.rs
+++ b/compiler/src/abi/builder.rs
@@ -116,10 +116,14 @@ fn func_def<'a>(
         .collect::<Result<Vec<FuncInput>, CompileError>>()?;
 
     let outputs = if let Some(return_type) = return_type {
-        vec![FuncOutput {
-            name: "".to_string(),
-            typ: type_desc(type_defs, &return_type.node)?,
-        }]
+        match type_desc(type_defs, &return_type.node)? {
+            // Should there be a different ABI for `pub def foo() -> ():` and `pub def foo():`?
+            VarType::Tuple(items) if items.is_empty() => vec![],
+            _ => vec![FuncOutput {
+                name: "".to_string(),
+                typ: type_desc(type_defs, &return_type.node)?,
+            }],
+        }
     } else {
         vec![]
     };
@@ -186,7 +190,13 @@ fn type_desc<'a>(
             Ok(VarType::FixedArray(Box::new(inner), *dimension))
         }
         fe::TypeDesc::Map { .. } => Err(CompileError::static_str("maps not supported in ABI")),
-        fe::TypeDesc::Tuple { .. } => unimplemented!(),
+        fe::TypeDesc::Tuple { items } => {
+            let items = items
+                .iter()
+                .map(|item| type_desc(type_defs, &item.node))
+                .collect::<Result<_, _>>()?;
+            Ok(VarType::Tuple(items))
+        }
     }
 }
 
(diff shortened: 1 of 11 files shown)


compiler/src/abi/builder.rs
```diff
@@ -116,10 +116,14 @@ fn func_def<'a>(
         .collect::<Result<Vec<FuncInput>, CompileError>>()?;
 
     let outputs = if let Some(return_type) = return_type {
-        vec![FuncOutput {
-            name: "".to_string(),
-            typ: type_desc(type_defs, &return_type.node)?,
-        }]
+        match type_desc(type_defs, &return_type.node)? {
+            // Should there be a different ABI for `pub def foo() -> ():` and `pub def foo():`?
```
cburgdorf, line 120: @g-r-a-n-t I'm unsure about this. On the one hand, it would feel weird if `pub def foo() -> ():` and `pub def foo():` resulted in different ABI types. On the other hand, **forcing**  `pub def foo() -> (): to become `pub def foo():` would prevent people from matching a given ABI spec they are trying to implement. Or is this a non-issue because of something that I might be missing?

g-r-a-n-t: I think we want to treat `pub def foo() -> ():` the same as `pub def foo():`, since both technically return an empty tuple. It might even make sense to emit a warning in the case of `pub def foo() -> ():`.

The ABI spec does say that you can have an empty tuple..

> [`(T1,...,Tk)` for `k >= 0` and any types `T1, …, Tk`](https://docs.soliditylang.org/en/v0.8.1/abi-spec.html#formal-specification-of-the-encoding)

So there is room for this sort of thing in the ABI, I'm just very skeptical of there being an instance where putting an empty tuple in the json ABI makes sense.

g-r-a-n-t (approved):

compiler/src/abi/builder.rs
cburgdorf: >I'm just very skeptical of there being an instance where putting an empty tuple in the json ABI makes sense

So am I but let's say there's a project that defines some interface `IStrategy` with a method `pub def do_something(val: Whatever) -> ()`. The way I understand it, one would not be able to write a strategy in Fe that produces an ABI that matches that ABI signature because the Fe compiler would always normalize it to `pub def do_something(val: Whatever):`

cburgdorf: Ok, I'll merge this as is. If this turns out to become a problem we can address it later.

PR #289, opened 2021-03-04
Reorganize contract tests

cburgdorf:
### What was wrong?

Our contract tests could be a bit more organized.

### How was it fixed?

1. All contract tests that relate to a demo from `fixtures/demos` where moved into its own `demo_<name>` file. (e.g. `demo_erc20`)
2. `evm_runtime` was renamed to just `runtime`. These are the tests that relate to the Fe runtime that is available to every Fe smart contract.
3. `evm_contracts` was renamed to `fixture_contracts`. These are the tests that relate to all other `fixtures` except for those that are in `fixtures/demos`. In the future we will probably break this up some more when it grows to large.


g-r-a-n-t (approved):
What do you think about renaming `fixture_contracts.rs` to `features.rs` and then moving the remaining fixtures under a new folder named `features`?

We could also move the stress test fixtures into their own directory too. So the directories would be:

- **demos**: contracts for real-world use
- **features**: simple contracts that narrowly test a given feature
- **stress**: complicated contracts that test broad behavior by doing strange and unneccessary things
- **compile_errors**: contracts that should fail to compile

cburgdorf:
I followed your proposed structure

PR #294, opened 2021-03-05
Fix visibility issue with subsequently defined contracts

cburgdorf:
### What is wrong?

If two contracts are defined in a module, only the second contract can see the interface of the first.

e.g.

```python
# can't see Bar
contract Foo:
   ...

# can see Foo
contract Bar:
   ...
```

### How was it fixed?

1. Contract analysis is split into multiple passes:
    1 The first pass collects just the names of the contracts and makes them available as types without further inspecting the contracts
    2. Next we inspect all contract bodies and we update all contract types to contain the actual function signatures
    3. Lastly we walk over all function bodies which is save at this point because now we have all information to e.g. parse something like `ExternalContract(some_address).some_api()`
2. To make this happen, two more APIs where added to the scope `add_contract_def` and `update_contract_def`. The last one is needed to update the contract types in the second phase (as explained above)
3. There's a bit of extra machinery to deal with the fact that part of the scope may be setup with contracts that were only partially resolved at the time when they were used but the scope has a way to mitigate that to ensure it always returns the up to date version when it is queried.
4. Because the module scope now contains all contracts it needs a way to filter out the main contract when it is aggregating `external_contracts`. To do that, `name` was added to `ContractScope` (we also have `name` for `BlockScope` so there's a precedence for having the `name` as a scope identifier) so that we can filter out the "self contract" upon aggregation.
5. The Uniswap demo was updated to use that feature

### DIFF
--- a/analyzer/src/namespace/scopes.rs
+++ b/analyzer/src/namespace/scopes.rs
@@ -43,6 +43,7 @@ pub struct ModuleScope {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct ContractScope {
+    pub name: String,
     pub parent: Shared<ModuleScope>,
     pub interface: Vec<String>,
     pub event_defs: HashMap<String, Event>,
@@ -173,8 +174,9 @@ impl ModuleScope {
 }
 
 impl ContractScope {
-    pub fn new(parent: Shared<ModuleScope>) -> Shared<Self> {
+    pub fn new(name: &str, parent: Shared<ModuleScope>) -> Shared<Self> {
         Rc::new(RefCell::new(ContractScope {
+            name: name.to_owned(),
             parent,
             function_defs: HashMap::new(),
             event_defs: HashMap::new(),
@@ -430,7 +432,7 @@ mod tests {
     #[test]
     fn test_scope_resolution_on_first_level_block_scope() {
         let module_scope = ModuleScope::new();
-        let contract_scope = ContractScope::new(module_scope);
+        let contract_scope = ContractScope::new("", module_scope);
         let block_scope_1 = BlockScope::from_contract_scope("", Rc::clone(&contract_scope));
         assert_eq!(block_scope_1, block_scope_1.borrow().function_scope());
         assert_eq!(contract_scope, block_scope_1.borrow().contract_scope());
@@ -439,7 +441,7 @@ mod tests {
     #[test]
     fn test_scope_resolution_on_second_level_block_scope() {
         let module_scope = ModuleScope::new();
-        let contract_scope = ContractScope::new(module_scope);
+        let contract_scope = ContractScope::new("", module_scope);
         let block_scope_1 = BlockScope::from_contract_scope("", Rc::clone(&contract_scope));
         let block_scope_2 =
             BlockScope::from_block_scope(BlockScopeType::IfElse, Rc::clone(&block_scope_1));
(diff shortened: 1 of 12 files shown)


analyzer/src/traversal/contracts.rs
```diff
@@ -26,12 +25,46 @@ use std::rc::Rc;
...
+        contract_scope
+            .borrow()
+            .module_scope()
+            .borrow_mut()
+            .add_contract_def(
+                name.node,
+                Rc::clone(&contract_scope),
+                Contract {
+                    name: name.node.to_owned(),
+                    // Function defintions are added in a subsequent walk
```
g-r-a-n-t, line 44: I'm thinking it would be cleaner to add the function definitions to the scope in this pass along with a new contract type definition.

If we've analyzed the signatures of the functions and added them to the scope, we should be able to create a contract type at the same time, eliminating the need for these partial contract definitions.

So:

`contract_def` pass
- adds the function definitions to the contract scope via a call `func_def`
- adds the contact type to the module scope

`contract_body` pass
- analyzes the body of each function in the contract via a call to `func_body`

analyzer/src/namespace/scopes.rs
```diff
@@ -31,10 +38,12 @@ pub struct ContractFieldDef {
 #[derive(Clone, Debug, PartialEq)]
 pub struct ModuleScope {
     pub type_defs: HashMap<String, Type>,
+    contract_defs: HashMap<String, ContractDef>,
```
g-r-a-n-t, line 41: As mentioned below, I think we could get rid of this field by adding the contract's type right away in `contract_def`

analyzer/src/traversal/module.rs
```diff
@@ -17,6 +19,9 @@ use std::rc::Rc;
 pub fn module(context: Shared<Context>, module: &fe::Module) -> Result<(), SemanticError> {
     let scope = ModuleScope::new();
 
+    // Walk over the module but ignore everything that is *inside* contracts.
+    // This is to make contract types known prior to where we know their exact shape
+    // so that they can already be used as types in other contracts.
     for stmt in module.body.iter() {
```
g-r-a-n-t, line 25: To reiterate, I think there should only be two passes here: a `def` pass and a `body` pass. The `def` pass would be responsible for adding definitions to the scope and the body `pass` would be responsible for analyzing function bodies (in both structs and contracts). To analyze definitions, we shouldn't need to dive into any bodies, but for function bodies, we will need information from the definition pass.

analyzer/src/traversal/contracts.rs
cburgdorf: I had this in the beginning but there are two problems:

1. If we do that we lose the ability to use contract types in function signatures which on second thought I do think is valuable e.g. consider this:

```
def get strategy -> IStrategy:
    if some_condition:
        return FooStrategy(self.foo_address)
    else:
        return BarStrategy(self.bar_address)
```

If this is in contract `A` and `IStrategy` is defined below then we aren't aware of `IStrategy` at the point where we want to collect the function definitions from contract `A`.

2. Even if we ignore the former we will still wouldn't be able to obtain the final `ContractAttributes` after the first pass because `ContractAttributes` contains `external_contracts` so if contract `A` uses contract `B` (even if only in a function body) then the `ContractAttributes` of contract `A` need to also contain contract `B` which we haven't analysed at that point. In other words, even if we decide to collect the function definitions of contract `A` in the pass that creates the type, we will still need to recompute the `ContractAttributes` at the final hour when we have walked all the contracts so that they can contain all the `external_contracts` that are needed.

But as I write that down I noticed that this second issue isn't strongly related to the former. We could still do what you say and keep 2.) as is but I think we would lose out being able to handle the example code as explained in 1.)

cburgdorf:
Closing in favor for #298

PR #298, opened 2021-03-09
Improve visibility of subsequently defined contracts

cburgdorf:
### What was wrong?

This superseded #294 

If two contracts are defined in a module, only the second contract can see the interface of the first.

e.g.

```python
# can't see Bar
contract Foo:
   ...

# can see Foo
contract Bar:
   ...
```

### How was it fixed?

1. Contract analysis is split into multiple passes:
    1 The first pass collects contracts with their function definition but does not inspect function bodies and contract fields
    2. Once every contract was analyzed, we analyze the function bodies and contract field of each contract in a subsequent pass
   
2. Because the module scope now contains all contracts it needs a way to filter out the main contract when it is aggregating `external_contracts`. To do that, `name` was added to `ContractScope` (we also have `name` for `BlockScope` so there's a precedence for having the `name` as a scope identifier) so that we can filter out the "self contract" upon aggregation.
3. The Uniswap demo was updated to use that feature

This PR doesn't yet address issue #299 though.

### DIFF
--- a/analyzer/src/traversal/contracts.rs
+++ b/analyzer/src/traversal/contracts.rs
@@ -31,7 +31,7 @@ pub fn contract_def(
     stmt: &Spanned<fe::ModuleStmt>,
 ) -> Result<Shared<ContractScope>, SemanticError> {
     if let fe::ModuleStmt::ContractDef { name, body } = &stmt.node {
-        let contract_scope = ContractScope::new(Rc::clone(&module_scope));
+        let contract_scope = ContractScope::new(name.node, Rc::clone(&module_scope));
 
         for stmt in body.iter() {
             match &stmt.node {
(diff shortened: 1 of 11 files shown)


analyzer/src/traversal/contracts.rs
```diff
@@ -29,14 +29,17 @@ pub fn contract_def(
...
+        let contract_scope = ContractScope::new(name.node, Rc::clone(&module_scope));
 
         for stmt in body.iter() {
             match &stmt.node {
                 fe::ContractStmt::ContractField { .. } => {
-                    contract_field(Rc::clone(&contract_scope), stmt)
+                    // Contract fields are evaluated in the next pass together with function bodies
+                    // so that they can use other contract types that may only be defined after the
+                    // current contract.
+                    Ok(())
```
cburgdorf, line 42: @g-r-a-n-t since fields aren't in the public API of contracts it omitting them here doesn't interfere with the contract representation in the scope. But by delaying analysis to the next pass we can at least support referring to subsequent contracts in contract fields.

cburgdorf:
@g-r-a-n-t Here's my second attempt which removes all the machinery to manage partial representation of contracts. I created  #299 to track the issue with not being able to use subsequently defined contract types in function definitions.

g-r-a-n-t (approved):

PR #300, opened 2021-03-09
Reorganize runtime functions

### LINKED ISSUE #269: Reorganize runtime functions in mapper
### What is wrong?

We should reorganize how we include the runtime functions here:

https://github.com/ethereum/fe/blob/f9782666de2d970ebd0080caedddf4c158afc91f/compiler/src/yul/runtime/functions/mod.rs#L11-L44

### How can it be fixed

Each module should expose an `all()` function so that `std` can be refactored to look like this:

```rust
/// Returns all functions that should be available during runtime.
pub fn std() -> Vec<yul::Statement> {
    [
        data::all(),
        abi::all(),
        contracts::all(),
        math::all(),
        whatever_else_may_exist::all()
    ].concat()
}
```

PascalPrecht:
### What was wrong?

Runtime functions of the `yul` module needed to be exposed individually, making it cumbersome to expose them all at ones, especially given that the runtime API might grow in the future

### How was it fixed?

This commit moves yul runtime functions of various modules into dedicated `*::all()` APIs to make exposing them easier.

Closes #269

### DIFF
--- a/compiler/src/yul/runtime/functions/abi.rs
+++ b/compiler/src/yul/runtime/functions/abi.rs
@@ -10,6 +10,15 @@ use fe_analyzer::namespace::types::{
 };
 use yultsur::*;
 
+/// Return all abi runtime functions
+pub fn all() -> Vec<yul::Statement> {
+    vec![
+        unpack(),
+        pack(AbiDecodeLocation::Calldata),
+        pack(AbiDecodeLocation::Memory),
+    ]
+}
+
 /// Creates a batch of encoding function for the given type arrays.
 ///
 /// It sorts the functions and removes duplicates.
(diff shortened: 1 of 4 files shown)


PascalPrecht:
Since this exposes new public APIs we might need to update the release note sections? 
Let me know if this is needed and I'll update the PR.

compiler/src/yul/runtime/functions/abi.rs
```diff
@@ -10,6 +10,11 @@ use fe_analyzer::namespace::types::{
 };
 use yultsur::*;
 
+// Return all abi runtime functions
```
cburgdorf, line 13: Can you change the comments to triple slashes `///` to make sure they are picked up as API docs by rustdoc (Applies to all the added comments in this PR.

cburgdorf (approved):
Hey Pascal, thanks for the PR. This looks good except for the comments that should use triple slashes and the failed linting job. You can run `cargo fmt --all ` to have cargo do the reformatting for you.

Oh, and there's no release notes updated needed for the refactoring but thanks for asking.

PascalPrecht:
Love the super quick feedback - thanks!

PR #301, opened 2021-03-09
Introduce new FixedSize::bool() API in fe::analyzer

### LINKED ISSUE #274: Create a FixedSize::bool() constructor API
### What is wrong?

We have quite a few places like this:

https://github.com/ethereum/fe/blob/177ee35ba4b12150f5cb14152e066d364b1a0161/analyzer/src/namespace/scopes.rs#L379

It would be nice if we could write `FixedSize::bool()` instead

### How can it be fixed

1. Create a `bool()` function on `FixedSize` that creates a `FixedSize::Base(Base::Bool)`
2. Replace every place that uses `FixedSize::Base(Base::Bool)` with ``FixedSize::bool()`

PascalPrecht:
### What was wrong?

The only way to get a `FixedSize::Base(Base::Bool)` variant is to use that very API. 
As mentioned in #274, a constructor API like `FixedSize::bool()` makes this easier.

### How was it fixed?

To improve developer ergonomics, this commit introduces a new API
```
FixedSize::bool()
```

which is equivalent to:

```
FixedSize::Base(Base::Bool)
```

Closes #274

### DIFF
--- a/analyzer/src/namespace/events.rs
+++ b/analyzer/src/namespace/events.rs
@@ -82,7 +82,7 @@ mod tests {
             vec![
                 FixedSize::Base(Base::Address),
                 FixedSize::Base(Base::Address),
-                FixedSize::Base(Base::Bool),
+                FixedSize::bool(),
             ],
             vec![1],
         );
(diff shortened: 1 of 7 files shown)


PascalPrecht:
Let me know if this is release note worthy

cburgdorf:
Sweet! Landed! :tada:

PR #304, opened 2021-03-11
Created contracts included in constructor object.

### LINKED ISSUE #284: "missing data object" when compiling a contract with both an `__init__` and `create` statement.
### What is wrong?

When a contract has an `__init__` function defined, we copy the content of the runtime object into the constructor object. This is to ensure that any functions defined within the contract can be called from the `__init__` function. This is problematic when creating external contracts, since the contract's code is only added to the runtime object.

### How can it be fixed

The simplest way to solve this would be to add any external contract code to the constructor object too. One issue with this, though, is that the Yul optimizer will not remove any unused data objects. These extra data object will not increase the contracts size, but will increase the contract creation calldata size. If there is not much of a cost here, then we could probably just leave them in there, otherwise we'll need to figure out how to optimize this.

g-r-a-n-t:
### What was wrong?

fixes #284

### How was it fixed?

Added the `created_contracts` vector to the constructor object.

---

The Yul optimizer does not currently remove these objects if they are unused. Meaning, created contracts will be included in the initialization code twice no matter what. This does not increase the contract size, but increases the initialization code size.

The cost of this extra tx data is 16 for non-zero bytes and 4 for zero bytes (see [appendix g](https://ethereum.github.io/yellowpaper/paper.pdf)).

For example, say we write a contract that deploys an ERC-20 contract and has an `__init__` that does **not** create an ERC-20 contract. A compiled Fe ERC-20 contract is about `2600` bytes. So, the extra cost would be slightly less than `42000`, which is roughly equivalent to the cost of 2 `sstore` operations. This is a non-trivial expense and we should find a way to remove these unused data object. It might be the case that the Yul team plans on implementing this, so we don't have to worry about. 

### To-Do

- [x] Create an issue to track the gas issue. #305
- [x] Add a test where a contract is created in `__init__`.
- [x] Clean up commit history

### DIFF
--- a/compiler/src/yul/mappers/contracts.rs
+++ b/compiler/src/yul/mappers/contracts.rs
@@ -15,9 +15,9 @@ pub fn contract_def(
     created_contracts: Vec<yul::Object>,
 ) -> Result<yul::Object, CompileError> {
     if let fe::ModuleStmt::ContractDef { name, body } = &stmt.node {
-        let mut init = None;
-        let mut user_functions = vec![];
         let contract_name = name.node;
+        let mut init_function = None;
+        let mut user_functions = vec![];
 
         // map user defined functions
         for stmt in body.iter() {
@@ -25,7 +25,7 @@ pub fn contract_def(
                 (context.get_function(stmt), &stmt.node)
             {
                 if name.node == "__init__" {
-                    init = Some((
+                    init_function = Some((
                         functions::func_def(context, stmt)?,
                         attributes.param_types(),
                     ))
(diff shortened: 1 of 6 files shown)


compiler/src/yul/mappers/contracts.rs
```diff
@@ -82,7 +82,7 @@ pub fn contract_def(
         return Ok(yul::Object {
             name: identifier! { (contract_name) },
             code: constructor,
-            objects: vec![runtime],
+            objects: [vec![runtime], created_contracts].concat(),
```
g-r-a-n-t, line 85: `created_contracts` should be added conditionally

compiler/src/yul/mappers/contracts.rs
```diff
@@ -61,28 +53,56 @@ pub fn contract_def(
...
-            objects: created_contracts,
+            objects: created_contracts.clone(),
             // We can't reach to data objects in the "contract" hierachy so in order to have
             // the data objects available in both places we have to put them in both places.
             data: data.clone(),
         };
 
+        // Build the code and and objects fields for the constructor object.
+        //
+        // If there is an `__init__` function defined, we must include the everything
```
g-r-a-n-t, line 77: typo

cburgdorf (approved):

PR #306, opened 2021-03-12
Lowering pass and tuple support

### LINKED ISSUE #205: Implement a "desugaring" pass that turns Fe code into other Fe code
### Abstract

Compilers often implement a desugaring pass where certain features are just syntactic sugar for a more verbose form.

For instance, if we have support for structs than tuples can be implemented as syntactic sugar that internally gets converted to a struct.

For instance, a Tuple of type `(u8, string100)` may just be syntactic sugar such a struct:

```
struct $tuple_u8_string100:
    0: u8
    1: string100
```

Another example would be rewriting string literals to constants such as:

```
def foo(val: string100):
    if val == "foo":
        pass
```

```
def foo(val: string100):
    if val == "$CONST_STRING_FOO:
        pass
```

### LINKED ISSUE #318: Fix ContextHarness
### What is wrong?

from #315

> ContextHarness was broken by this change, since it relies on adding attributes to the context via the span. I've disabled this with the feature fix-context-harness and will create an issue to fix this (if merging this with the failing tests sounds ok).

### How can it be fixed

This isn't very straight forward.

g-r-a-n-t:
closes #205 

The goal of this PR is to lay the groundwork for lowering and use it to support tuples.

See how Rust does it: https://rustc-dev-guide.rust-lang.org/lowering.html.

Terms:
- High-level Intermediate Representation (HIR): A subset of Fe-lang that can be digested by the Yul generator.
- Lowering: The process of mapping the source AST to an HIR AST.

Design questions/thoughts:
- :heavy_check_mark:  We will not be able use spans as primary node IDs anymore, since the HIR will not be parsed from a string. So, we will need to introduce a new form of identification. 
  - Should node IDs be deterministic or random? I'm leaning towards random since it's easier to implement. See the [uuid crate](https://docs.rs/uuid/0.8.2/uuid/).
  - See [Identifiers in rustc](https://rustc-dev-guide.rust-lang.org/identifiers.html?highlight=NodeId#in-the-ast).
  - Where exactly do we add an ID? `Spanned` or the node itself? or somewhere else?
    - `Spanned` would probably be least invasive.
    - Rust assigns IDs to each node. They also have different types of IDs for different nodes.
    - ~~I'm leaning towards placing IDs in the individual nodes and leaving `Spanned` the way it is.~~
    - On second thought, our AST doesn't have `Kinds` like rustc (see [here](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/struct.Expr.html)). So assigning IDs to nodes won't be very clean. Maybe we should add IDs to `Spanned` afterall. It may warrant a name change from `Spanned` to `Node`. 
    - Fixed in #315. We now need to fix #318.
- Should we reanalyze the HIR AST, or do we want to build up node attributes as we go?
  - Reanalyzing adds more computational overhead, but may be simpler to implement.
  - Reanalyzing also adds more transparency, since definitions would be added to the HIR AST, which could be printed by the compiler. Otherwise, we would just be altering the attributes of a given node, which doesn't event constitute lowering.
  - Overall, I'm leaning towards reanalyzing.
- Should we reuse the existing AST structs or create new ones for the HIR?
  - I'm thinking we should just reuse the existing ones. The HIR will just consist of a subset of node types. If the Yul codegen pass encounters a non-HIR node, it should panic.
  - Rust has separate structs for the HIR. We could do the same in the future, but I don't think it's necessary right now. 
- Does the lowering API and general architecture need to be any different than the current Fe -> Yul pass? In that, we traverse the AST once and map each node.
  - Probably not. We just need to make sure everything that is needed at each node is provided in the `Context` attributes. For example, if a tuple is used inside of a contract, we should record this on the contract or module node, so the tuple's struct definition can be added.
- We should make sure naming conflicts are impossible during lowering. This can be avoided by using illegal names and adding some sort of `non_strict_mode` parameter to `analysis` that eases up name checks.
- Shorthand macros for writing Fe HIR ASTs would be helpful.

Currently, I'm thinking all of the passes will look like this:

- **Parsing**: takes source code and produces an AST.
- **First analysis**: takes the source AST and produces node attributes.
  - The purpose of this pass is to find errors and build up contextual information.
- **Lowering**: Takes the source AST and the node attributes and produces an HIR AST (new).
- **Second analysis**: Takes the HIR AST and produces node attributes (also new, but doesn't require new code).
  - The purpose of this pass is only to build up contextual information.
- **Yul codegen**: Takes the HIR AST and node attributes and produces Yul code.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/analyzer/src/lib.rs
+++ b/analyzer/src/lib.rs
@@ -15,6 +15,7 @@ use crate::namespace::events::EventDef;
 use crate::namespace::scopes::{
     ContractFunctionDef,
     ContractScope,
+    ModuleScope,
     Shared,
 };
 use crate::namespace::types::{
@@ -282,6 +283,19 @@ impl From<ContractFunctionDef> for FunctionAttributes {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct ModuleAttributes {
+    pub type_defs: HashMap<String, Type>,
+}
+
+impl From<Shared<ModuleScope>> for ModuleAttributes {
+    fn from(scope: Shared<ModuleScope>) -> Self {
+        Self {
+            type_defs: scope.borrow().type_defs.clone(),
+        }
+    }
+}
+
 /// Contains contextual information about a Fe module and can be queried using
 /// `Spanned` AST nodes.
 #[derive(Clone, Debug, Default)]
@@ -293,6 +307,7 @@ pub struct Context {
     contracts: HashMap<NodeId, ContractAttributes>,
     calls: HashMap<NodeId, CallType>,
     events: HashMap<NodeId, EventDef>,
+    module: Option<ModuleAttributes>,
 }
 
 impl Context {
@@ -309,6 +324,7 @@ impl Context {
             contracts: HashMap::new(),
             calls: HashMap::new(),
             events: HashMap::new(),
+            module: None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


g-r-a-n-t:
lowering pass was added in #338 

will continue working on tuple support in a different branch

PR #311, opened 2021-03-16
Implement msg.sig

### LINKED ISSUE #278: Implement msg.sig
### What is wrong?

We do not yet implement reading `msg.sig`

https://github.com/ethereum/fe/blob/177ee35ba4b12150f5cb14152e066d364b1a0161/compiler/src/yul/mappers/expressions.rs#L440

### How can it be fixed

Start here:

https://github.com/ethereum/fe/blob/177ee35ba4b12150f5cb14152e066d364b1a0161/compiler/src/yul/mappers/expressions.rs#L440

Background:

https://docs.soliditylang.org/en/develop/units-and-global-variables.html#block-and-transaction-properties

georgeroman:
Fixes #278 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/compiler/tests/fixtures/features/return_msg_sig.fe
+++ b/compiler/tests/fixtures/features/return_msg_sig.fe
@@ -1,0 +1,5 @@
+contract Foo:
+    pub def bar() -> bytes[32]:
+        msg_sig: bytes[32]
+        msg_sig = msg.sig
+        return msg_sig
--- a/compiler/tests/utils.rs
+++ b/compiler/tests/utils.rs
@@ -6,6 +6,7 @@ use evm_runtime::{
 use fe_compiler as compiler;
 use primitive_types::{
     H160,
+    H256,
     U256,
 };
 use std::collections::BTreeMap;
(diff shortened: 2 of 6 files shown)


compiler/tests/fixtures/features/return_msg_sig.fe
```diff
@@ -0,0 +1,5 @@
+contract Foo:
+    pub def bar() -> bytes[32]:
+        msgSig: bytes[32]
```
cburgdorf, line 3: The idiomatic way to write Fe code is with `snake_case`. Please change `msgSig` to `msg_sig`.

compiler/tests/utils.rs
```diff
@@ -316,6 +317,19 @@ pub fn bytes_token(s: &str) -> ethabi::Token {
     ethabi::Token::FixedBytes(ethabi::FixedBytes::from(s))
 }
 
+#[allow(dead_code)]
+pub fn bytes32(s: &str) -> Vec<u8> {
+    H256::from_str(s)
```
cburgdorf, line 322: Just pointing out that we try to not use single letter variable names such as `s`. That said, this file violates that rule numerous times and we don't need to fix that in this PR :)

cburgdorf (approved):
Thanks for the PR. Looks good. Just a tiny nitpick about the casing. Also do you mind creating a file `311.feature.md` in https://github.com/ethereum/fe/tree/master/newsfragments with the content being something like:

>Support for `msg.sig` to read the function identifier

This is to ensure the feature gets included in the next release notes.

georgeroman:
Addressed all comments.

cburgdorf:
Awesome! Landed :tada:

PR #315, opened 2021-03-18
Dummy lowering and node IDs

g-r-a-n-t:
### What was wrong?

We need some form of node identification for lowering. See #306.
 
This is an attempt to break up the mentioned PR.

### How was it fixed?

- Added a lowering module that doesn't do anything yet.
- Renamed `Spanned` to `Node` and added an ID field.

https://github.com/ethereum/fe/blob/953b429100faaa106a5b35ebe05c97b4423d10b0/parser/src/node.rs#L48-L54

- Refactored broken code and tests. It's worth mentioning that the hacky `spanned_expression` helpers were removed.

[ContextHarness](https://github.com/ethereum/fe/blob/953b429100faaa106a5b35ebe05c97b4423d10b0/analyzer/src/lib.rs#L402-L463) was broken by this change, since it relies on adding attributes to the context via the span. I've disabled this with the feature `fix-context-harness` and will create an issue to fix this (if merging this with the failing tests sounds ok).

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
What a huge diff! Good you separated it from the coming changes. It looks good. I don't like the disabled tests but if you say you'll track it with an issue and make sure to get them re-enabled as soon as possible then I'm ok with it given the complexity of the overall task.

Issue #310, opened 2021-03-15
Yul compilation failed: DeclarationError: Function name struct_House_get_price_ptr already taken in this scope


agroce:
Compiling this file:

```
struct House:
    price: u256
    price: u256
contract Foo:
    my_house: House
```

```
thread 'main' panicked at 'Yul compilation failed: {"errors":[{"component":"general","formattedMessage":"DeclarationError: Function name struct_House_get_price_ptr already taken in this scope.\n --> input.yul:1:16722:\n  |\n1 |  ... _val { return_val := add(ptr, 0) } function struct_House_get_price_ptr(ptr) -> return_val { return_val := add(ptr, 0) } pop(0) return(0, 0) }   }  }\n  |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n","message":"Function name struct_House_get_price_ptr already taken in this scope.","severity":"error","sourceLocation":{"end":16805,"file":"input.yul","start":16721},"type":"DeclarationError"},{"component":"general","formattedMessage":"DeclarationError: Variable name price already taken in this scope.\n --> input.yul:1:16477:\n  |\n1 |  ... = avail() pop(mcopym(val_0, 64)) } function struct_House_new(price, pr ... lloc(32) mstore(price_ptr, price) } function struct_House_get_price_pt ...\n  |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n","message":"Variable name price already taken in this scope.","severity":"error","sourceLocation":{"end":16635,"file":"input.yul","start":16476},"type":"DeclarationError"}]}', compiler/src/lib.rs:48:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:435:5
   2: fe_compiler::compile
   3: fe::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted

```

Looks like a struct variant of the fixed function thing.

[...]

Issue #275, opened 2021-03-01
Yul compilation failed: Variable name $sum already taken in this scope. [NEW VARIANT of #223]


agroce:
Compiling this file:

```
contract Foo:
    pub def bar() -> u256:
        my_array: u256[3]
        sum: u256 = 0
        for i in my_array:        sum: u256 = 0
        return sum
```

```
thread 'main' panicked at 'Yul compilation failed: {"errors":[{"component":"general","formattedMessage":"DeclarationError: Variable name $sum already taken in this scope.\n --> input.yul:1:314:\n  |\n1 |  ... dn(add($my_array, mul(i, 32)), 32) let $sum := 0 } } { return_val := $sum leave } } ...\n  |                                         ^^^^^^^^^^^^^\n\n","message":"Variable name $sum already taken in this scope.","severity":"error","sourceLocation":{"end":326,"file":"input.yul","start":313},"type":"DeclarationError"}]}', compiler/src/lib.rs:48:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:435:5
   2: fe_compiler::compile
   3: fe::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted
```
Basically a variant of #223 that is not fixed by the "fix".

On latest github master, built with

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer + afl.rs.

PR #317, opened 2021-03-19
Prevent several cases of making duplicate definitions under the same name

cburgdorf:
### What was wrong?

1. It is currently possible to define structs with multiple conflicting fields (#310)
2. It is currently possible to define multiple type definitions, structs or contracts in the same module with conflicting names
3. It is currently possible to define a var in a block scope that is already defined in a parent scope (#275)

### How was it fixed?

- Made `add_type_def` return `Result<(), SemanticError>` and consume it everywhere where it is used
- Made `add_field` on struct return `Result<(), SemanticError>` and consume it everywhere where it is used
- Made `add_var` first check `get_variable_def(name)` which also looks up in parent scopes
- Added a bunch of tests also for existing but untested behavior


g-r-a-n-t (approved):
Great! looking forward to seeing more PRs like this.

PR #323, opened 2021-03-22
Fix ICE when '__init__' returns any other type than '()'

### LINKED ISSUE #307: Yul compilation failed: Top-level expressions are not supposed to return values (this expression returns 1 value). Use ``pop()`` or assign them.
Compiling this file:

```
contract C:
 pub def __init__(t:address)->u256:
     return 0
```

```
thread 'main' panicked at 'Yul compilation failed: {"errors":[{"component":"general","formattedMessage":"TypeError: Top-level expressions are not supposed to return values (this expression returns 1 value). Use ``pop()`` or assign them.\n --> input.yul:1:323:\n  |\n1 |  ... _val { { return_val := 0 leave } } $$__init__(abi_decode_address_mem(params_start_mem, 0)) function contract_create2(data_ptr ...\n  |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n","message":"Top-level expressions are not supposed to return values (this expression returns 1 value). Use ``pop()`` or assign them.","severity":"error","sourceLocation":{"end":377,"file":"input.yul","start":322},"type":"TypeError"}]}', compiler/src/lib.rs:48:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:435:5
   2: fe_compiler::compile
   3: fe::main
Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

Y-Nak:
fixes #307.

### What was wrong?
Semantic checks missed `__init__` return type.

### How was it fixed?
Assert the return type is `()`.

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/traversal/functions.rs
+++ b/analyzer/src/traversal/functions.rs
@@ -63,6 +63,11 @@ pub fn func_def(
             .transpose()?
             .unwrap_or_else(|| Tuple::empty().into());
 
+        // `__init__` must not return any type other than `()`.
+        if name == "__init__" && !return_type.is_empty_tuple() {
+            return Err(SemanticError::type_error());
+        }
+
         let attributes: FunctionAttributes = contract_scope
             .borrow_mut()
             .add_function(
--- a/compiler/tests/compile_errors.rs
+++ b/compiler/tests/compile_errors.rs
@@ -71,6 +71,7 @@ use std::fs;
     case("return_addition_with_mixed_types.fe", "TypeError"),
     case("return_call_to_fn_with_param_type_mismatch.fe", "TypeError"),
     case("return_call_to_fn_without_return.fe", "TypeError"),
+    case("return_from_init.fe", "TypeError"),
     case("return_lt_mixed_types.fe", "TypeError"),
     case("strict_boolean_if_else.fe", "TypeError"),
     case("string_capacity_mismatch.fe", "StringCapacityMismatch"),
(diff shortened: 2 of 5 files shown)


cburgdorf (approved):
Olá Yoshitomo,

thanks for the fix. Looks good :+1: 
Can you also add a file `323.bugfix.md` to the `newsfragments` directory with the content being something like:

>Ensure analyzer rejects code that uses return values for `__init__` function

This is so that the bugfix will be mentioned in the upcoming release notes.

Y-Nak:
Changes:
* Add a file to `newsfragments`
* Replace `ReST` with `md` in `newsfragments/README.md`

Question:
  the `README.md` says 
  > If the PR fixes an issue, use that number here. If there is no issue, then open up the PR first and use the PR number for the 
  newsfragment.

  while it seems PR number is used currently even if the corresponding issue is opened.
  Should I also fix this document?

cburgdorf:
>If the PR fixes an issue, use that number here. If there is no issue, then open up the PR first and use the PR number for the
newsfragment.

I think that advice is still correct even though we often use the PR number even though the issue number would be slightly preferable. It's just a minor sloppiness on our side :upside_down_face:

Y-Nak:
Thanks for clarifying! I leave it as is then.

cburgdorf:
Landed :tada: Btw, I'm taking care of the failed coverage job in https://github.com/ethereum/fe/pull/325
It is unrelated to your changes.

PR #327, opened 2021-03-22
Panic when Yul mapping fails; print friendly message when fe panics

### LINKED ISSUE #168: Compiler bug panics

sbillig:
### What was wrong?

A yul mapping failure is indicative of a bug in the compiler, but the yul mappers were returning Result types. Fixes #168.

### How was it fixed?

Yul mappers now all panic on failure. I also added a panic hook to log the following message when fe panics:

```

```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,6 +646,7 @@ dependencies = [
  "fe-parser",
  "hex",
  "maplit",
+ "once_cell",
  "primitive-types",
  "rand",
  "rstest",
--- a/compiler/Cargo.toml
+++ b/compiler/Cargo.toml
@@ -26,6 +26,7 @@ stringreader = "0.1"
 # This fork supports concurrent compilation, which is required for Rust tests.
 solc = { git = "https://github.com/g-r-a-n-t/solc-rust", optional = true }
 maplit = "1.0.2"
+once_cell = "1.5.2"
 
 [dev-dependencies]
 evm-runtime = "0.18"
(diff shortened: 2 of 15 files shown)


g-r-a-n-t (approved):
Looks good. I think we could have a `327.internal.md` news fragment though.

PR #328, opened 2021-03-23
Properly reject emitting undefined events

### LINKED ISSUE #250: missing expression attributes at compiler/src/yul/mappers/expressions.rs:60:9
Compiling this file:

```
contract o:
 pub def p():
    emit t("")
```

```
thread 'main' panicked at 'missing expression attributes', compiler/src/yul/mappers/expressions.rs:60:9
stack backtrace:
   0: std::panicking::begin_panic
   1: fe_compiler::yul::mappers::expressions::expr
   2: fe_compiler::yul::mappers::expressions::call_arg
Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer + afl.rs.

cburgdorf:
### What was wrong?

Currently we do not reject code that uses `emit` with some undefined event such as `emit DoesNotExist`

Fixes #250

### How was it fixed?

1. Introduced a new error type `MissingEventDefinition`
2. Return `Err(SemanticError::missing_event_definition())` when the emitted event does not exit
3. Added a test case

### DIFF
--- a/analyzer/src/traversal/functions.rs
+++ b/analyzer/src/traversal/functions.rs
@@ -359,7 +359,7 @@ fn emit(
     {
         let event_name = expressions::expr_name_str(func)?;
 
-        if let Some(event) = scope.borrow().contract_event_def(event_name) {
+        return if let Some(event) = scope.borrow().contract_event_def(event_name) {
             context.borrow_mut().add_emit(stmt, event.clone());
 
             let argument_attributes = args
@@ -371,11 +371,13 @@ fn emit(
             if fixed_sizes_to_types(event.all_field_types())
                 != expression_attributes_to_types(argument_attributes)
             {
-                return Err(SemanticError::type_error());
+                Err(SemanticError::type_error())
+            } else {
+                Ok(())
             }
-        }
-
-        return Ok(());
+        } else {
+            Err(SemanticError::missing_event_definition())
+        };
     }
 
     unreachable!()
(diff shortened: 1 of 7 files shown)


analyzer/src/traversal/functions.rs
```diff
@@ -351,31 +351,37 @@ fn emit(
...
+                        Ok(())
+                    }
+                } else {
+                    Err(SemanticError::missing_event_definition())
+                };
             }
-        }
-
-        return Ok(());
+            _ => Err(SemanticError::event_invocation_expected()),
```
cburgdorf, line 383: I was wondering if this should rather be handled on the parser level. Emit is defined as 

```
Emit {
        value: Node<Expr<'a>>,
    },
```

So, every expression is possible but I'm not entirely sure if that's what we want. I guess in terms of AST nodes this could also be defined as:

```
Emit {
        value: Node<EventExpr<'a>>,
    },
```

Where `EventExpr` would then already match against the call syntax. So something like `emit foo` (instead of `foo()` could be rejected directly on the parsing stage. The same is true for the initialization of structs I think.

g-r-a-n-t: I'm pretty indifferent here.

Another idea is that we could go with a slightly more natural syntax like `MyEvent(42, "foo", [0,1,3,4]).emit()`, which would let us remove the `emit` statement entirely.

g-r-a-n-t (approved):

analyzer/src/traversal/functions.rs
cburgdorf: >which would let us remove the `emit` statement entirely

There's something I prefer about the `emit MyEvent()` syntax. It might just be that it stands out more due to the extra keyword being used. That can make it easier to spot in a code base.

Anyway, I'm going to take it as is for now but I'm happy to revisit that in the future.

PR #332, opened 2021-03-24
Give the AST ownership of its strings.

g-r-a-n-t:
### What was wrong?

Before, the AST used slices that point to parts of the source string. This is not compatible with the lowering pass, since we will be adding nodes to the AST.

part of #306 

### How was it fixed?

Refactored the AST to use `String`s instead of `&str`s.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/traversal/functions.rs
+++ b/analyzer/src/traversal/functions.rs
@@ -47,7 +47,7 @@ pub fn func_def(
         body: _,
     } = &def.kind
     {
-        let name = name.kind;
+        let name = &name.kind;
         let function_scope = BlockScope::from_contract_scope(name, Rc::clone(&contract_scope));
 
         let is_public = qual.is_some();
@@ -104,7 +104,7 @@ pub fn func_body(
     {
         let host_func_def = contract_scope
             .borrow()
-            .function_def(name.kind)
+            .function_def(&name.kind)
             .unwrap_or_else(|| panic!("Failed to lookup function definition for {}", &name.kind));
 
         // If the return type is an empty tuple we do not have to validate any further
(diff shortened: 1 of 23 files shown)


analyzer/src/traversal/functions.rs
```diff
@@ -71,7 +71,7 @@ pub fn func_def(
         let attributes: FunctionAttributes = contract_scope
             .borrow_mut()
             .add_function(
-                name,
+                &name,
```
g-r-a-n-t, line 74: already a ref

parser/src/parsers.rs
```diff
@@ -764,12 +769,12 @@ pub fn arr_dim(input: Cursor) -> ParseResult<Node<usize>> {
     let (input, num_tok) = number_token(num_input)?;
     let (input, r_bracket) = op("]")(input)?;
 
-    let n: usize = match num_tok.string.parse() {
+    let n: usize = match num_tok.string.to_string().parse() {
```
g-r-a-n-t, line 772: unnecessary find-replace

parser/src/parsers.rs
```diff
@@ -764,12 +769,12 @@ pub fn arr_dim(input: Cursor) -> ParseResult<Node<usize>> {
...
         Ok(n) => n,
         Err(_) => {
             return Err(ParseError::str(
                 num_input,
-                &format!("invalid integer literal \"{}\"", num_tok.string),
+                &format!("invalid integer literal \"{}\"", num_tok.string.to_string()),
```
g-r-a-n-t, line 777: also unecessary

parser/src/tokenizer/types.rs
```diff
@@ -54,3 +54,9 @@ impl<'a> From<&Token<'a>> for Node<&'a str> {
         Node::new(tok.string, tok.span)
     }
 }
+
```
g-r-a-n-t, line 57: delete the impl above this

g-r-a-n-t:
Another small change with lots of small implications. Overall I think it looks cleaner without all the lifetimes.

cburgdorf (approved):
:+1:

PR #335, opened 2021-03-24
Reject non-string assert reasons with TypeError

cburgdorf:
### What was wrong?

We do currently not enforce that assert reasons are of type string.

### How was it fixed?

- added type check
- added test

### DIFF
--- a/analyzer/src/traversal/functions.rs
+++ b/analyzer/src/traversal/functions.rs
@@ -395,8 +395,10 @@ fn assert(
     if let fe::FuncStmt::Assert { test, msg } = &stmt.kind {
         verify_is_boolean(Rc::clone(&scope), Rc::clone(&context), test)?;
         if let Some(msg) = msg {
-            // TODO: type check for a string once strings are supported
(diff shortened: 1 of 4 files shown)


sbillig (approved):
Looks good to me! This might be a case where the AST should be more strict, and not allow any Expr here in the first place. Or maybe we want to make asserts support format strings someday, like `assert(x, "{} is broken", foo)`?

cburgdorf:
>This might be a case where the AST should be more strict, and not allow any Expr here in the first place

We currently allow any expression which evaluates to a string which is inline with Python syntax and allows something like this:

```
contract Foo:

    def get_assert_reason(error_code: u256) -> string100:
      if error_code == 1:
        return string100("meh, bad")
      elif error_code == 2:
        return string100("meh, worse")
      else:
        return string100("I don't even")

    pub def foo(val: u256, other_val: u256):
        assert val != other_val, self.get_assert_reason(other_val)
```

Not trying to promote that as a pattern but having the flexibility of allowing the assert reason to come from any kind of expression strikes me as a net positive.

PR #336, opened 2021-03-24
Reject invalid emit

### LINKED ISSUE #211: internal error: entered unreachable code at semantics/src/traversal/expressions.rs:141:5
Compiling this file:

```
contract Foo:
    event MyEvent:
        my_u8s: u8[255]
    pub def set_my_u8s(my_u8s: u8[255]):
        emit MyEvent(self.my_bytes.to_mem)
```

```
$ RUST_BACKTRACE=1 target/debug/fe finish.fe 
Warning: bytecode output requires 'solc-backend' feature. Try `cargo build --release --features solc-backend`. Skipping.
thread 'main' panicked at 'internal error: entered unreachable code', semantics/src/traversal/expressions.rs:141:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/core/src/panicking.rs:50:5
   3: fe_semantics::traversal::expressions::expr
   4: fe_semantics::traversal::expressions::assignable_expr
   5: fe_semantics::traversal::expressions::call_arg
Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer (we've found more than 50 bugs (that were worth fixing, and have been fixed) in solc using this afl variant) + afl.rs.

This might be a more interesting type of bug, if you think fe is ready for fuzzing at all.

cburgdorf:
### What was wrong?

We currently crash on code such as `emit foo()()` which is mostly due because we don't properly reject it in the parser stage (See https://github.com/ethereum/fe/issues/337)

### How was it fixed?

Fixed it with an extra check in the analyzer pass and filed #337 because the parser is currently getting rewritten.

### DIFF
--- a/analyzer/src/traversal/functions.rs
+++ b/analyzer/src/traversal/functions.rs
@@ -355,10 +355,10 @@ fn emit(
     } = &stmt.kind
     {
         return match kind {
-            fe::Expr::Call { func, args } => {
-                let event_name = &expressions::expr_name_string(func)?;
+            fe::Expr::Call { func, args } if matches!(&func.kind, fe::Expr::Name(_)) => {
+                let event_name = expressions::expr_name_string(func)?;
 
-                return if let Some(event) = scope.borrow().contract_event_def(event_name) {
+                return if let Some(event) = scope.borrow().contract_event_def(&event_name) {
                     context.borrow_mut().add_emit(stmt, event.clone());
 
                     let argument_attributes = args
--- a/compiler/tests/compile_errors.rs
+++ b/compiler/tests/compile_errors.rs
@@ -29,6 +29,7 @@ use std::fs;
     case("duplicate_var_in_contract_method.fe", "AlreadyDefined"),
     case("emit_undefined_event.fe", "MissingEventDefinition"),
     case("emit_without_call.fe", "EventInvocationExpected"),
+    case("emit_with_invalid_call.fe", "EventInvocationExpected"),
     case("external_call_type_error.fe", "TypeError"),
     case("external_call_wrong_number_of_params.fe", "WrongNumberOfParams"),
     case("indexed_event.fe", "MoreThanThreeIndexedParams"),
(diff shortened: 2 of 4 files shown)


analyzer/src/traversal/functions.rs
```diff
@@ -355,10 +355,10 @@ fn emit(
     } = &stmt.kind
     {
         return match kind {
-            fe::Expr::Call { func, args } => {
-                let event_name = &expressions::expr_name_string(func)?;
+            fe::Expr::Call { func, args } if matches!(&func.kind, fe::Expr::Name(_)) => {
```
cburgdorf, line 358: This feels like a band aid fix and builds more evidence that we should change the representation of `emit` in our AST. It's currently represented as:

```
Emit {
        value: Node<Expr>,
    },
```

But at least for now the right hand side can not be any `Expr` so it should probably not be represented as such but rather as something with just `name` and `args`.

Unless..at some point we would allow something like:

```
pub def some_action():
  emit get_actual_event()

def get_actual_event(some_condition) -> MyEvent:
  # Some checks
  return MyEvent(..)
```

But it's not clear to me if that would ever be possible (e.g. it isn't in solidity afaik).

Anyway, I did not go for the AST and parser change but for this band aid instead because its probably not worth investing time into parser changes when its going to be rewritten anyway.

I will open an issue about the parser change instead.

cburgdorf: Created an issue here https://github.com/ethereum/fe/issues/337

g-r-a-n-t (approved):

Issue #326, opened 2021-03-22
Pretty-print AST


g-r-a-n-t:
### What is wrong?

We cannot print ASTs. This is useful for at least two things:

- Formatting source.
- Printing the AST after lowering.

### How can it be fixed

Take a look at how rustc [does it](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/index.html?search=pretty).

PR #338, opened 2021-03-25
Lowering pass and augmented assignments

g-r-a-n-t:
Added a lowering pass to the compiler module and used it to support augmented assignments.

e.g. `foo += 1` becomes `foo = foo + 1`

### To-Do

- [x] Clean up commit history


compiler/src/yul/mappers/expressions.rs
```diff
@@ -278,7 +279,12 @@ pub fn expr_bin_operation(context: &Context, exp: &Node<fe::Expr>) -> yul::Expre
                 }
                 _ => unreachable!(),
             },
-            _ => unimplemented!(),
+            BinOperator::FloorDiv => match typ {
```
g-r-a-n-t, line 282: @cburgdorf do you think we need both `//` and `/`?

compiler/src/lowering/mappers/contracts.rs
```diff
@@ -1 +1,32 @@
...
+                ContractStmt::FuncDef { .. } => functions::func_def(context, stmt),
+            })
+            .collect();
+
+        return Node::new(
+            fe::ModuleStmt::ContractDef {
+                name,
+                body: lowered_body,
+            },
+            Span::new(0, 0),
```
g-r-a-n-t, line 27: use the original span

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,47 @@
...
+use fe_parser::ast as fe;
+use fe_parser::node::Node;
 
+/// Builds a Yul expression from a Fe expression.
+pub fn expr(context: &Context, exp: Node<fe::Expr>) -> Node<fe::Expr> {
+    let lowered_kind = match exp.kind {
+        fe::Expr::Name(_) => exp.kind,
+        fe::Expr::Num(_) => exp.kind,
+        fe::Expr::Bool(_) => exp.kind,
+        // TODO finish
```
g-r-a-n-t, line 11: finish

compiler/src/lowering/mappers/functions.rs
```diff
@@ -1 +1,124 @@
...
+            or_else: multiple_stmts(context, or_else),
+        },
+        fe::FuncStmt::Assert { test, msg } => fe::FuncStmt::Assert {
+            test: expressions::expr(context, test),
+            msg: expressions::optional_expr(context, msg),
+        },
+        fe::FuncStmt::Expr { value } => fe::FuncStmt::Expr {
+            value: expressions::expr(context, value),
+        },
+        fe::FuncStmt::Pass => stmt.kind.to_owned(),
```
g-r-a-n-t, line 88: `to_owned` not needed

compiler/src/lowering/mappers/functions.rs
```diff
@@ -1 +1,124 @@
...
+}
+
+fn aug_assign(context: &Context, stmt: Node<fe::FuncStmt>) -> fe::FuncStmt {
+    if let fe::FuncStmt::AugAssign { target, op, value } = stmt.kind {
+        let lowered_target = expressions::expr(context, target);
+        let lowered_value = expressions::expr(context, value.clone());
+
+        let new_value_kind = fe::Expr::BinOperation {
+            left: Box::new(lowered_target.clone().new_id()),
+            op: op.to_owned(),
```
g-r-a-n-t, line 111: remove all unnecessary `to_owned`s

compiler/tests/fixtures/features/aug_assign.fe
```diff
@@ -0,0 +1,61 @@
+contract Foo:
```
g-r-a-n-t, line 1: add compilation failure tests

compiler/tests/features.rs
```diff
@@ -1186,3 +1186,35 @@ fn self_address() {
         );
     });
 }
+
+#[rstest(
```
g-r-a-n-t, line 1190: add test that validates the lowering of this file

compiler/src/lowering/mappers/functions.rs
```diff
@@ -1 +1,124 @@
...
+            body: lowered_body,
+        };
+
+        return Node::new(lowered_kind, def.span);
+    }
+
+    unreachable!()
+}
+
+fn func_stmt(context: &Context, stmt: Node<fe::FuncStmt>) -> Node<fe::FuncStmt> {
```
g-r-a-n-t, line 34: this should lower to a vec of nodes

compiler/src/yul/mappers/expressions.rs
cburgdorf: I don't think we need `//`. My understanding is that Python provides this because it only provides one numeric base type and via `/` and `//` it lets you switch between *floating point division* and *integer division*. We currently only have *integer division* and if we ever add another base type to represent numbers with a decimal point then one would be free to use that datatype to perform such a division operations. So it is the chosen base type that would decide about it, not the operator.

analyzer/src/traversal/assignments.rs
```diff
@@ -50,6 +54,33 @@ pub fn assign(
     unreachable!()
 }
 
+pub fn aug_assign(
```
g-r-a-n-t, line 57: docs

compiler/src/lowering/mappers/contracts.rs
```diff
@@ -1 +1,29 @@
+use fe_analyzer::Context;
 
+use crate::lowering::mappers::functions;
+use fe_parser::ast as fe;
+use fe_parser::ast::ContractStmt;
+use fe_parser::node::Node;
+
+pub fn contract_def(context: &Context, stmt: Node<fe::ModuleStmt>) -> Node<fe::ModuleStmt> {
```
g-r-a-n-t, line 8: docs

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
+use fe_analyzer::Context;
+use fe_parser::ast as fe;
+use fe_parser::node::Node;
 
+/// Builds a Yul expression from a Fe expression.
```
g-r-a-n-t, line 5: fix docs

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+        fe::Expr::UnaryOperation { op, operand } => fe::Expr::UnaryOperation {
+            op,
+            operand: boxed_expr(context, operand),
+        },
+        fe::Expr::CompOperation { left, op, right } => fe::Expr::CompOperation {
+            left: boxed_expr(context, left),
+            op,
+            right: boxed_expr(context, right),
+        },
+        // TODO finish
```
g-r-a-n-t, line 47: remove comment

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+        fe::Expr::Call { func, args } => fe::Expr::Call {
+            func: boxed_expr(context, func),
+            args: call_args(context, args),
+        },
+        fe::Expr::List { .. } => unimplemented!(),
+        fe::Expr::ListComp { .. } => unimplemented!(),
+        fe::Expr::Tuple { .. } => exp.kind,
```
g-r-a-n-t, line 54: add comment explaining why this is like this

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+        fe::Expr::Str(_) => exp.kind,
+        fe::Expr::Ellipsis => unimplemented!(),
+    };
+
+    Node::new(lowered_kind, exp.span)
+}
+
+pub fn slices_index_expr(
```
g-r-a-n-t, line 62: private

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+                first_slice.span,
+            )],
+            slices.span,
+        );
+    }
+
+    unreachable!()
+}
+
+pub fn optional_expr(context: &Context, exp: Option<Node<fe::Expr>>) -> Option<Node<fe::Expr>> {
```
g-r-a-n-t, line 81: docs

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+    exp.map(|exp| expr(context, exp))
+}
+
+#[allow(clippy::boxed_local)]
+pub fn boxed_expr(context: &Context, exp: Box<Node<fe::Expr>>) -> Box<Node<fe::Expr>> {
```
g-r-a-n-t, line 86: docs

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,119 @@
...
+    Box::new(expr(context, *exp))
+}
+
+pub fn multiple_exprs(context: &Context, exp: Vec<Node<fe::Expr>>) -> Vec<Node<fe::Expr>> {
```
g-r-a-n-t, line 90: docs

compiler/src/lowering/mappers/functions.rs
```diff
@@ -1 +1,127 @@
...
+            body: lowered_body,
+        };
+
+        return Node::new(lowered_kind, def.span);
+    }
+
+    unreachable!()
+}
+
+fn func_stmt(context: &Context, stmt: Node<fe::FuncStmt>) -> Vec<Node<fe::FuncStmt>> {
```
g-r-a-n-t, line 31: Note: returning a vec here because we'll have lowerings like:

 ```rust
(foo, bar): (u256, u256) = my_tuple
```

to

```rust
foo: u256 = my_tuple.item0
bar: u256 = my_tuple.item1
```

which requires that we be able to map a single statement to multiple statements.

compiler/src/lowering/mappers/functions.rs
```diff
@@ -1 +1,127 @@
...
+}
+
+fn aug_assign(
+    context: &Context,
+    target: Node<fe::Expr>,
+    op: Node<fe::BinOperator>,
+    value: Node<fe::Expr>,
+) -> Vec<fe::FuncStmt> {
+    let lowered_target = expressions::expr(context, target);
+    let lowered_value = expressions::expr(context, value.clone());
```
g-r-a-n-t, line 113: clean up the naming in here and get rid of the clone

compiler/src/lowering/mod.rs
```diff
@@ -7,6 +7,6 @@ mod mappers;
 mod names;
 
 /// Lowers the Fe source AST to a Fe HIR AST.
-pub fn lower(_context: &Context, module: &FeModuleAst) -> FeModuleAst {
-    module.clone()
+pub fn lower(_context: &Context, module: FeModuleAst) -> FeModuleAst {
+    mappers::module::module(_context, module)
```
g-r-a-n-t, line 11: context is used

parser/src/node.rs
```diff
@@ -61,6 +61,11 @@ impl<T> Node<T> {
             span,
         }
     }
+
+    pub fn new_id(mut self) -> Self {
```
g-r-a-n-t, line 65: docs

src/main.rs
```diff
@@ -134,11 +134,15 @@ fn write_compiled_module(
     fs::create_dir_all(output_dir).map_err(ioerr_to_string)?;
 
     if targets.contains(&CompilationTarget::Ast) {
-        write_output(&output_dir.join("module.ast"), &module.fe_ast)?;
+        write_output(&output_dir.join("module.ast"), &module.src_ast)?;
```
g-r-a-n-t, line 137: add a new target for the lowered AST

compiler/src/yul/mappers/expressions.rs
```diff
@@ -278,7 +278,7 @@ pub fn expr_bin_operation(context: &Context, exp: &Node<fe::Expr>) -> yul::Expre
                 }
                 _ => unreachable!(),
             },
-            _ => unimplemented!(),
+            fe::BinOperator::FloorDiv => unimplemented!(),
```
g-r-a-n-t, line 281: see #341

compiler/tests/fixtures/features/aug_assign.fe
```diff
@@ -0,0 +1,57 @@
...
```
g-r-a-n-t, line 1: It would be nice to have some tests that check that source files are correctly lowered. We need to be able to pretty-print ASTs for this, tho. See #326.

analyzer/src/traversal/assignments.rs
```diff
@@ -50,6 +54,36 @@ pub fn assign(
...
+    context: Shared<Context>,
+    stmt: &Node<fe::FuncStmt>,
+) -> Result<(), SemanticError> {
+    if let fe::FuncStmt::AugAssign {
+        target,
+        op: _,
+        value,
+    } = &stmt.kind
+    {
+        let target_attributes = expressions::expr(Rc::clone(&scope), Rc::clone(&context), target)?;
```
g-r-a-n-t, line 69: relevant: https://github.com/ethereum/fe/issues/345

compiler/src/lowering/mappers/functions.rs
sbillig: This is a good candidate for `SmallVec<[Node<..>, 1]>` in some possible future performance optimization.

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -1 +1,123 @@
...
+}
+
+/// Lowers and optional expression.
+pub fn optional_expr(context: &Context, exp: Option<Node<fe::Expr>>) -> Option<Node<fe::Expr>> {
+    exp.map(|exp| expr(context, exp))
+}
+
+/// Lowers a boxed expression.
+#[allow(clippy::boxed_local)]
+pub fn boxed_expr(context: &Context, exp: Box<Node<fe::Expr>>) -> Box<Node<fe::Expr>> {
```
sbillig, line 89: This is another pattern that might be worth optimizing in the future, or might not matter; we take a boxed expr, probably don't modify it, and rebox. An extra allocation for every expr.

sbillig (approved):
Looks good. My only (mild) concern is potential performance issues with all the allocations happening. Probably won't matter in practice, and if it does we can work on that later. (See inline comments above).

compiler/src/lowering/mappers/functions.rs
g-r-a-n-t: Makes sense. I'll track this in an issue.

Issue #70, opened 2020-10-04
Create an online editor for people to try out Fe


cburgdorf:
### What is wrong?

This doesn't strictly belong here but for the lack of a better place and to make sure to remember it, I'll file it here as a documentation issue.

We want an online playground for people to try out Fe similar to e.g. https://play.rust-lang.org for Rust.
@pipermerriam recently pointed out that it makes sense to prioritize this and get something up early even at a stage where the language isn't yet fully functional.

### How can it be fixed

There are generally two possible ways:

1. Frontend + Backend where the frontend sends code over to a backend that runs the code and returns the compiler output back to the frontend

2. Compile the compiler to WASM and have it sit in the frontend directly, eliminating the need for a backend.

g-r-a-n-t:
I'd prefer option 2.

Have you thought about the contract interaction UX? It seems like this can become quite complex.

It might be easiest to just start with displaying the compiled Yul code and nicely printing any errors.

pipermerriam:
Looks like you might just look at integration with remix via their plugin system.

vasa-develop:
I can help with this. Can somebody point me to the resources regarding this:
```
Compile the compiler to WASM and have it sit in the frontend directly, eliminating the need for a backend.
```

cburgdorf:
>I can help with this.

@vasa-develop That would be really awesome!

So, the general idea is that Rust code can be compiled to WASM code which can then be run natively in the browser. Some general resources:

- https://webassembly.org/
- https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm
- https://rustwasm.github.io/docs/book/

I have never compiled rust to WASM myself and so there are a lot of unknowns to me like:

- Do we have any dependencies that would prevent us from compiling to WASM?
- Would the generated WASM code be too large to be used in browser?
- unknown unknowns

One thing I know, is that we actually run some kind of WASM test in CI but frankly I have no idea what it does:

https://github.com/ethereum/fe/blob/703e31b5c0cf781f9cb10121b2a49d64896cbb19/Makefile#L15-L21

Maybe @davesque can shed some light on this issue because he created this and might be able to give us some directions?

@vasa-develop 
[...]

PR #344, opened 2021-03-29
[NEVER MERGE] PoC work: add playground

Y-Nak:
This is PoC work for Fe playground, ***never merge this PR***. 
I succeeded in running Fe compiler and EVM on the browser, though there are still some limitations.
ref: #70

Screenshot_2021-03-30 Fe playground(1)

## How to run
1. Install `wasm-pack` and `npm`
2. `cd playground` and `wasm-pack build`
3. `cd www`, `npm install` and `npm run start`

Currently, the playground can't work on chrome because of an issue related to `solc-js`,  please use firefox.

### [solc-js](https://github.com/ethereum/solc-js)
`solc-js` doesn't work on the browser because of [some issues](https://github.com/ethereum/solc-js/issues?q=is%3Aissue+is%3Aopen+browserhttps://github.com/ethereum/solc-js/issues?q=is%3Aissue+is%3Aopen+browser), so we need to wrap `solc-js` somehow. I used [browser-solc](https://github.com/ericxtang/browser-solc) as a wrapper for now, but I think it would be better to write a new wrapper instead of using `browser-solc` because this also has some issues.

### dynamically sized value
When input the below code, `panicked at 'tried to get the static encoding size of dynamically sized value', compiler/src/yul/operations/abi.rs:82:25` is reported(all panic information is output to console).
```
contract Bar:
    pub def __main__() -> string100:
        return string100("hi!")
```

### error report
Currently, the type of `compiler::errors::CompileError` is actually `Cow<'static, str>`  and `compiler::compile` converts all error types to it, this is quite inconvenient to implement a user-friendly error report on the playground. 
For example, 
```
contract Foo:
    pub def __main__() -> u8:
        return u16(1)
```
would produce
``` 
TypeError on line 3
pub def __main__() -> u8:
        [31mreturn u16(1)[0m
```
on the browser. 

## TODOs
1.  Address known issues
2. Allow to run arbitrary functions and pass args to it(Currently the implementation runs only   `__main__`  without args)
3. UI/UX improvements

## Misc
I'm really dreaming of Yul compiler written by pure Rust. I'll start implementing the PoC of it as a private OSS activity. Please notify me if you are interested in this project.


cburgdorf:
Woah! :heart_eyes:  That's seriously cool!
Just a heads up that I'm on vacation this and next week (and @g-r-a-n-t also from next week) so if we are slow to respond it's just that. I'm super happy to see this progressing :pray:

cburgdorf:
>I'm really dreaming of Yul compiler written by pure Rust. I'll start implementing the PoC of it as a private OSS activity. Please notify me if you are interested in this project.

We are definitely interested. I think I have heard @g-r-a-n-t dreaming of that, too. While I share the excitement about it I just want to highlight that we currently also get the YUL optimizer for free. Also, security wise, it's a nice property to be able to say that we share the last mile of the processing pipeline with Solidity. I think that could help us gain some reputation in terms of safety, or mitigate concerns. But apart from that, I fully agree that it would be nice to have a native YUL compiler in Rust.

cburgdorf:
@PascalPrecht As you are about to look into the website I just wanted to raise awareness that a wasm compiler for an in-browser compiler is in the works. Nothing actionable, just hinting :)

Y-Nak:
I'll rework it sometime later and send another PR.

PR #347, opened 2021-03-30
Binary operation checks.

### LINKED ISSUE #345: Tighter checking on binary operations.
### What is wrong?

We are not [checking binary operations](https://github.com/ethereum/fe/blob/master/analyzer/src/traversal/expressions.rs#L408) very thoroughly. For example, addition over two string would make it past the analyzer.

### How can it be fixed

We should match the operation and then check the types. This should be performed in the [operations module](https://github.com/ethereum/fe/blob/5006695511248d99ec388c68f02837c5edd2b701/analyzer/src/namespace/operations.rs#L1) (which should be moved one module down), this way the same checking function can be used for augmented assignments.

g-r-a-n-t:
### What was wrong?

fixes #345 

### How was it fixed?

- added a checking method for binary operations to the `operations` module
- used this function to check binary operations and augmented assignments
- added more compiler error tests

### To-Do

- [x] Clean up commit history


analyzer/src/operations.rs
```diff
@@ -0,0 +1,178 @@
...
+            Err(SemanticError::type_error())
+        }
+    } else {
+        Err(SemanticError::type_error())
+    }
+}
+
+fn bin_bit_shift(left: &Type, right: &Type) -> Result<Type, SemanticError> {
+    if let (Type::Base(Base::Numeric(left)), Type::Base(Base::Numeric(right))) = (left, right) {
+        // The right side must be signed.
```
cburgdorf, line 104: Looks like the right side must be **un**signed

sbillig (approved):

PR #350, opened 2021-03-31
Parser support for `pub const`; add `fields` to ast::ContractDef

sbillig:
### What was wrong?

1) The AST and parser didn't allow for `pub const` fields in structs or contracts (only one or the other)
2) The parser allowed contract fields to be intermixed with function and event defs. 
3) Grant thought it would be handy to have contract fields in their own vec in the `ContractDef` struct.

### How was it fixed?

via some AST and parser refactoring

### To-Do

- [x] Clean up commit history


parser/src/parsers.rs
```diff
@@ -365,17 +365,23 @@ pub fn contract_def(input: Cursor) -> ParseResult<Node<ModuleStmt>> {
...
-    let (input, body) = many1(contract_stmt)(input)?;
+    let (input, fields) = many0(field)(input)?;
+    let (input, body) = many0(contract_stmt)(input)?;
     let (input, _) = dedent_token(input)?;
 
-    let last_stmt = body.last().unwrap();
+    let last_stmt = body
+        .last()
+        .map(|node| node.span)
+        .or(fields.last().map(|node| node.span))
```
g-r-a-n-t, line 375:
```
error: use of `or` followed by a function call
   --> parser/src/parsers.rs:375:10
    |
375 |         .or(fields.last().map(|node| node.span))
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `or_else(|| fields.last().map(|node| node.span))`
    |
    = note: `-D clippy::or-fun-call` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
```

g-r-a-n-t (approved):
looks good

PR #355, opened 2021-04-13
Update g++ in CI system dependencies

cburgdorf:
### What was wrong?

CI is borked

### How was it fixed?

1. A bit of cleanup, moving the repeated installation steps into a single script
2. `g++-8` appears to not be available in `ubuntu-latest` but that's ok, changed to `g++-10` instead
3. Temporary fixing rust nightly to a nightly from mid-march as it appears the current nightly errors which might be temporary or means we will need to cut a new `solc-rust` and/or `yultsur` release with updated dependencies
4. Created an issue to unpin the nightly in the future when the issues are resolved: https://github.com/ethereum/fe/issues/357
5. Sneaked in a minor CI perf improvement by removing an unnecessary build step from the linting job

### DIFF
--- a/.github/install_deps.sh
+++ b/.github/install_deps.sh
@@ -1,3 +1,3 @@
 sudo apt-get install -y libboost-all-dev
-sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-8 50
-sudo update-alternatives --set g++ "/usr/bin/g++-8"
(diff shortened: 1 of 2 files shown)


cburgdorf:
I rage merged this forgoing a review to unbreak CI. I've put #357 on the Mai milestone to unsure we don't forget to unpin the rust nightly again.

Issue #362, opened 2021-04-21
'no entry found for key' at compiler/src/yul/mappers/module.rs:25:46


agroce:
Compiling this file:

```
contract Foo:
 pub def o()->address:
        foo:Foo=Foo.create2(2,0)
        return address(foo)
```

```
thread 'main' panicked at 'no entry found for key', compiler/src/yul/mappers/module.rs:25:46
stack backtrace:
   0:     0x55bc7b0e6390 - std::backtrace_rs::backtrace::libunwind::trace::h9d49145f95eb5894
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55bc7b0e6390 - std::backtrace_rs::backtrace::trace_unsynchronized::hab1d020365bb6864
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55bc7b0e6390 - std::sys_common::backtrace::_print_fmt::h7659588431e304bd
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55bc7b0e6390 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h09f4a9e3befae3c7
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x55bc7b1012ec - core::fmt::write::hf3fdfde304b9a088
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/fmt/mod.rs:1092:17
   5:     0x55bc7b0e32f2 - std::io::Write::write_fmt::h1cb850689c7116f0
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/io/mod.rs:1568:15
   6:     0x55bc7b0e8555 - std::sys_common::backtrace::_print::hdbccd5aa093ba544
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:49:5
   7:     0x55bc7b0e8555 - std::sys_common::backtrace::print::hc639c4f320222558
[...]

PR #363, opened 2021-04-22
Properly reject circular dependency when using create/create2

cburgdorf:
### What was wrong?

As found by the fuzzer in #362, circular dependencies when using `create` or `create2` were not correctly rejected yet.

### How was it fixed?

In the analyzer we check if the name of the contract to create is the same as the contract of the inherited scope. If so, we reject it with a newly added `CircularDependency` kind.

### DIFF
--- a/compiler/tests/fixtures/compile_errors/circular_dependency_create.fe
+++ b/compiler/tests/fixtures/compile_errors/circular_dependency_create.fe
@@ -1,0 +1,5 @@
+contract Foo:
+    pub def bar() -> address:
+        foo: Foo = Foo.create(0)
+
+        return address(foo)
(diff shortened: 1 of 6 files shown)


cburgdorf:
I can't yet assign @satyamakgec as a reviewer (assigning issues does work though)

compiler/tests/fixtures/compile_errors/circular_dependency_create.fe
```diff
@@ -0,0 +1,5 @@
+contract Foo:
+  pub def bar()->address:
+    foo:Foo=Foo.create(0)
```
g-r-a-n-t, line 3: np: format the code

g-r-a-n-t (approved):
:+1:

PR #364, opened 2021-04-24
Stable rust

sbillig:
### What was wrong?

Fe requires nightly rust, but there's no particularly compelling reason for that.

### How was it fixed?

- Replaced `#![feature(int_error_matching)]` with the lexical-core crate.
- Replaced `#![feature(external_doc)]` with cargo-readme. This isn't quite the same, but it works well enough. README.md is now generated from the main.rs doc comment, which now contains the readme content. (`make README.md` will run the appropriate command, if you first `cargo install cargo-readme`)
- Removed the nightly rustfmt options. I deleted rustfmt.toml. max_width was set to the default; imports_layout, wrap_comments, and comment_width are nightly-only. This changes the `use` lines in most files, of course.

(Both of the compiler features above are likely to be stabilized eventually; I'd be happy to switch back when they are. No idea whether the rustfmt options ever will be.)

We could use stable rustc and nightly rustfmt, but that would basically guarantee that new contributors will fail the rustfmt lint check. I think the rustfmt defaults are fine; probably 99% of rust projects use the defaults.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,11 +31,11 @@ jobs:
           target
           !target/**/*serde*
         key: ${{ runner.OS }}-build-v2-${{ hashFiles('**/Cargo.lock') }}
-    - name: Install latest nightly
+    - name: Install rust
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: nightly-2021-03-15
+        toolchain: stable
         override: true
     - name: coverage with tarpaulin
       run: |
@@ -58,11 +58,11 @@ jobs:
           target
           !target/**/*serde*
         key: ${{ runner.OS }}-build-v2-${{ hashFiles('**/Cargo.lock') }}
-    - name: Install latest nightly
+    - name: Install rust
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: nightly-2021-03-15
+        toolchain: stable
         override: true
         components: rustfmt, clippy
     - name: Validate release notes entry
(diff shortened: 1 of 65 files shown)


cburgdorf:
I'm a little bit indifferent about this. Having to include the Readme inside `main.rs` makes me a bit sad. What's the big advantage we get? Sure, if Fe was a library we would want people to be able to use it with stable Rust but that's not the case here. It's not even on cargo (because of the git dependencies) and users will end up consuming Fe via binary distribution. Right now, it's mostly the Readme that makes me sad but in the future it may be some other nice upcoming Rust feature that makes our life easier that we can't use just because we restrict ourselves to stable rust. I'm not terribly against this but I want to bring up these concerns...what does @g-r-a-n-t think?

sbillig:
I'm mostly indifferent too, but slightly in favor of using stable. Regarding the readme, I think it would be fine if the readme and the rust api docs don't have the same content. We could just scrap the `cargo readme` thing.

The benefits IMO are:
- nightly is a moving target, so builds might randomly fail for contributors or CI unless we use a rust_toolchain file and/or specify a nightly date in the CI config
- stable is thoroughly tested. Nightly could conceivably introduce a bug into the fe binaries that we distribute (though not terribly likely)
- some people *may* perceive fe to be less trustworthy if it's built with a non-stable rust :shrug:

I'm not suggesting that we restrict ourselves to stable forever. If some great nightly feature comes along that we really want, we could certainly switch back. Our current feature use doesn't justify it though, IMO.

g-r-a-n-t:
My thinking was that we would switch to stable at some point before our `1.0.0` release. Seeing as this may be in the not-so-distant future, I think we might as well merge this.

And about the git dependencies.. I'll ask the owner about moving them again. It would be really nice to have these libraries on crates.io.

sbillig:
I'll leave this up to you @cburgdorf, I'm totally fine either way. It was a very small amount of work; I'm happy to resurrect this closer to fe 1.0 if you'd prefer.

cburgdorf (approved):

cburgdorf:
Alright, sounds good. Merged :tada:

PR #365, opened 2021-04-26
Tighten AST for emit statements

cburgdorf:
### What was wrong?

As explained in #337 our current AST allows any expression to follow the `emit` keyword and the analyzer ensures
that what follows actually is an event call. Clearly the analyzer is currently handling something that should rather be rejected at the parsing stage.

### How was it fixed?

1. The `emit` AST was changed to not take a `value` of `Expr` but rather a `name` of `String` and as well as a list of `CallArg`.
2. Parser, analyzer and lowering were changed accordingly
3. The analyzer error `EventInvocationExpected` was removed

### To-Do

- We should have parser tests for this but they are out of scope for this PR

### DIFF
--- a/compiler/tests/compile_errors.rs
+++ b/compiler/tests/compile_errors.rs
@@ -36,8 +36,6 @@ use rstest::rstest;
     case("duplicate_var_in_child_scope.fe", AlreadyDefined),
     case("duplicate_var_in_contract_method.fe", AlreadyDefined),
     case("emit_undefined_event.fe", MissingEventDefinition),
-    case("emit_without_call.fe", EventInvocationExpected),
-    case("emit_with_invalid_call.fe", EventInvocationExpected),
     case("external_call_type_error.fe", TypeError),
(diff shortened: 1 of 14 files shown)


compiler/tests/compile_errors.rs
```diff
@@ -36,8 +36,6 @@ use rstest::rstest;
     case("duplicate_var_in_child_scope.fe", AlreadyDefined),
     case("duplicate_var_in_contract_method.fe", AlreadyDefined),
     case("emit_undefined_event.fe", MissingEventDefinition),
-    case("emit_without_call.fe", EventInvocationExpected),
-    case("emit_with_invalid_call.fe", EventInvocationExpected),
```
cburgdorf, line 40: @sbillig These two analyzer errors aren't needed anymore because the AST will simply not allow `emit` to be followed by arbitrary expressions (which we previously rejected in the analyzer pass). But do we have a way yet to check for expected parser errors so that I can ensure we won't regress in the future.

sbillig (approved):
Looks good!

Issue #279, opened 2021-03-01
Remove Ellipsis from AST


cburgdorf:
### What is wrong?

The `Ellipsis` Type in our AST is based on the `Ellipsis` syntax in Python which is used for:

a.) placeholders

```
def foo():
   ...
```

b.) slicing

`a[...,0]`

We don't intend to use the syntax for placeholders and we are more likely to follow rust for the slicing syntax. Therfore, we should remove `Ellipsis` from the code base for now.

### How can it be fixed

Remove it from here and all usages:

https://github.com/ethereum/fe/blob/177ee35ba4b12150f5cb14152e066d364b1a0161/parser/src/ast.rs#L267

PR #374, opened 2021-04-28
Remove stub support for Ellipsis

cburgdorf:
### What was wrong?

As mentioned in #279 we do have some stub code for the Ellipsis `...` syntax which we carried over from our Python heritage but don't actually plan to support.

### How was it fixed?

Removed all affected code


sbillig (approved):
:+1:

PR #375, opened 2021-04-28
Use insta for parser ast tests, remove defunct tokenizer tests

sbillig:
Just some quality-of-life stuff. Adding new tests is easier, comparing/accepting output changes is super nice with `cargo insta review`, and we can see all the test cases in one file so it's easy to tell how thorough they are.

Some of the test cases are better than before. We still need more thorough tests around things like operator precedence, but the ron output makes that a pain.

~~I'm also testing a change to the tarpaulin/codecov gh action here; hoping that using a pre-compiled cargo-tarpaulin with shave about 5 minutes off the coverage run.~~ (actions-rs/install has issues https://github.com/actions-rs/install/issues/12; I'll revisit this or actions-rs/tarpaulin another time)

### To-Do
- [x] Clean up commit history


g-r-a-n-t (approved):
:+1:

PR #381, opened 2021-04-29
Moved context into its own module.

g-r-a-n-t:
### What was wrong?

`Context` was cluttering the analyzer's `lib.rs` file. 

### How was it fixed?

Moved `Context` into its own module.

### To-Do

- [ ] Clean up commit history


sbillig (approved):
:+1:

PR #383, opened 2021-04-30
Speed up building and running of tests

sbillig:
### What was wrong?

Tests take a long time to run.

### How was it fixed?

Mostly by telling cargo to build solc with optimizations for dev builds. I also moved the integration tests into a single module (and thus a single binary), to cut down on the time spent linking the test binaries, removed some dependencies and updated the rest.

In practice, this will slow down the build of solc in favor of much faster test execution. I'm not exactly sure by how much, because I'm sick of timing builds :). If solc is built using sccache (which will now happen automatically, if sccache is installed), then every build after the first will be much faster than it was prior to this pr.

Times on my laptop, before:
(with rust artifacts cached via sccache, but not solc because it wasn't possible/easy prior to https://github.com/g-r-a-n-t/solc-rust/pull/2)

`cargo clean && cargo test --workspace --all-features`: 687s
`cargo test --workspace --all-features`: 85s

after:
(with rust and solc cached via sccache)
`cargo clean && cargo test --workspace --all-features` 176s
`cargo test --workspace --all-features` 18s


cburgdorf (approved):
Looks good. You know a whole lot of tricks and tools man :+1:

PR #385, opened 2021-05-03
Initial implementation of differential fuzzing

satyamakgec:
### What was wrong?
- Needs fuzzing of contracts.

### How was it fixed?
It consists of the fuzzing setup for the erc20.fe and `erc20.sol` and has the rearrangements for the compiler tests as an independent directory to avoid circular dependency and it will also decrease the test execution time as it doesn't require rebuilding the dependencies again.

### To-Do
 

- [x] Need to add fuzzing for ERC20.sol.

- [ ] Add support for output gathering to differentiate outputs of the contract.

- [ ] Add support for the gas analyzing for the contracts

[//]: # (Stay ahead of things, add list items here!)
- [ ] OPTIONAL: Update [Spec](https://github.com/ethereum/fe/blob/master/spec/index.md) if applicable
- [ ] Add entry to the [release notes](https://github.com/ethereum/fe/blob/master/newsfragments/README.md) (may forgo for trivial changes)

- [ ] Clean up commit history

PR #387, opened 2021-05-03
Approval testing for the Analyzer

g-r-a-n-t:
### What was wrong?

Testing around the Analyzer was not very thorough.

### How was it fixed?

We now generate snapshots of `Context` using insta. The snapshots are formatted using codespan and the debug implementation of attributes. Here are a few things to note:

- Nodes are labeled with the hash of their attributes. This makes it so that the set of changed nodes is clearly visible and lets us place the attributes debug further down, where it reads more easily.
- Usage of hash maps and sets were removed in favor of b trees. This makes ordering consistent between snapshots.
- The snapshot format is clear enough for someone to skim over and gain meaningful insight. 
- For now, I've copied the fixture directory into the analyzer, but when #385 is complete, we should move these tests into the testing crate and use the other fixture directory. **TODO: write an issue for this**

As an example, this is what would show up during insta review if you changed the location of numeric literals from `value` to `memory`:

insta-attributes

Notice that the same node is displayed twice. This is because the list of nodes used to generate the output is built by pushing nodes passed into `context.add_` methods. These add methods are called twice in some situations, so we end up with duplicates. I think this should be fixed by adding update methods to `Context`, so we're not calling adds multiple times. **TODO: write an issue for this**

### To-Do

- [x] Clean up commit history
- [ ] Create the issues mentioned in this comment.


analyzer/Cargo.toml
```diff
@@ -16,4 +16,6 @@ strum = { version = "0.20.0", features = ["derive"] }
 lexical-core = "0.7.6"
 
 [dev-dependencies]
-rstest = "0.6.4"
\ No newline at end of file
+insta = "1.7.1"
+rstest = "0.6.4"
+fn-fixture = "1.0.2"
```
g-r-a-n-t, line 21: delete

analyzer/src/context.rs
```diff
@@ -403,4 +412,84 @@ impl Context {
     pub fn get_module(&self) -> Option<&ModuleAttributes> {
         self.module.as_ref()
     }
+
+    pub fn get_spanned_expressions(&self) -> Vec<(Span, ExpressionAttributes)> {
```
g-r-a-n-t, line 416: docs on these

common/src/diagnostics.rs
```diff
@@ -25,7 +25,7 @@ pub fn diagnostics_string(diagnostics: &[Diagnostic], files: &FileStore) -> Stri
     let config = term::Config::default();
 
     for diag in diagnostics {
-        term::emit(&mut buffer, &config, files, &diag).unwrap();
+        term::emit(&mut buffer, &config, files, &diag).expect("emit failed");
```
g-r-a-n-t, line 28: better message

analyzer/tests/fixtures
```diff
@@ -0,0 +1 @@
+/home/grant/workshop/fe/compiler/tests/fixtures
```
g-r-a-n-t, line 1: cp not ln

analyzer/tests/analysis.rs
```diff
@@ -1,105 +1,177 @@
...
-            key: Base::Address,
-            value: Box::new(Type::Array(Array {
-                size: 100,
-                inner: Base::Byte,
-            })),
-        }),
-        Location::Storage { nonce: Some(0) },
-    )
-}
+fn build_snapshot(path: &str, src: &str, context: &Context) -> String {
```
g-r-a-n-t, line 129: add module attributes

analyzer/tests/analysis.rs
```diff
@@ -1,105 +1,181 @@
...
+    case::call_statement_without_args("features/call_statement_without_args.fe"),
+    case::checked_arithmetic("features/checked_arithmetic.fe"),
+    case::constructor("features/constructor.fe"),
+    case::create2_contract("features/create2_contract.fe"),
+    case::create_contract("features/create_contract.fe"),
+    case::create_contract_from_init("features/create_contract_from_init.fe"),
+    case::empty("features/empty.fe"),
+    case::events("features/events.fe"),
+    case::external_contract("features/external_contract.fe"),
+    case::for_loop_with_break("features/for_loop_with_break.fe"),
```
cburgdorf, line 33: Just to double check: We now test the state of the context for all these fixtures as an additional category of tests that we run. These tests do not actually test the return values of the called functions which we continue to do as before here. Is that right?

https://github.com/ethereum/fe/blob/c3b1f31f5dd01d6bb108beb856c905654771b4e2/compiler/tests/cases/features.rs#L112-L237

cburgdorf (approved):
Looks good, left a question to double check I'm not misunderstanding what this does :sweat_smile:

analyzer/tests/analysis.rs
g-r-a-n-t: Yes, we are just testing the state of the context. Anytime we change something in the analyzer that results in a different context output for a given fixture, it will show up here.

There have been a number of times where I've been reworking things in the analyzer and ended up breaking some unrelated full-compilation fixture test. Resolving these types of issues is tedious, since it's not clear what changed in the analyzer output. With this, it will be obvious what changed.

PR #388, opened 2021-05-04
Lower list expressions

cburgdorf:
### What was wrong?

We do not yet support list expressions e.g. `[1, 2, 3]`

### How was it fixed?

1. List expressions are taken care of in the lowering pass. They are lowered to a function call where in the function we do an array declaration and individual assignments.
2. Added a blanket trait with `into_node()` and `into_boxed_node()` methods to make working without significant spans less annoying 
3. Added tests for the lowering as well as for the feature itself

### DIFF
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -73,12 +73,20 @@ pub fn expr_list(
             // TODO: Right now we are only supporting Base type arrays
             // Potential we can support the tuples as well.
             if let Type::Base(base) = attribute_to_be_matched.typ {
+                let array_typ = Array {
+                    size: elts.len(),
+                    inner: base,
+                };
+
+                scope
+                    .borrow()
+                    .contract_scope()
+                    .borrow_mut()
+                    .add_used_list_expression(array_typ.clone());
+
                 return Ok(ExpressionAttributes {
-                    typ: Type::Array(Array {
-                        size: elts.len(),
-                        inner: base,
-                    }),
-                    location: attribute_to_be_matched.location,
+                    typ: Type::Array(array_typ),
+                    location: Location::Memory,
                     move_location: None,
                 });
             }
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -641,15 +641,17 @@ fn expr_call_self_attribute(
     func_name: &str,
     args: &Node<Vec<Node<fe::CallArg>>>,
 ) -> Result<ExpressionAttributes, SemanticError> {
+    let called_func = scope
+        .borrow()
+        .contract_scope()
+        .borrow()
+        .function_def(func_name);
+
     if let Some(ContractFunctionDef {
         params,
         return_type,
         ..
-    }) = scope
-        .borrow()
-        .contract_scope()
-        .borrow()
-        .function_def(func_name)
+    }) = called_func
     {
         let argument_attributes = expr_call_args(Rc::clone(&scope), Rc::clone(&context), args)?;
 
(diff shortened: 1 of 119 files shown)


sbillig:
BTW, the parser doesn't allow newlines inside of square brackets at the moment. I made an issue for it #389.

analyzer/src/traversal/expressions.rs
```diff
@@ -641,15 +649,17 @@ fn expr_call_self_attribute(
...
     if let Some(ContractFunctionDef {
         params,
         return_type,
         ..
-    }) = scope
-        .borrow()
-        .contract_scope()
-        .borrow()
-        .function_def(func_name)
+    }) = called_func
```
cburgdorf, line 662: Moving this out of the `if let` was needed to mitigate a `borrow_mut()` failure...this took me way to long to find :sweat_smile:

compiler/src/lowering/mappers/expressions.rs
```diff
@@ -163,3 +165,32 @@ fn expr_tuple(context: &Context, exp: Node<fe::Expr>) -> fe::Expr {
...
+                .into_iter()
+                .map(|list_val| fe::CallArg::Arg(list_val).into_node())
+                .collect::<Vec<_>>()
+                .into_node();
+
+            let _self: &'static str = Object::Self_.into();
+            // Turn List Expression into a function call
+            return fe::Expr::Call {
+                func: fe::Expr::Attribute {
+                    value: fe::Expr::Name(_self.to_string()).into_boxed_node(),
```
cburgdorf, line 186: I'm sure there's a nicer way to get the string of the `Object::Self` that would allow me to move it back inline here but I wasn't able to find it. /cc @sbillig

compiler/tests/fixtures/features/return_sum_list_expression_1.fe
```diff
@@ -0,0 +1,8 @@
+contract Foo:
+
+    pub def bar() -> u256:
+        my_array: u256[3] = [10, 20, 30]
```
g-r-a-n-t, line 4: what happens if we, say, use an array of size 20? :imp: 

Im wondering if we have stack problems during yul compilation.

analyzer/src/traversal/expressions.rs
g-r-a-n-t: yeah, Ive ran into this too

compiler/tests/fixtures/features/return_sum_list_expression_1.fe
cburgdorf: Oh yeah, I missed to check that...I'll check it.

g-r-a-n-t (approved):
looks good :+1:

compiler/src/lowering/mappers/expressions.rs
sbillig: I guess the best way is probably to derive `strum::ToString` instead of (or in addition to) `IntoStaticStr`, and do `fe::Expr::Name(Object::Self_.to_string())`

compiler/tests/fixtures/features/return_sum_list_expression_1.fe
cburgdorf: Updated the tests, appears to be working fine :sweat_smile:

compiler/src/lowering/mappers/expressions.rs
cburgdorf: haha, it's that simple, thanks!

compiler/tests/fixtures/features/return_sum_list_expression_1.fe
g-r-a-n-t: bruns

PR #392, opened 2021-05-05
Panic if context.add_ methods are called twice.

g-r-a-n-t:
### What was wrong?

The context was letting us add attributes to the same ID multiple times.

### How was it fixed?

- added panics to each of the add methods
- fixed two spots that panicked

All of the snapshot noise is from there no longer being any duplicate expressions printouts.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/context.rs
+++ b/analyzer/src/context.rs
@@ -319,9 +319,27 @@ impl Context {
     }
 
     /// Attribute contextual information to an expression node.
+    ///
+    /// # Panics
+    ///
+    /// Panics if an entry already exists for the node id.
     pub fn add_expression(&mut self, node: &Node<fe::Expr>, attributes: ExpressionAttributes) {
         self.add_node(node);
-        self.expressions.insert(node.id, attributes);
+        expect_none(
+            self.expressions.insert(node.id, attributes),
+            "expression attributes already exist",
+        );
+    }
+
+    /// Update the expression attributes.
+    ///
+    /// # Panics
+    ///
+    /// Panics if an entry does not already exist for the node id.
+    pub fn update_expression(&mut self, node: &Node<fe::Expr>, attributes: ExpressionAttributes) {
+        self.expressions
+            .insert(node.id, attributes)
+            .expect("expression attributes do not exist");
     }
 
     /// Get information that has been attributed to an expression node.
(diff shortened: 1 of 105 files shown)


analyzer/src/context.rs
```diff
@@ -318,7 +318,15 @@ impl Context {
     /// Attribute contextual information to an expression node.
     pub fn add_expression(&mut self, node: &Node<fe::Expr>, attributes: ExpressionAttributes) {
         self.add_node(node);
-        self.expressions.insert(node.id, attributes);
+        expect_none(
+            self.expressions.insert(node.id, attributes),
+            "expression attributes already exist",
+        );
+    }
+    pub fn update_expression(&mut self, node: &Node<fe::Expr>, attributes: ExpressionAttributes) {
```
g-r-a-n-t, line 326: docs

analyzer/src/context.rs
```diff
@@ -318,7 +318,15 @@ impl Context {
     /// Attribute contextual information to an expression node.
     pub fn add_expression(&mut self, node: &Node<fe::Expr>, attributes: ExpressionAttributes) {
```
g-r-a-n-t, line 319: add panic docs

compiler/src/lowering/mappers/functions.rs
```diff
@@ -129,14 +129,15 @@ fn aug_assign(
...
 ) -> Vec<fe::FuncStmt> {
-    let lowered_target = expressions::expr(context, target);
     let original_value_span = value.span;
-    let lowered_value = expressions::expr(context, value);
+    let lowered_target = expressions::expr(context, target.clone());
+    let lowered_lh_value = expressions::expr(context, target);
+    let lowered_rh_value = expressions::expr(context, value);
 
     let new_value_kind = fe::Expr::BinOperation {
-        left: Box::new(lowered_target.clone().new_id()),
```
g-r-a-n-t, line 137: Copying the lowered expression created two separate AST branches with the same nested IDs, which resulted in us calling `add_expression` twice for the same ID. It wasn't necessarily causing a problem, but may have in more complex scenarios.

We now copy the target expression and lower it twice so we have different IDs.

analyzer/src/traversal/expressions.rs
```diff
@@ -65,7 +65,7 @@ pub fn expr_list(
             // Rationale - A list contains only one type of element.
             // No need to iterate the whole list if the first attribute doesn't match next
             // one.
-            for elt in elts.iter() {
+            for elt in elts.iter().skip(1) {
```
g-r-a-n-t, line 68: We were re-analyzing the first element.

cburgdorf (approved):
LGTM

PR #394, opened 2021-05-07
Bring minimal viable website to life

cburgdorf:
### What was wrong?

We've received an amazing website draft from @PascalPrecht that we now need to integrate with docs, fill with actual content and just iterate on to reach a minimal viable version that can go live.

Well...strictly speaking it **is** already live at http://fe.ethereum.org as I'm just treating that as my staging ground for now (we don't promote that URL anywhere yet).

### How was it fixed?

- Created a website directory in the root that hosts the html and logos
  - The logos are actually duplicated from the "logos" directory but as far as git bloat is concerned that isn't a problem since git deduplicates everything on a level of binary blobs
- Turned the existing docs sites and spec into an mdBook
- Migrated the `deployWebsite.sh` into a few new `make` tasks
  - `make build-website` -> Build the website, docs, api-docs into `target/website`
  - `make serve-website` -> Build and serve on `localhost:8000`
  - `make deploy-website` -> Deploy to live site
  - It's rudimentary but gets the job done for now
- Code of Conduct
- Created a Quickstart stub
- Came up with actual texts for the various design elements and linked to places that make sense


cburgdorf:
@g-r-a-n-t @sbillig This is ready for review. It's not entirely finished yet. E.g. the Quickstart doesn't have any actual content yet but I created a stub site for it. I plan to spend some more time on that. I actually want to write a proper tutorial that even deploys some code to the Görli network. But I think there's enough in this PR to get it in and then keep iterating.

website/index.html
```diff
@@ -0,0 +1,294 @@
...
+  17
+  18
+  19
+  20
+  21
+  22
+  23
+
+                      </div>
+                      <code class="flex-auto relative block text-white pt-4 pb-4 px-4 overflow-auto">contract Ownable:
```
g-r-a-n-t, line 181: I think we should use guest book demo here, since ownable is used as an inherited.

website/index.html
```diff
@@ -0,0 +1,294 @@
...
+      </nav>
+    </div>
+  </header> 
+  <main role="main">
+    <section class="px-4 py-12 pt-12 sm:py-16 lg:pt-24 max-w-screen-lg xl:max-w-screen-xl mx-auto md:flex justify-between lg:justify-around items-center bg-red-500 rounded-sm">
+      <img class="h-52 lg:h-72 lg:ml-8 xl:ml-0 hidden sm:block" src="fe-logo.svg" alt="Fe Programming language">
+      <div class="sm:mt-14 md:ml-16 md:mt-0 max-w-3xl lg:max-w-2xl xl:max-w-3xl xl:-ml-4">
+        <h2 class="text-5xl sm:text-6xl xl:text-6xl leading-none font-semibold tracking-tight">The <span class="text-gray-100">next generation</span><br> smart contract language for Ethereum</h2>
+        <p class="text-lg sm:text-2xl text-gray-700 leading-6 sm:leading-8 mt-4">Create <span class="text-gray-900 font-semibold">decentralized applications</span> in a powerful, future-proof and <span class="text-gray-900 font-semibold">statically typed</span> language that is very <span class="text-gray-900 font-semibold">easy to learn</span>.<p>
```
g-r-a-n-t, line 58: IMO the `text-gray-700` and `text-gray-900` colors are a bit to close. Also, I think we want 900 to pop a bit more, so maybe that one should be the lighter shade.

website/index.html
cburgdorf: Ah good point. Eventually it would be nice to demo something that uses the new `string<100>` syntax when it is ready. I'll use the guest book for now.

cburgdorf: I've put in some comments and syntax highlighting (Python for now) because the example is shorter and I needed it to consume a bit more space to fit in nicely

image

website/index.html
cburgdorf: I increased the contrast between them but switching them didn't look good to me. Probably because then we would not have the contrast to the headline above. But if truly suck at design work so if there is potential for optimizations I would ask Pascal post-merge to give it a make over.

sbillig:
Looks great! I think we should add a note that fe is alpha software and not ready for production use, near the download button. Or a flashing 'under construction' gif.

website/index.html
g-r-a-n-t: looks good :+1:

cburgdorf:
@g-r-a-n-t Thanks for the suggestions. I took them all :+1: 

So, are we good to merge this? I'll keep working on Quickstart, Tutorial and other guides for a bit but I can ping Pascal post-merge to check my changes and maybe tweak them design wise.

g-r-a-n-t:
LGTM

g-r-a-n-t (approved):

Issue #378, opened 2021-04-28
Remove or support `else` block from `for` and `while


sbillig:
### What is wrong?

The ast and parser allow for an `else` block following `for` and `while` loops. IMO, this is a weird pythonism that we shouldn't adopt. The `else` is executed if no `break` stopped the loop. (I had to look this up; my guess was that it would execute if the loop never executed, like shorthand for `if x { while x { ... } } else { ... }`.)

```
for x in xs:
  if x > 10:
    break
else:
  print("didn't break")
```

### How can it be fixed

Remove the relevant field on the ast structs, tidy up code that uses those structs accordingly, remove support from the parser.

g-r-a-n-t:
@cburgdorf and I discussed this a while back and reached the same conclusion, but I guess never got around to writing an issue.

cburgdorf:
Yep, we should remove that :+1:

Issue #341, opened 2021-03-26
Remove FloorDiv


g-r-a-n-t:
I don't think we need `//`. My understanding is that Python provides this because it only provides one numeric base type and via `/` and `//` it lets you switch between *floating point division* and *integer division*. We currently only have *integer division* and if we ever add another base type to represent numbers with a decimal point then one would be free to use that datatype to perform such a division operations. So it is the chosen base type that would decide about it, not the operator.

_Originally posted by @cburgdorf in https://github.com/ethereum/fe/pull/338#discussion_r602104185_

PR #395, opened 2021-05-07
AST cleanup

sbillig:
### What was wrong?

The AST contains stuff we don't plan to support, or don't yet support. This muddies up code that operates on the ast. In most cases, the compiler panics when processing fe code that uses these unimplemented features; it should output nice error messages instead.

### How was it fixed?

Changes:

- Already discussed: for/while else block (#378), floordiv (#341), ListComp, Is/IsNot
- removed In/NotIn. I guess we'll probably support these someday, or maybe we'll decide that collection types should have member functions that accomplish the same thing without a special operator.
- Subscript now has a single index Expr, with no accommodation for slices. I think if/when we decide to implement slicing, we should follow rust's lead and create a Range expr and type, which could be used here and in `for x in 0..5`.
- Assign now has a single target, rather than a vec of targets. I'm not sure if this was for chained assignment (`x = y = 10`) or tuple destructuring (`x, y = foo()`). I don't think we need to support the former; the latter we should only allow with parentheses, which would be a single tuple expr.
- Expr::Str now holds a single string, instead of a vec of strings. Python's `a = "hi " "mom"` is weird. I don't think we should implement it.
- The `for` target is now a name string instead of an expr, and anything other than a name here is a syntax error. When we implement tuple destructuring, this should probably be changed to a `VarDeclTarget`.
- I scrapped `Node<PubQualifier>` etc in favor of a bare bool.

I'm happy to revert any of the above if anyone is strongly opposed.

### To-Do

- [ ] Clean up commit history


compiler/src/yul/mappers/expressions.rs
```diff
@@ -31,8 +31,7 @@ pub fn expr(context: &Context, exp: &Node<fe::Expr>) -> yul::Expression {
             fe::Expr::UnaryOperation { .. } => expr_unary_operation(context, exp),
             fe::Expr::CompOperation { .. } => expr_comp_operation(context, exp),
             fe::Expr::Call { .. } => expr_call(context, exp),
-            fe::Expr::List { .. } => unimplemented!(),
-            fe::Expr::ListComp { .. } => unimplemented!(),
+            fe::Expr::List { .. } => panic!("list expressions should be lowered"),
```
cburgdorf, line 34: Ah good point, I forgot that in my PR where I lowered the list expressions :+1:

cburgdorf (approved):
Great cleanup! :+1:

PR #403, opened 2021-05-11
Fill in Quickstart in Fe guide

cburgdorf:
### What was wrong?

A small collection of things around the website:
  - The download button is linking to the release page introducing an extra barrier before the actual file
  - We have some stubs in our Fe guide such as the Quickstart that needs to be filled in.
  - The guestbook example could be slightly improved

### How was it fixed?

1. The buttons now link to the actual binary. To do that the build pipeline reads the current compiler version and replaces a marker in the website. The website shows the correct button according to the users OS *unless* if we don't support the OS in which case we will link to the release page so that the user can see what kind of OS are supported.

2. The guestbook example makes the `map` private and renames it to be not repeat the contract name
3. Created Quickstart with Installation & First contract guide (deployment guide has yet to follow)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ build-website:
 	# Copy the entire website directory to target/website
 	cp -r website/ target/
 
+	# Fill version marker with current version
+	sed -i "s/{{FE_VERSION}}/$$(cargo pkgid | cut -d# -f2 | cut -d: -f2)/g" target/website/index.html
+
 	# Generate the compiler API docs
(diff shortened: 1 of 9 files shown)


cburgdorf:
The state of this PR is already live at https://fe.ethereum.org/. I know that I can't *test in production*™ forever. I will setup my own little github pages site that I can deploy to give previews. Bear with me :sweat_smile:

g-r-a-n-t (approved):

PR #410, opened 2021-05-20
Allow hex/octal/binary radix numeric literals

### LINKED ISSUE #333: Allow hexadecimal/octal/binary numeric literals

Y-Nak:
### What was wrong?
closes #333 

### How was it fixed?
1. Added `common::numeric::Literal` to abstract string represented numeric literals.
2. Consistently use `BigInt` for semantic analysis, this approach is the same as `rustc` does (it holds `u128` for numeric integers).
2. Removed `lexical_core` because it doesn't provide `from_str_radix` like methods. And also, I think we don't need to check the details of `IntErrorKind` in almost all cases. This is because, looking at the definition of [`IntErrorKind`](https://doc.rust-lang.org/std/num/enum.IntErrorKind.html) in detail, `IntErrorKind::Empty` is the only error kind that we have to consider it an internal compiler error.

### To-Do
We may need to reconsider the usage of `BigInt`, I think it would be better to use [`U256`](https://docs.rs/primitive-types/0.9.0/primitive_types/struct.U256.html) because:
1. it provides more reliability than multiple-precision arithmetic.
2. it improves performance a little bit.

---
- [x] Clean up commit history

### DIFF
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -14,8 +14,10 @@ use builtins::{
     BlockField, ChainField, ContractTypeMethod, GlobalMethod, MsgField, Object, TxField,
     ValueMethod,
 };
+use fe_common::numeric;
 use fe_parser::ast as fe;
 use fe_parser::node::Node;
+use num_bigint::BigInt;
 use std::convert::TryFrom;
 use std::rc::Rc;
 use std::str::FromStr;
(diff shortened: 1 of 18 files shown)


analyzer/src/traversal/expressions.rs
```diff
@@ -945,21 +950,27 @@ fn validate_are_kw_args(args: &[Node<fe::CallArg>]) -> Result<(), SemanticError>
...
+            let num = numeric::Literal::new(num)
+                .parse::<BigInt>()
+                .map_err(|_| SemanticError::numeric_literal_expected())?;
+            return Ok(-num);
         }
     } else if let fe::Expr::Num(num) = &value.kind {
-        return Ok(num.to_string());
+        let num = numeric::Literal::new(num)
+            .parse::<BigInt>()
+            .map_err(|_| SemanticError::numeric_literal_expected())?;
```
sbillig, line 966: Should a parse failure here just panic? It'd be indicative of a bug in the lexer/parser, right?

sbillig (approved):
Looks great!

analyzer/src/traversal/expressions.rs
Y-Nak: Exactly! I fixed it.

PR #415, opened 2021-05-21
Dynamically sized return support for external calls

### LINKED ISSUE #321: tried to get the static encoding size of dynamically sized value at compiler/src/yul/runtime/functions/contracts.rs:67:26
Compiling this file:

```
contract C:
 pub def r(t:address,n:u256)->bool:return false
contract r:
 pub def g()->string100:return string100("")
```

```
thread 'main' panicked at 'failed to get the static encoding size: "tried to get the static encoding size of dynamically sized value"', compiler/src/yul/runtime/functions/contracts.rs:67:26
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/option.rs:1300:5
   3: fe_compiler::yul::runtime::functions::contracts::calls::{{closure}}
   4: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   5: alloc::vec::source_iter_marker::<impl alloc::vec::spec_from_iter::SpecFromIter<T,I> for alloc::vec::Vec<T>>::from_iter
   6: fe_compiler::yul::runtime::functions::contracts::calls
   7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   8: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   9: fe_compiler::yul::runtime::build
  10: fe_compiler::yul::runtime::build_with_abi_dispatcher
  11: fe_compiler::yul::mappers::contracts::contract_def
  12: fe_compiler::yul::mappers::module::module
  13: fe_compiler::yul::compile
Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

### LINKED ISSUE #400: tried to get the static encoding size of dynamically sized value at compiler/src/yul/operations/abi.rs:78:25
Compiling this file:

```
contract C:
 pub def r(t:address,n:u256)->bool:
  return false
contract r:
 pub def g()->string100:
  return string100("")
```

```
thread 'main' panicked at 'tried to get the static encoding size of dynamically sized value', compiler/src/yul/operations/abi.rs:78:25
stack backtrace:
   0:     0x5580744b0710 - std::backtrace_rs::backtrace::libunwind::trace::h9d49145f95eb5894
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x5580744b0710 - std::backtrace_rs::backtrace::trace_unsynchronized::hab1d020365bb6864
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5580744b0710 - std::sys_common::backtrace::_print_fmt::h7659588431e304bd
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x5580744b0710 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h09f4a9e3befae3c7
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x5580744ca46c - core::fmt::write::hf3fdfde304b9a088
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/core/src/fmt/mod.rs:1092:17
   5:     0x5580744ad672 - std::io::Write::write_fmt::h1cb850689c7116f0
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/io/mod.rs:1568:15
   6:     0x5580744b28d5 - std::sys_common::backtrace::_print::hdbccd5aa093ba544
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:49:5
   7:     0x5580744b28d5 - std::sys_common::backtrace::print::hc639c4f320222558
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/sys_common/backtrace.rs:36:9
   8:     0x5580744b28d5 - std::panicking::default_hook::{{closure}}::hdb012dd7a485bb5d
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:208:50
   9:     0x5580744b2383 - std::panicking::default_hook::h75facbce77b6ba91
                               at /rustc/9c09c1f7cfcf9de0522bcd1cfda32b552195c464/library/std/src/panicking.rs:225:9
  10:     0x558073be1b4f - fe_compiler::errors::DEFAULT_PANIC_HOOK::{{closure}}::{{closure}}::hd71e839699e0bfd6

Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

### LINKED ISSUE #405: Support dynamically sized return values when calling external contracts.
See discussion on #321:

# "
Yes, this is on the right track.

It boils down to the fact that we don't know how much memory will be used in the call return for dynamically sized types. It's not necessarily a bug, just a missing feature that is not handled well.

`call(g, a, v, in, insize, out, outsize)`

> call contract at address a with input mem[in…(in+insize)) providing g gas and v wei and output area mem[out…(out+outsize)) returning 0 on error (eg. out of gas) and 1 on success

See the Yul [docs](https://docs.soliditylang.org/en/latest/yul.html#evm-dialect).

I see two possible solutions:
1. We add a `max_size` value to `AbiType` and use that instead of this `static_encode_size` function. We of course know the max size of `string3`, since the type is bounded. One drawback to this is that we'll have trouble dealing with external contracts imported using their JSON ABIs, since dynamically sized types have no bounds there. It would definitely work in the short-term, tho.
2. We set a very high `outsize` and then only allocate what's used. I'm not quite sure if `outsize` influences memory access gas costs :thinking:.

I think for now we should just go with (1), since importing external contracts using their json ABIs is not happening anytime soon.
# "

g-r-a-n-t:
### What was wrong?

closes #405 
closes #400 
closes #321 

I also noticed that there was a bug preventing us from calling functions on `contract` fields (e.g. `self.my_contract`).

### How was it fixed?

Used `returndatasize` and `returndatacopy` to deal with sizes that aren't known at compile-time (it's now much simpler).

See the note at the bottom of this [section](https://docs.soliditylang.org/en/v0.8.4/yul.html#evm-dialect):

> The call* instructions use the out and outsize parameters to define an area in memory where the return or failure data is placed. This area is written to depending on how many bytes the called contract returns. If it returns more data, only the first outsize bytes are written. You can access the rest of the data using the returndatacopy opcode. If it returns less data, then the remaining bytes are not touched at all. You need to use the returndatasize opcode to check which part of this memory area contains the return data. The remaining bytes will retain their values as of before the call.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -661,8 +661,18 @@ fn expr_call_value_attribute(
     if let fe::Expr::Attribute { value, attr } = &func.kind {
         let value_attributes = expr(Rc::clone(&scope), Rc::clone(&context), &value)?;
 
-        if let Type::Contract(contract) = value_attributes.typ {
-            return expr_call_contract_attribute(scope, context, contract, &attr.kind, args);
+        if let Type::Contract(contract) = &value_attributes.typ {
+            // We must ensure the expression is loaded onto the stack.
+            context
+                .borrow_mut()
+                .update_expression(value, value_attributes.clone().into_loaded()?);
+            return expr_call_contract_attribute(
+                scope,
+                context,
+                contract.to_owned(),
+                &attr.kind,
+                args,
+            );
         }
 
         // for now all of these function expect 0 arguments
@@ -790,12 +800,6 @@ fn expr_call_contract_attribute(
         .find(|function| function.name == func_name)
     {
         let return_type = function.return_type.to_owned();
-
-        if matches!(return_type, FixedSize::String(_)) {
-            // we need figure out how to deal with dynamically sized returns
-            // for now, this only affects strings
-            todo!("external call string returns")
-        }
 
         let argument_attributes = args
             .kind
--- a/compiler/src/yul/operations/abi.rs
+++ b/compiler/src/yul/operations/abi.rs
@@ -51,41 +51,6 @@ pub fn encode_size<T: AbiEncoding>(types: Vec<T>, vals: Vec<yul::Expression>) ->
     return expression! { add([static_size], [data_operations::sum(dyn_size)]) };
 }
 
-/// Returns an expression that gives the max size of the encoded values.
-///
-/// # Panics
-/// This will panic if any of the elements of the `types` vector have dynamic
-/// size.
-pub fn static_encode_size<T: AbiEncoding>(types: Vec<T>) -> yul::Expression {
-    let mut static_size = 0;
-
-    for typ in types {
-        match typ.abi_type() {
-            AbiType::Uint { .. } => static_size += 32,
-            AbiType::Array { inner, size } => {
-                let inner_size = match *inner {
-                    AbiType::Uint {
-                        size: AbiUintSize { padded_size, .. },
-                    } => padded_size,
-                    AbiType::Array { .. } => todo!(),
-                    AbiType::Tuple { .. } => todo!(),
-                };
-                match size {
-                    AbiArraySize::Static { size } => {
-                        static_size += utils::ceil_32(inner_size * size)
-                    }
-                    AbiArraySize::Dynamic => {
-                        panic!("tried to get the static encoding size of dynamically sized value")
-                    }
-                }
-            }
-            AbiType::Tuple { elems } => static_size += elems.len() * 32,
-        }
-    }
-
-    literal_expression! { (static_size) }
-}
-
 /// Returns a list of expressions that can be used to decode given types.
 /// `start` is where the encoding starts and `loc` indicates whether the data is
 /// in calldata or memory.
@@ -110,10 +75,8 @@ pub fn decode<T: AbiEncoding>(
 
 #[cfg(test)]
 mod tests {
-    use crate::yul::operations::abi::{decode, encode, encode_size, static_encode_size};
-    use fe_analyzer::namespace::types::{
-        AbiDecodeLocation, Array, Base, FeString, FixedSize, U256,
-    };
+    use crate::yul::operations::abi::{decode, encode, encode_size};
+    use fe_analyzer::namespace::types::{AbiDecodeLocation, FeString, U256};
     use yultsur::*;
 
     #[test]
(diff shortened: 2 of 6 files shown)


analyzer/src/traversal/expressions.rs
```diff
@@ -661,8 +661,17 @@ fn expr_call_value_attribute(
     if let fe::Expr::Attribute { value, attr } = &func.kind {
         let value_attributes = expr(Rc::clone(&scope), Rc::clone(&context), &value)?;
 
-        if let Type::Contract(contract) = value_attributes.typ {
-            return expr_call_contract_attribute(scope, context, contract, &attr.kind, args);
+        if let Type::Contract(contract) = &value_attributes.typ {
+            context
```
g-r-a-n-t, line 665: comment here

analyzer/src/traversal/expressions.rs
```diff
@@ -661,8 +661,18 @@ fn expr_call_value_attribute(
     if let fe::Expr::Attribute { value, attr } = &func.kind {
         let value_attributes = expr(Rc::clone(&scope), Rc::clone(&context), &value)?;
 
-        if let Type::Contract(contract) = value_attributes.typ {
-            return expr_call_contract_attribute(scope, context, contract, &attr.kind, args);
+        if let Type::Contract(contract) = &value_attributes.typ {
+            // We must ensure the expression is loaded onto the stack.
+            context
+                .borrow_mut()
+                .update_expression(value, value_attributes.clone().into_loaded()?);
```
g-r-a-n-t, line 668: Originally, calls to contract fields would fail (e.g. `self.foo.get_my_string()`). This is because the call was being made to a non-existent contract at the address of the last 20 bytes of the storage pointer.

cburgdorf (approved):
Adding features by simplifying the code, I like!

PR #398, opened 2021-05-09
Use codespan for analyzer errors

sbillig:
### What was wrong?

Analyzer errors weren't fancy enough.

### How was it fixed?

This has become a behemoth of a pr, sorry for the hassle.

The context object now contains a vec of diagnostics. Analysis functions will return `Ok` with correct type information wherever possible, even if they emit an error diagnostic. In cases where correct type information can't be returned, `Err(SemanticError::fatal())` is returned. There are still a lot of old errors yet to be converted. In some cases, fancy errors will be followed by an old error, so the user experience will be a little awkward for a while. When they've all been converted to codespan diagnostics, SemanticError/ErrorKind can finally be retired.

I had to do a lot of refactoring to give myself enough context to emit pretty error messages. I've been inlining some utility functions and passing around more nodes and spans, and I'm starting to wonder if there's a better way. Eg rustc's "hir" analysis uses various query functions to retrieve such info. Maybe we could structure things similarly as part of the salsa experiment and avoid passing around spans everywhere. Either way I need to take a breather from this for a while; maybe some enlightenment will come later. Y'all are more than welcome to pick away at the remaining old errors, and/or suggest better wording for the errors I did convert. The quality of the error messages is very reflective of my energy level at the time they were written :)

I added a basic check of the left-hand side of assignments; this could be extended to verify that the lhs isn't a read-only attribute of a built-in, for example. The labels of event and struct constructor args are now verified as well. The labels of function args are still ignored (I'd like extend the same checking logic of events/structs to function calls, if the ideas in #397 are agreed upon).

Along the way, I noticed that tuples inside maps weren't being lowered and would thus panic; couldn't resist fixing that, even though it could have been a separate pr.

closes #408
closes #382

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] OPTIONAL: Update [Spec](https://github.com/ethereum/fe/blob/master/spec/index.md) if applicable
- [x] Add entry to the [release notes](https://github.com/ethereum/fe/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [x] Clean up commit history

PR #416, opened 2021-05-24
Basic type inference of int literals and empty arrays

### LINKED ISSUE #256: Empty list reaches unreachable!()
Compiling this file:

```
contract Ownable:
  _owner: address
  pub def renounceOwnership():
    self._owner < address([ ])
```

```
thread 'main' panicked at 'internal error: entered unreachable code', analyzer/src/traversal/expressions.rs:105:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/c0b64d97beebb09325b5587abed39f4f1621026f/library/core/src/panicking.rs:50:5
   3: fe_analyzer::traversal::expressions::expr
   4: fe_analyzer::traversal::expressions::assignable_expr
   5: fe_analyzer::traversal::expressions::call_arg
Aborted
```

```
[profile.dev]
debug = true
panic = "abort"
```

for fuzzing.  Using https://github.com/agroce/afl-compiler-fuzzer + afl.rs.

sbillig:
### What was wrong?

Int literals can only be used where a u256 is expected unless they're wrappen in an explicit type constructor call. For example:
```
a: u8 = 100 # type error
def f(x: u8): ...
f(100) # type error
f(u8(100)) # ok
```
 
Also closes #256 (panic on `[]`).

### How was it fixed?

Some of the analyzer traversal fns now take an `expected_type` parameter. This is used to infer the type of int literals and the element type of arrays. Array size mismatch is still an error, so `x: u8[10] = []` is a type error, but at least it doesn't panic.

This isn't applied to strings. `x: string10 = "hi"` is still a type error. This will presumably require some work to not break compilation. This also doesn't coerce ints that aren't literals. Eg`y: u8 = ... \n x: u16 = y` is still a type error.

Tangential changes:
- Removed Shared wrapper from Context
- Moved the Unit type into Base, so empty arrays with no expected type can have inner type `Base::Unit` (this should only happen if the empty array is a standalone expr statement)

### To-Do

- [x] Clean up commit history
- [ ] Rebase when #398 is merged

### DIFF
--- a/analyzer/src/context.rs
+++ b/analyzer/src/context.rs
@@ -40,7 +40,6 @@ impl Location {
             Type::String(_) => Ok(Location::Memory),
             Type::Struct(_) => Ok(Location::Memory),
             Type::Map(_) => Err(SemanticError::cannot_move()),
-            Type::Unit => Ok(Location::Value),
         }
     }
 }
@@ -88,7 +87,7 @@ impl From<Shared<ContractScope>> for ContractAttributes {
                     is_public: def.is_public,
                     name: name.clone(),
                     params: def.params.to_owned(),
-                    return_type: FixedSize::Unit,
+                    return_type: FixedSize::unit(),
                 })
             }
         }
@@ -159,7 +158,6 @@ impl ExpressionAttributes {
     pub fn into_loaded(mut self) -> Result<Self, SemanticError> {
         match self.typ {
             Type::Base(_) => {}
-            Type::Unit => {}
             Type::Contract(_) => {}
             _ => return Err(SemanticError::cannot_move()),
         }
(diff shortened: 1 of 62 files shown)


sbillig:
I'll have this rebased on master shortly, but the changes here are fairly simple so feel free to take a look.

sbillig:
Reopened as #429, which targets the master branch.

Issue #273, opened 2021-02-26
Change array type syntax to Array<u8, 100>


cburgdorf:
### Array type definiton syntax

The syntax that we currently use for array definitions was copied from Vyper and looks like this:

```
some_ints: u8[10] # A fixed size array for 10 values of type u8
cars: Car[10] # A fixed size array for 10 values of type Car
```

The Rust syntax for these would be:

```
some_ints: [u8; 10] // A fixed size array for 10 values of type u8
cars: [Car; 10] // A fixed size array for 10 values of type Car
```

I want to point out that we usually look to Python and Rust for inspiration regarding syntax and that Python doesn't have fixed size array type and the closest would be a list which one would type out using generics (e.g. `List[int]` which would translate to `List<u8>` in Fe). In other words, since we can't really lookup to Python for inspiration in this case, it makes sense to me to take inspiration from Rust instead.

Apart from just having *personal preference* for the Rust syntax I also think it is clearer and probably easier to parse when we get to more complex cases. E.g consider an array of tuples.

Current syntax

```
some_array: (u8, u8)[10] # A fixed size array for 10 values of type (u8, u8)
```

Rust syntax
```
some_array: [(u8, u8); 10] # A fixed size array for 10 values of type (u8, u8)
```
The Rust syntax feels more self contained (being enclosed in square brackets) and I expect it to be easier to parse as well.

cburgdorf:
Actually, with the introduction of const generics which we first try to sneak in for builtin types such as `string<N>` I wonder if it would make sense to change our array type to `array<T, N>`, e.g. `array<u8, 10>`

sbillig:
We discussed this on the weekly call; consensus is to use Array<u8, 100> (and use title case for all reference types).

davesque:
Just wanted to comment and say that I love seeing all this interesting use of the generics syntax.  When I was working on the original Vyper project, it always felt like we were running up against the limitations of having no control over the parser.  I love seeing the benefits of removing that obstacle.

I figured at one point, I also had the maybe sort of wacky idea of supporting generic keyword parameters.  So you could do things like this:
```python
Array<u8, maxlen=100>  # internally, a dynamic length array but with extra machinery to restrict length
# or
Array<u8, len=100>  # a constant length array
```

[...]

PR #425, opened 2021-05-27
Allow generic arguments in `Expr::Call`

### LINKED ISSUE #379: Replace `stringN` type names with `String<N>`
### What is wrong?

The parser will parse `string<100>` etc as a `TypeDesc::Generic`, so it's time to retire the old `string100` syntax in favor of the const generic style. We should also do `bytes<N>`, if we're keeping the bytes type. (See #258)

Y-Nak:
### What was wrong?
fixes #379 

### How was it fixed?
1. Add `BTParser` which thinly wraps `Parser` and enables backtracking
2. Add `generic_args` to `ast::Expr::Call`
3. Add `if_chain` for improving readability

### To-Do
* https://github.com/ethereum/fe/issues/273

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -677,6 +677,7 @@ name = "fe-parser"
 version = "0.4.0-alpha"
 dependencies = [
  "fe-common",
+ "if_chain",
  "insta",
  "logos",
  "serde",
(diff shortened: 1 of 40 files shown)


sbillig (approved):
@Y-Nak This looks really great. The BtParser idea is very elegant.

Issue #361, opened 2021-04-21
Add support for pragma (like) keyword


cburgdorf:
### What is wrong?

We need something like a `pragma` statement to ensure that a certain piece of code is only compiled with a specific version (or version range) of the Fe compiler.

### How can it be fixed

1. Settle on a keyword + syntax
  1.1 https://github.com/npm/node-semver#advanced-range-syntax
2. Check if the expected version (range) matches the actual version (range) and reject compilation if it doesn't.

cburgdorf:
I'm going to work on this

PR #426, opened 2021-05-28
Add support for pragma statement

cburgdorf:
### What was wrong?

As described in #361 developers need to have a way to express that a piece of code can only be compiled with a compiler where the version matches a given requirement.

### How was it fixed?

- Added support for `pragma` keyword in the parser
- Version requirement is parsed using the `semver` crate
- The Analyzer checks the compiler version against the version requirement and raises a `WrongCompilerError` if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,7 @@ dependencies = [
  "num-bigint",
  "num-traits",
  "rstest",
+ "semver 1.0.0",
  "strum",
  "vec1",
 ]
@@ -720,6 +721,7 @@ dependencies = [
  "if_chain",
  "insta",
  "logos",
+ "semver 1.0.0",
  "serde",
  "unescape",
  "uuid",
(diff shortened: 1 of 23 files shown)


sbillig:
More use cases for this:

```
`#![forbid(dynamic)] // disallow any code that prevents gas analysis
```

```
#[target(evm)]
fn deploy() -> Foo:
  return Foo::create2(1, 2)

#[target(ewasm)]
fn deploy() -> Foo:
  ...
```

sbillig:
Sorry, I realize this discussion should have happened *before* you started implementing this. We can always change the syntax later, the important part is that we have the compiler version lock functionality.

cburgdorf:
You are raising good points here. But as you say, we have enough time to change the syntax at a later point in time. My immediate focus is to make sure we support rejecting code that was written to be compiled with a different compiler version so I just followed soliditys `pragma <version-requirement>` syntax (where `<version-requirement>` is any semver range syntax that is also valid on `Cargo.toml` files).

sbillig (approved):
Looks great. It'd probably be good to have an analyzer error test of the version mismatch case.

cburgdorf:
>It'd probably be good to have an analyzer error test of the version mismatch case.

Good point, added!

Issue #432, opened 2021-06-01
Convert remaining old-style analyzer errors to new style error messages


cburgdorf:
### What is wrong?

With https://github.com/ethereum/fe/pull/398/files the analyzer was taught how to produce context-rich error messages in the same style that the parser produces them (using `context.fancy_error(...)`. The PR changes a lot but not all of them.

### How can it be fixed

Continue porting them over.

PR #433, opened 2021-06-01
Migrate remaining old style errors

cburgdorf:
### What was wrong?

As described in #432 we have a bunch of analyzer errors that still need to be migrated to the new system

### How was it fixed?

This doesn't migrate all errors but a substantial enough chunk of them to be wrapped up for now.

General path taken:

1. Added call to `context.fancy_error(...)`
2. Returned a `SemanticError::fatal()` if the error prevents us from returning meaningful information
3. Removed migrated semantic error type
4. If needed, created a standalone error to carry an error up the call stack in cases where we have to raise errors from a place that doesn't allow us to create a context rich error

The general direction of the migration is to eventually get rid of `SemanticError` entirely and just have a standalone `FatalError` struct that would be returned from the analyzer.

### DIFF
--- a/analyzer/src/constants.rs
+++ b/analyzer/src/constants.rs
@@ -1,0 +1,1 @@
+pub const MAX_INDEXED_EVENT_FIELDS: usize = 3;
--- a/analyzer/src/errors.rs
+++ b/analyzer/src/errors.rs
@@ -4,6 +4,10 @@ use ansi_term::Color::Red;
 use fe_common::diagnostics::Diagnostic;
 use fe_parser::node::Span;
 
+/// Error to be returned from APIs that should reject duplicate definitions
+#[derive(Debug)]
+pub struct AlreadyDefined;
+
 #[derive(Debug)]
 pub struct AnalyzerError {
     pub diagnostics: Vec<Diagnostic>,
@@ -13,7 +17,6 @@ pub struct AnalyzerError {
 /// Errors for things that may arise in a valid Fe AST.
 #[derive(Debug, PartialEq)]
 pub enum ErrorKind {
-    AlreadyDefined,
     CannotMove,
     NotCallable,
     NotSubscriptable,
@@ -75,14 +78,6 @@ impl SemanticError {
     pub fn cannot_move() -> Self {
         SemanticError {
             kind: ErrorKind::CannotMove,
-            context: vec![],
-        }
-    }
-
-    /// Create a new error with kind `AlreadyDefined`
-    pub fn already_defined() -> Self {
-        SemanticError {
-            kind: ErrorKind::AlreadyDefined,
             context: vec![],
         }
     }
--- a/analyzer/src/errors.rs
+++ b/analyzer/src/errors.rs
@@ -7,6 +7,9 @@ use fe_parser::node::Span;
 /// Error to be returned from APIs that should reject duplicate definitions
 #[derive(Debug)]
 pub struct AlreadyDefined;
+/// Error indicating that a value can not move between memory and storage
+#[derive(Debug)]
+pub struct CannotMove;
 
 #[derive(Debug)]
 pub struct AnalyzerError {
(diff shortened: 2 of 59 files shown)


analyzer/src/constants.rs
```diff
@@ -0,0 +1 @@
+pub const MAX_INDEXED_EVENT_FIELDS: usize = 3;
```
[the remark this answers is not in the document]
cburgdorf: I would bet that it is an EVM restriction but my Google fu is not strong enough to yield a definitive answer. In any case, this PR doesn't change anything about it, just converting the error messages.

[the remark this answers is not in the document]
g-r-a-n-t: Indexed fields are passed as arguments to `log1`, `log2`, or `log3`. There is no `log4`.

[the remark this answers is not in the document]
satyamakgec: That makes sense, Thanks @g-r-a-n-t

cburgdorf:
@sbillig This can be reviewed. It's not yet complete but it has grown vast enough to not grow it further. I'll continue with the remaining refactoring in a new PR.

analyzer/src/traversal/expressions.rs
```diff
@@ -922,10 +1035,16 @@ fn expr_call_value_attribute(
...
         validate_arg_count(context, &attr.kind, attr.span, args, 0);
 
-        return match ValueMethod::from_str(&attr.kind)
-            .map_err(|_| SemanticError::undefined_value())?
-        {
-            ValueMethod::Clone => {
+        return match ValueMethod::from_str(&attr.kind) {
+            Err(_) => {
+                context.fancy_error(
+                    format!("Invalid call to `{}`", &attr.kind),
```
sbillig, line 1041: "no function `{}` exists on type `{}`" like below?

analyzer/src/traversal/expressions.rs
```diff
@@ -1228,14 +1380,20 @@ fn expr_name_call_type(
...
+    context: &mut Context,
     exp: &Node<fe::Expr>,
 ) -> Result<CallType, SemanticError> {
     if let fe::Expr::Attribute { value, attr } = &exp.kind {
         if let fe::Expr::Name(name) = &value.kind {
             match Object::from_str(&name) {
                 Ok(Object::Block) | Ok(Object::Chain) | Ok(Object::Msg) | Ok(Object::Tx) => {
-                    return Err(SemanticError::undefined_value())
+                    context.fancy_error(
+                        "illegal call on builtin object",
```
sbillig, line 1391: "no function `{}` exists on `{}`"?

analyzer/src/traversal/expressions.rs
```diff
@@ -893,7 +988,12 @@ fn expr_call_self_attribute(
             return_location,
         ))
     } else {
-        Err(SemanticError::undefined_value())
+        context.fancy_error(
+            "Call to undefined function",
```
sbillig, line 992: "no function `{}` exists on contract `{}`"?

analyzer/src/errors.rs
```diff
@@ -13,18 +20,9 @@ pub struct AnalyzerError {
...
-    NotCallable,
     NotSubscriptable,
-    NumericCapacityMismatch,
-    NumericLiteralExpected,
     SignedExponentNotAllowed,
-    StringCapacityMismatch,
     TypeError,
-    UndefinedValue,
     Fatal,
 }
```
sbillig, line 27: :metal:

sbillig (approved):
Epic effort, thanks for chugging through all of these. I left some nitpicky comments; I figure we'll probably do another pass through all the error messages to make the style and verbiage consistent. Up to you whether you feel like making the suggested changes.

analyzer/src/errors.rs
sbillig: Feels good to see all these error kinds deleted.

cburgdorf:
>I figure we'll probably do another pass through all the error messages to make the style and verbiage consistent. 

Yep, refactor fatigue is real..you start out to build nice, consistent, context rich error messages and somewhere down the road you just want to get through it and become bored and sloppy haha.

>Up to you whether you feel like making the suggested changes.

I made the suggested changes. But there's still a lot room to tune and enrich them in the future.

PR #437, opened 2021-06-01
Ensure that the compiler only panics for legitimate bugs

### LINKED ISSUE #436: Remove user-facing `todo`s and `unimplemented`s
### What is wrong?

The compiler will crash during analysis of features that are not yet implemented. For this reason, we have been filtering out crashes that display the text "not implemented" or "not yet implemented" when fuzzing. This would be fine if we didn't also use these panic messages in later stages of compilation, but we are, and it's likely that we've been filtering out meaningful crashes.

### How can it be fixed

- replace all justifiable `todo`s and `unimplemented`s with graceful analyzer errors
- let agroce know that any panic is considered a bug

g-r-a-n-t:
### What was wrong?

closes #436

### How was it fixed?

Refactored front-end "not implemented" errors into analyzer errors and removed questionable variants.

### To-Do

- [x] Clean up commit history


analyzer/src/traversal/expressions.rs
```diff
@@ -983,7 +996,7 @@ fn expr_call_value_attribute(
             ValueMethod::AbiEncode => match &value_attributes.typ {
                 Type::Struct(struct_) => {
                     if value_attributes.final_location() != Location::Memory {
```
g-r-a-n-t, line 998: these should just be "not in memory errors"

analyzer/src/traversal/expressions.rs
```diff
@@ -1007,9 +1020,21 @@ fn expr_call_value_attribute(
                         Location::Memory,
                     ))
                 }
-                _ => todo!(),
+                _ => {
+                    context.not_yet_implemented("fix this error", func.span);
```
g-r-a-n-t, line 1024: come up with a good message

analyzer/src/traversal/expressions.rs
```diff
@@ -1007,9 +1020,21 @@ fn expr_call_value_attribute(
...
+                    Ok(ExpressionAttributes::new(
+                        Type::unit(),
+                        Location::Value,
+                    ))
+                },
+            },
+            ValueMethod::AbiEncodePacked => {
```
g-r-a-n-t, line 1031: delete this variant

analyzer/src/traversal/expressions.rs
```diff
@@ -556,7 +556,20 @@ fn expr_unary_operation(
...
             }
-            _ => todo!(),
+            UnaryOperator::Invert => {
+                context.not_yet_implemented("unary invert", exp.span);
+                Ok(ExpressionAttributes::new(
+                    Type::unit(),
+                    Location::Value,
+                ))
+            },
+            UnaryOperator::UAdd => {
```
g-r-a-n-t, line 566: what is this supposed to do?

analyzer/src/traversal/contracts.rs
```diff
@@ -126,10 +126,11 @@ fn event_def(
         }
 
         // check if they are trying to index an array type
+        // todo clean all this up
         for index in indexed_fields.clone() {
             match fields[index].1.to_owned() {
                 FixedSize::Base(_) => {}
-                _ => unimplemented!("non-base type indexed event params"),
+                _ => context.not_yet_implemented("non-base type indexed event parameters", stmt.span),
```
g-r-a-n-t, line 133: fields

analyzer/src/traversal/structs.rs
```diff
@@ -20,7 +20,7 @@ pub fn struct_def(
         if let Type::Base(base_typ) = field_type {
             val.add_field(&name.kind, &FixedSize::Base(base_typ))?;
         } else {
-            todo!("Non-Base type fields aren't yet supported")
+            context.not_yet_implemented("non-base type struct parameters", field.span)
```
g-r-a-n-t, line 23: fields

analyzer/src/traversal/expressions.rs
g-r-a-n-t: as far as I can tell, this is sometimes used to convert things to numbers.

is there another use case that Im missing? if not, I think we should just ditch it.

analyzer/src/traversal/expressions.rs
```diff
@@ -983,7 +990,14 @@ fn expr_call_value_attribute(
...
                     if value_attributes.final_location() != Location::Memory {
-                        todo!("encode structs from storage")
+                        context.fancy_error(
+                            "value must be copied to memory",
+                            vec![Label::primary(value.span, "this value is in storage")],
+                            vec!["Hint: values located in storage can be copied to memory using the `to_mem` function.".into(),
+                                 "Example: `self.my_array.to_mem().abi_encode()`".into(),
+                            ],
+                        );
+                        // value_attributes.move_location = Some(Location::Memory);
```
g-r-a-n-t, line 1000: fix this

analyzer/src/traversal/module.rs
```diff
@@ -28,8 +28,12 @@ pub fn module(context: &mut Context, module: &fe::Module) -> Result<(), Semantic
...
                 contracts.push((stmt, contract_scope))
             }
-            fe::ModuleStmt::FromImport { .. } => unimplemented!(),
-            fe::ModuleStmt::SimpleImport { .. } => unimplemented!(),
+            fe::ModuleStmt::FromImport { .. } => {
+                context.not_yet_implemented("from import", stmt.span)
+            }
+            fe::ModuleStmt::SimpleImport { .. } => {
+                context.not_yet_implemented("simple import", stmt.span)
+            }
```
sbillig, line 36: We could probably cut these from the ast too (though I have no strong feelings about it)

sbillig (approved):

satyamakgec (approved):

analyzer/src/traversal/module.rs
g-r-a-n-t: Decided I'd leave it in there, since we may still use this syntax for imports.

PR #440, opened 2021-06-07
Runtime decoding checks

g-r-a-n-t:
We  now check the following when decoding data:

- The size of the encoded data fits within the size range known at compile-time.
- Values are correctly padded.
  - unsigned integers, addresses, and bools are checked to have correct left zero padding
  - the sizes of signed integers are checked
  - bytes and strings are checked to have correct right zero padding
- Data section offsets are consistent with the size of preceding values in the data section.
- The dynamic size of strings does not exceed their maximum size.
- The dynamic size of byte arrays (`u8[n]`) is equal to the size of the array.

### Decoding function walk through:

https://github.com/ethereum/fe/blob/169e95fe75fc8d7d39a120a6811ca2615daafcfa/crates/yulgen/tests/snapshots/yulgen__abi_decode_data_u256_bytes_string_bool_address_bytes_calldata_function.snap#L6-L31

*Note: At the moment, we have separate `decode_data_...` and `decode_component_tuple_...` functions. In the future when we add support for recursive encoding, we will just be using a `decode_tuple_...` function.*

**encoding size check**

```
     let encoding_size := sub(data_end, head_start) 
     if or(lt(encoding_size, 544), gt(encoding_size, 608)) { revert(0, 0) } 
```

We check that the size of the entire encoding fits within known bounds.

**head offset section**

```
     let head_offset_0 := 0 
     let head_offset_1 := 32 
     let head_offset_2 := 64 
     let head_offset_3 := 96 
     let head_offset_4 := 128 
     let head_offset_5 := 160 
```

Head offsets are known at compile-time, so we just assign a literal value.

**component decoding section**

```
     let decoded_val_0 := abi_decode_component_uint256_calldata(head_start, head_offset_0) 
     let decoded_val_1, data_start_offset_1, data_end_offset_1 := abi_decode_component_bytes_100_calldata(head_start, head_offset_1) 
     let decoded_val_2, data_start_offset_2, data_end_offset_2 := abi_decode_component_string_42_calldata(head_start, head_offset_2) 
     let decoded_val_3 := abi_decode_component_bool_calldata(head_start, head_offset_3) 
     let decoded_val_4 := abi_decode_component_address_calldata(head_start, head_offset_4) 
     let decoded_val_5, data_start_offset_5, data_end_offset_5 := abi_decode_component_bytes_100_calldata(head_start, head_offset_5) 
```

We call the decoding function for each component value. The decoding functions return the decoded value along with data section offsets for dynamically sized values.

**data offset checks section**

```
     if iszero(eq(data_start_offset_1, 192)) { revert(0, 0) } 
     if iszero(eq(data_start_offset_2, data_end_offset_1)) { revert(0, 0) } 
     if iszero(eq(data_start_offset_5, data_end_offset_2)) { revert(0, 0) } 
     if iszero(eq(encoding_size, data_end_offset_5)) { revert(0, 0) } 
```

We check that the offsets given in the last step are consistent with one another. Basically we're checking that each encoding starts where the last one ends. For the first offset, we check that it starts where the head section ends (`192` in this case).

**return section**

```
     return_val_0 := decoded_val_0 
     return_val_1 := decoded_val_1 
     return_val_2 := decoded_val_2 
     return_val_3 := decoded_val_3 
     return_val_4 := decoded_val_4 
     return_val_5 := decoded_val_5 
```

### To-Do

- [x] Clean up commit history


crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -133,47 +146,55 @@ pub fn decode_component<T: AbiEncoding>(typ: &T, location: AbiDecodeLocation) ->
...
             (return_val := [decode_expr])
+            [check_padding]
          }
     }
 }
 
 pub fn decode_component_int(size: usize, location: AbiDecodeLocation) -> yul::Statement {
     let func_name = abi_names::decode_component_int(size, location);
     let decode_expr = load_word(expression! { ptr }, location);
+    // let check_padding = abi_operations::check_left_padding((32 - size) * 8, expression! { return_val });
```
g-r-a-n-t, line 163: @cburgdorf I think our signed integers might have some problems..

This check failed, since all signed integers are encoded as if they are an `i256`, which means, for example, that an `i8` will use the leftmost bit in the word for signedness, as opposed to the *8th* bit from the right. Our decoding functions should account for this by moving the signdness bit to the appropriate spot and validating that all of the other bits are zero. They aren't, though, and the existing tests seem to work fine, which leads me to believe that all signed integers are effectively being treated like `i256`s. If this is the case, certain data operations will erase signs. For example, storing a negative `i8` value in an array and then reading it will yield a different (positive) number, since the sign was erased.

g-r-a-n-t: > Our decoding functions should account for this by moving the signdness bit to the appropriate spot

Well, maybe not actually. It might be better to treat all signed integers like `i256` values when they're on the stack, and only move the sign when reading and writing to memory/storage. This way we can still make use of the existing signed math functions.

cburgdorf: >since all signed integers are encoded as if they are an `i256`

Ah, that's right.

>Well, maybe not actually. It might be better to treat all signed integers like i256 values when they're on the stack, and only move the sign when reading and writing to memory/storage

Yep, that makes sense to me :+1: If you aren't going to address this in this PR we should create an issue and flag it with `mvp-required`

g-r-a-n-t: I'll write an issue

crates/analyzer/src/namespace/types.rs
```diff
@@ -70,6 +70,29 @@ impl AbiType {
             _ => todo!("recursive encoding"),
         }
     }
+
+    pub fn padding_size_bits(&self) -> usize {
```
g-r-a-n-t, line 74: docs

g-r-a-n-t, line 74: docs

crates/tests/src/features.rs
```diff
@@ -1278,3 +1278,235 @@ fn tuple_destructuring() {
...
+            harness.test_call_reverts(&mut executor, tampered_data);
+
+            // place non-zero byte in padded region of the tuple
+            let mut tampered_data = data.clone();
+            // 5 bytes past the end of address
+            let byte_index = 4 + 32 + 32 + 32 + 5;
+            tampered_data[byte_index] = 26;
+            harness.test_call_reverts(&mut executor, tampered_data);
+
+            // TODO: add tests for i16s
```
g-r-a-n-t, line 1398: add integer test

crates/yulgen/src/operations/abi.rs
```diff
@@ -4,6 +4,14 @@ use crate::utils::ceil_32;
 use fe_analyzer::namespace::types::{AbiDecodeLocation, AbiEncoding, AbiType};
 use yultsur::*;
 
+pub enum EncodingSize {
```
g-r-a-n-t, line 7: docs

crates/yulgen/src/operations/abi.rs
```diff
@@ -37,10 +45,41 @@ pub fn encode_size<T: AbiEncoding>(types: &[T], vals: Vec<yul::Expression>) -> y
 }
 
 /// Returns an expression that gives the size of the encoding head.
-pub fn encode_head_size<T: AbiEncoding>(types: &[T]) -> yul::Expression {
+pub fn encoding_head_size<T: AbiEncoding>(types: &[T]) -> yul::Expression {
     literal_expression! { (types.iter().map(|typ| typ.abi_type().head_size()).sum::<usize>()) }
 }
 
+pub fn encoding_known_size<T: AbiEncoding>(types: &[T]) -> EncodingSize {
```
g-r-a-n-t, line 52: docs

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -60,177 +76,293 @@ pub fn decode_data<T: AbiEncoding>(types: &[T], location: AbiDecodeLocation) ->
 
     // Create corresponding "val_" and "offset_" names and associate each one
     // with a component type.
-    let typed_names: Vec<(&T, String, String)> = types
+    let typed_names: Vec<(AbiType, String, String, String, Option<(String, String)>)> = types
```
g-r-a-n-t, line 79: look over this

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -60,177 +76,293 @@ pub fn decode_data<T: AbiEncoding>(types: &[T], location: AbiDecodeLocation) ->
...
-        AbiType::Bytes { size } => decode_component_bytes(size, location),
+        AbiType::String { max_size } => decode_component_string(*max_size, location),
+        AbiType::Bytes { size } => decode_component_bytes(*size, location),
     }
 }
 
 pub fn decode_component_uint(size: usize, location: AbiDecodeLocation) -> yul::Statement {
     let func_name = abi_names::decode_component_uint(size, location);
     let decode_expr = load_word(expression! { ptr }, location);
+    let check_padding = abi_operations::check_left_padding(
```
g-r-a-n-t, line 225: use padded size.. or just change method to take the value size

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -60,177 +76,293 @@ pub fn decode_data<T: AbiEncoding>(types: &[T], location: AbiDecodeLocation) ->
...
+        expression! { head_start },
+        expression! { inner_offset },
+        location,
+    );
 
     function_definition! {
          function [func_name](head_start, offset) -> return_val {
             (let ptr := add(head_start, offset))
-            (return_val := [decode_expr])
+            (return_val := avail())
```
g-r-a-n-t, line 301: delete pack method

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -60,177 +76,293 @@ pub fn decode_data<T: AbiEncoding>(types: &[T], location: AbiDecodeLocation) ->
...
+            (let end_word := [load_word(expression! { sub((add(head_start, data_end_offset)), 32) }, location)])
+            (let padding_size_bits := mul((sub(padded_data_size, data_size)), 8))
+            [abi_operations::check_right_padding(
+                expression! { padding_size_bits },
+                expression! { end_word }
+            )]
             (return_val := [copy_data(
-                expression! { data_start },
-                expression! { data_size },
+                expression! { add(data_start, 32) },
```
g-r-a-n-t, line 364: comment

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -255,6 +398,26 @@ pub fn decode_component_string(max_size: usize, location: AbiDecodeLocation) ->
     }
 }
 
+pub fn is_left_padded() -> yul::Statement {
+    function_definition! {
+        function is_left_padded(size, val) -> return_val {
```
g-r-a-n-t, line 403: docs

crates/yulgen/src/runtime/functions/abi.rs
```diff
@@ -300,7 +463,7 @@ pub fn pack(location: AbiDecodeLocation) -> yul::Statement {
     };
 
     function_definition! {
-        function [name](mptr, array_size, inner_data_size) -> packed_ptr {
```
g-r-a-n-t, line 303: again, delete this

crates/yulgen/tests/snapshots/yulgen__abi_pack_calldata_function.snap
```diff
@@ -3,7 +3,7 @@ source: crates/yulgen/tests/yulgen.rs
 expression: "abi_functions::pack(AbiDecodeLocation::Calldata)"
 
 ---
-function abi_pack_calldata(mptr, array_size, inner_data_size) -> packed_ptr {
+function abi_pack_calldata(mptr, array_size, inner_data_size, left_padding_size) -> packed_ptr {
```
g-r-a-n-t, line 6: delete this

crates/yulgen/tests/snapshots/yulgen__abi_pack_mem_function.snap
```diff
@@ -3,7 +3,7 @@ source: crates/yulgen/tests/yulgen.rs
 expression: "abi_functions::pack(AbiDecodeLocation::Memory)"
 
 ---
-function abi_pack_mem(mptr, array_size, inner_data_size) -> packed_ptr {
+function abi_pack_mem(mptr, array_size, inner_data_size, left_padding_size) -> packed_ptr {
```
g-r-a-n-t, line 6: delete this

crates/yulgen/tests/snapshots/yulgen__abi_decode_data_u256_bytes_string_bool_address_bytes_calldata_function.snap
```diff
@@ -0,0 +1,31 @@
...
+    let decoded_val_0 := abi_decode_component_uint256_calldata(head_start, head_offset_0)
+    let decoded_val_1, data_start_offset_1, data_end_offset_1 := abi_decode_component_bytes_100_calldata(head_start, head_offset_1)
+    let decoded_val_2, data_start_offset_2, data_end_offset_2 := abi_decode_component_string_42_calldata(head_start, head_offset_2)
+    let decoded_val_3 := abi_decode_component_bool_calldata(head_start, head_offset_3)
+    let decoded_val_4 := abi_decode_component_address_calldata(head_start, head_offset_4)
+    let decoded_val_5, data_start_offset_5, data_end_offset_5 := abi_decode_component_bytes_100_calldata(head_start, head_offset_5)
+    if iszero(eq(data_start_offset_1, 192)) { revert(0, 0) }
+    if iszero(eq(data_start_offset_2, data_end_offset_1)) { revert(0, 0) }
+    if iszero(eq(data_start_offset_5, data_end_offset_2)) { revert(0, 0) }
+    if iszero(eq(encoding_size, data_end_offset_5)) { revert(0, 0) }
```
g-r-a-n-t, line 24: @cburgdorf the Solidity code I looked at didn't revert with any specific values, but it looks like they plan on adding it. I'll create an issue to track this.

crates/yulgen/src/operations/abi.rs
```diff
@@ -4,6 +4,17 @@ use crate::utils::ceil_32;
...
+    Exact(yul::Expression),
+    Bounded {
+        min: yul::Expression,
+        max: yul::Expression,
+    },
+}
+
 /// Returns an expression that encodes the given values and returns a pointer to
 /// the encoding.
 pub fn encode<T: AbiEncoding>(types: &[T], vals: Vec<yul::Expression>) -> yul::Expression {
```
g-r-a-n-t, line 20: @sbillig I'll follow up this PR with another one tidying up these traits.

crates/tests/src/features.rs
```diff
@@ -1276,3 +1276,245 @@ fn tuple_destructuring() {
...
+                ),
+                tuple_token(&[address_token("a"), uint_token(42)]),
+                int_array_token(&(-10..16).collect::<Vec<_>>()),
+            ];
+            let data = harness.build_calldata("decode_u256_bytes_tuple_array", &input);
+
+            // add a byte
+            let mut tampered_data = data.clone();
+            tampered_data.push(42);
+            harness.test_call_reverts(&mut executor, tampered_data);
```
cburgdorf, line 1343: I think you said that solidity also just reverts with zero data for these checks but I wonder if that's just a todo on their end and if we should just introduce a panic code to be used for these reverts. From a developer perspective I think reverts with zero information are very annoying and a single panic code used for all calldata violation can already be quite helpful during debugging.

crates/yulgen/src/operations/abi.rs
```diff
@@ -88,3 +116,36 @@ pub fn unpack(
 ) -> yul::Statement {
     statement! { abi_unpack([ptr], [array_size], [inner_data_size]) }
 }
+
+/// Reverts if the value is not left padded with the given number of bits.
+pub fn check_left_padding(size_bits: yul::Expression, val: yul::Expression) -> yul::Statement {
+    statement! {
+        if (iszero((is_left_padded([size_bits], [val])))) {
+            (revert(0, 0))
```
cburgdorf, line 124: Just re-stating what I commented above on the test code. I think we should consider introducing our own panic code and then do a `revert_with_panic(<some-code>)`. Basically I think the only time where we should revert with zero data is when someone actually uses `revert` with no data. In all other circumstances we should revert with a panic code and we should go with whatever panic code Solidity uses and if they don't use one we make up our own.

cburgdorf (approved):
Looks good to me :+1: 
My only comment would be to think about introducing our own panic code for when we revert from these checks and use `revert_with_panic(..)`. I basically think we should never revert with zero data unless for when a user writes `revert` with no extra data. This should improve the debugging story for developers as tooling can map these panic codes to a user friendly error which might save a developer from having to manually step to a debugger to find out where the revert came from.

crates/tests/src/features.rs
g-r-a-n-t: I'm pretty sure it's a todo.

Any suggestions for the specific panic code?

g-r-a-n-t: Using the panic code `0x99` for now.

https://github.com/ethereum/fe/pull/440/commits/b9fbd323dd2ed5286248b3c278e99acc7c36e98e

crates/yulgen/tests/snapshots/yulgen__abi_decode_data_u256_bytes_string_bool_address_bytes_calldata_function.snap
cburgdorf: Oh I didn't notice until know that you had a comment on this. Anyway, `0x99` is good for now :+1:

PR #441, opened 2021-06-08
Migrate all remaining errors

cburgdorf:
### What was wrong?

Finalization of #432 (Migrating errors)

### How was it fixed?

The general approach taken was as follows:

1. If we have to return errors (because not enough access to report them directly) we create very narrow function specific errors
2. The only error that can be returned from the analyzer is `FatalError` which is an empty struct
3. There's a safety check in place that would panic in case the analyzer returns a `FatalError` but has not logged any errors
4. `SemanticError` and everything around it is gone :fire: 
5. The `AnaylzerError` was turned into a `pub struct AnalyzerError(pub Vec<Diagnostic>)` (since the `classic` property is gone)
6. Lots of new errors and test cases

### To-Do

There's still a lot to improve about the errors but at least this cleans up the migration work

### DIFF
--- a/analyzer/src/lib.rs
+++ b/analyzer/src/lib.rs
@@ -12,7 +12,7 @@ pub mod namespace;
 mod operations;
 mod traversal;
 
-use crate::errors::{AnalyzerError, ErrorKind};
+use crate::errors::{AnalyzerError, FatalError};
 use context::Context;
 use fe_common::files::SourceFileId;
 use fe_parser::ast as fe;
@@ -28,20 +28,15 @@ pub fn analyze(module: &fe::Module, file_id: SourceFileId) -> Result<Context, An
             if context.diagnostics.is_empty() {
                 Ok(context)
             } else {
-                Err(AnalyzerError {
-                    diagnostics: context.diagnostics,
-                    classic: None,
-                })
+                Err(AnalyzerError(context.diagnostics))
             }
         }
-        Err(err) => Err(AnalyzerError {
-            diagnostics: context.diagnostics,
-            classic: if err.kind == ErrorKind::Fatal {
-                None
-            } else {
-                Some(err)
-            },
-        }),
+        Err(FatalError) => {
+            if context.diagnostics.is_empty() {
+                panic!("Expected at least one error")
+            }
+            Err(AnalyzerError(context.diagnostics))
+        }
     }
 }
 
--- a/analyzer/src/traversal/expressions.rs
+++ b/analyzer/src/traversal/expressions.rs
@@ -1,15 +1,14 @@
-use crate::builtins;
 use crate::context::{CallType, Context, ExpressionAttributes, Location};
-use crate::errors::SemanticError;
+use crate::errors::{FatalError, IndexingError, TypeError};
 use crate::namespace::scopes::{BlockScope, ContractFunctionDef, Shared};
 use crate::namespace::types::{
     Array, Base, Contract, FeString, FixedSize, Integer, Struct, Tuple, Type, TypeDowncast, U256,
 };
 use crate::operations;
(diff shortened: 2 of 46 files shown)


analyzer/src/lib.rs
```diff
@@ -28,20 +28,15 @@ pub fn analyze(module: &fe::Module, file_id: SourceFileId) -> Result<Context, An
...
-            diagnostics: context.diagnostics,
-            classic: if err.kind == ErrorKind::Fatal {
-                None
-            } else {
-                Some(err)
-            },
-        }),
+        Err(FatalError) => {
+            if context.diagnostics.is_empty() {
+                panic!("Expected at least one error")
```
cburgdorf, line 36: This is a safety catch for us. If we returned a `FatalError` but the `diagnostics` are empty we've obviously forgot to report an error somewhere.

analyzer/src/traversal/expressions.rs
```diff
@@ -693,21 +746,36 @@ fn expr_call_builtin_function(
...
-                return Err(SemanticError::type_error());
+                context.fancy_error(
+                    format!(
+                        "`{}` can not be used as a parameter to `keccak(..)`",
+                        arg_typ
+                    ),
+                    vec![Label::primary(args.span, "wrong type")],
+                    vec!["Note: keccak(..) expects a byte array as parameter".into()],
+                );
+                return Err(FatalError);
```
sbillig, line 778: This doesn't need to be fatal, because we know the return type. The labeled span should be the first arg.

analyzer/src/traversal/expressions.rs
```diff
@@ -931,23 +999,30 @@ fn expr_call_type_constructor(
...
                     | Type::Base(Base::Numeric(_))
                     | Type::Base(Base::Address) => {}
-                    _ => return Err(SemanticError::type_error()),
+                    _ => {
+                        context.fancy_error(
+                            format!("`{}` can not be used as a parameter to `address(..)`", arg_attr.typ),
+                            vec![Label::primary(arg.span, "wrong type")],
+                            vec!["Note: address(..) expects a parameter of a contract type, numeric or address".into()],
+                        );
+                        return Err(FatalError);
```
sbillig, line 1008: Doesn't need to be fatal.

analyzer/src/traversal/expressions.rs
```diff
@@ -1207,7 +1282,12 @@ fn expr_call_type_attribute(
                     Location::Value,
                 ))
             } else {
-                Err(SemanticError::type_error())
+                context.fancy_error(
+                    "function `create2` expects numeric parameters",
+                    vec![Label::primary(args.span, "invalid argument")],
+                    vec![],
+                );
+                Err(FatalError)
```
sbillig, line 1290: Doesn't need to be fatal.

analyzer/src/traversal/expressions.rs
```diff
@@ -1229,7 +1309,12 @@ fn expr_call_type_attribute(
                     Location::Value,
                 ))
             } else {
-                Err(SemanticError::type_error())
+                context.fancy_error(
+                    "function `create` expects numeric parameter",
+                    vec![Label::primary(args.span, "invalid argument")],
+                    vec![],
+                );
+                Err(FatalError)
```
sbillig, line 1317: Doesn't need to be fatal.

sbillig (approved):
Thanks again for enduring the refactor slog. Looks great. There are a few cases that needn't be fatal (because we can return the expected output type, even if the inputs are bad), but I think I'll just merge this and we can tidy up such things in a future pass.

cburgdorf:
>There are a few cases that needn't be fatal (because we can return the expected output type, even if the inputs are bad)

Are you're right. I'll send a cleanup PR tomorrow.

PR #446, opened 2021-06-11
Fix broken CI build

cburgdorf:
### What was wrong?

CI is currently broken.

### How was it fixed?

Turns out our system dependencies install script was missing an `apt-get update` which started causing problems with the boost install not getting carried out as entirely causing the build to fail). Adding `apt-get update` was all it needed to fix it.


cburgdorf:
I will merge this without review as soon as it goes green since it is so minor and fixes the build

Issue #439, opened 2021-06-06
Make --optimize the default


cburgdorf:
### What is wrong?

Optimization is currently opt-in via `--optimize`. Unoptimized Fe code is virtually impractical to deploy and use so we should make that the default.

### How can it be fixed

- Decide on how to handle the CLI args (sorry, this is just a quick weekend brain dump)
- :hocho:

mjobuda:
sent pull request

cburgdorf:
This landed

PR #449, opened 2021-06-12
Make --optimize the default #439

mjobuda:
--optimize is now an option,--optimize=true is default

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -163,7 +163,7 @@ pub fn main() {
     let input_file = matches.value_of("input").unwrap();
     let output_dir = matches.value_of("output-dir").unwrap();
     let overwrite = matches.is_present("overwrite");
-    let optimize = matches.value_of("optimize")==Some("true");
+    let optimize = matches.value_of("optimize") == Some("true");
     let targets =
         values_t!(matches.values_of("emit"), CompilationTarget).unwrap_or_else(|e| e.exit());
     let with_bytecode = targets.contains(&CompilationTarget::Bytecode);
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -152,14 +152,18 @@ pub fn main() {
...
+                .use_delimiter(false)
+                .takes_value(true),
         )
         .get_matches();
 
     let input_file = matches.value_of("input").unwrap();
     let output_dir = matches.value_of("output-dir").unwrap();
     let overwrite = matches.is_present("overwrite");
-    let optimize = matches.is_present("optimize");
+    let optimize = matches.value_of("optimize")==Some("true");
```
g-r-a-n-t, line 166: The formatter would like spaces around `==`.

g-r-a-n-t (approved):
Looks great. Thanks for sending this. You can run `cargo fmt` to fix the formatting or do it manually.

src/main.rs
mjobuda: Thanks! The formatter is right. It should be with spaces. I corrected it. And updated the docs at https://github.com/mjobuda/fe/blob/master/docs/src/quickstart/installation.md
Should I do another pull request for the updated doc or how is the procedure?

newsfragments/439.feature.rst
```diff
@@ -0,0 +1,2 @@
+Enable the optimizer by default. The optimizer can still be disabled
+by supplying `--optimize=false` as an argument.
```
cburgdorf, line 2: @mjobuda I added this file to your PR. It ensures your change will be mentioned in the upcoming release notes. I will wait for CI to finish and then merge.

PR #454, opened 2021-06-15
Remove some fatal errors

cburgdorf:
### What was wrong?

There were a few places where we returned a `FatalError` even though the return type was known. Doing so prevents us from collecting more errors without the need to do so.

### How was it fixed?

Removed the `FatalError`s


sbillig (approved):
👍

PR #455, opened 2021-06-15
Make duplicate function definitons a fatal error

cburgdorf:
### What was wrong?

#451 demonstrates an interesting issue where we crash when trying to present a certain error. The reason for that is that the function is defined multiple times and when we retrieve the type information for another error we don't get the type for that function `bar` that we assume but for that other `function` bar that has a different type.

### How was it fixed?

An easy way to fix this is to say that duplicate definitions are treated as fatal errors with the reasoning that it becomes unsafe to retrieve type information for definitions of which we have multiple for the same name.

### DIFF
--- a/tests/src/snapshots/fe_compiler_tests__compile_errors__issue_451.snap
+++ b/tests/src/snapshots/fe_compiler_tests__compile_errors__issue_451.snap
@@ -4,11 +4,11 @@ expression: "error_string(&path, &src)"
 
 ---
 error: a function with the same name already exists
-  ┌─ fixtures/compile_errors/duplicate_method_in_contract.fe:6:5
+  ┌─ fixtures/compile_errors/issue_451.fe:4:5
   │  
-6 │ ╭     pub def bar():
-7 │ │         pass
-  │ ╰────────────^ Conflicting definition of function `bar`
+4 │ ╭     pub def bar:
+5 │ │      s
+  │ ╰──────^ Conflicting definition of function `bar`
   │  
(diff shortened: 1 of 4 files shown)


tests/src/snapshots/fe_compiler_tests__compile_errors__issue_451.snap
```diff
@@ -0,0 +1,15 @@
...
+source: tests/src/compile_errors.rs
+expression: "error_string(&path, &src)"
+
+---
+error: a function with the same name already exists
+  ┌─ fixtures/compile_errors/issue_451.fe:4:5
+  │  
+4 │ ╭     pub def bar:
+5 │ │      s
+  │ ╰──────^ Conflicting definition of contract `bar`
```
sbillig, line 11: I guess "contract" is a copy-paste issue neither of us noticed last time. Probably worth fixing now.

sbillig (approved):
Looks good. I have some ideas about further reducing error fatalities that I might implement when the salsa stuff is in. Basically in this case the 'bar' function would be marked as "tainted", and only the use of 'bar' would be fatal. When (or if??) the analyzer is query-based, error fatality won't stop all analysis, just that of the current definition.

tests/src/snapshots/fe_compiler_tests__compile_errors__issue_451.snap
cburgdorf: Oops, right :+1:

PR #458, opened 2021-06-19
Move ast::ModuleStmt and ContractStmt variants into separate structs

sbillig:
Splitting my salsa refactorfest into some smaller PRs.

### What was wrong?

It's nice to refer to `StructDef`, `FuncDef`, etc directly instead of always matching on `ModuleStmt` or `ContractStmt` enums. This cleans up some of the ugly
```
if let ModuleStmt::StructDef { .. } = &stmt.kind {
  ... 
} else {
  unreachable!() 
}
```
blocks, and makes the salsa stuff cleaner.

I also renamed ast::FuncDef to Function, StructDef to Struct, etc, because it looks nicer to me.

### How was it fixed?
:raised_hands:

### DIFF
--- a/analyzer/src/context.rs
+++ b/analyzer/src/context.rs
@@ -343,7 +343,7 @@ impl Context {
     /// # Panics
     ///
     /// Panics if an entry already exists for the node id.
-    pub fn add_function(&mut self, node: &Node<fe::FuncDef>, attributes: FunctionAttributes) {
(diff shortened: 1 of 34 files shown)


cburgdorf (approved):
I like this. It's a surprisingly simple mitigation of the many `unreachable!()`s that we previously had to use :+1:

PR #459, opened 2021-06-20
Move tuple and list expr collection to lowering pass; support lowering of nested tuples

sbillig:
### What was wrong?

Another salsa-inspired change. The analyzer collects a bunch of stuff for the later passes, but those passes can collect the stuff just as easily.

### How was it fixed?

Added a couple context structs to the lowering pass, and removed tuples_used and list_expressions stuff from the analyzer.

While I was mucking about, I added lowering support for nested tuples, because I can't stop myself from sneaking unrelated stuff into a PR. Generated tuple struct names now include a trailing underscore to differentiate between different tuple nestings. I initially made this a special case for nested tuples, but the special casing didn't seem as aesthetically pleasing to me. Conceptually, the generated tuple struct name replaces each paren and comma with an underscore, and each item type name with its lowered type name.
```
((A, B), C, D) // tuple_tuple_A_B__C_D_
((A, B, C), D) // tuple_tuple_A_B_C__D_
((A, B, C, D),) // tuple_tuple_A_B_C_D__
```

Also tuples in type aliases weren't being lowered; eg `type T = (u8, u8)`; they are now. I called this a feature, because I guess tuple type aliases are new functionality.

### To-Do

- [x] a bit o' cleanup
- [x] rebase on master when #458 is merged
- [x] add type alias fix to release notes
- [x] add test for lowering of nested tuples

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,6 @@ on:
       - v*
 
   pull_request:
-    branches: [master]
 
 env:
   CARGO_TERM_COLOR: always
--- a/tests/fixtures/lowering/base_tuple_lowered.fe
+++ b/tests/fixtures/lowering/base_tuple_lowered.fe
@@ -1,26 +1,26 @@
-struct tuple_u256_bool:
+struct tuple_bool_address_:
+    item0: bool
+    item1: address
+
+struct tuple_bool_bool_:
+    item0: bool
+    item1: bool
+
+struct tuple_address_address_:
+    item0: address
+    item1: address
+
+struct tuple_u256_bool_:
     item0: u256
     item1: bool
 
-struct tuple_u256_bool_u8_address:
+struct tuple_u256_bool_u8_address_:
     item0: u256
     item1: bool
     item2: u8
     item3: address
 
-struct tuple_bool_bool:
-    item0: bool
-    item1: bool
-
-struct tuple_bool_address:
-    item0: bool
-    item1: address
-
-struct tuple_address_address:
-    item0: address
-    item1: address
-
-struct tuple_address_address_u16_i32_bool:
+struct tuple_address_address_u16_i32_bool_:
     item0: address
     item1: address
     item2: u16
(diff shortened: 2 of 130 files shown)


tests/fixtures/lowering/base_tuple_lowered.fe
```diff
@@ -1,52 +1,52 @@
-struct tuple_u256_bool:
+struct __Tuple0:
```
cburgdorf, line 1: Did you consider alterations to the current naming scheme that would preserve the readability better? E.g. we could encode the opening and closing parenthesis with some character like an extra `_`.


```
((A, B), C, D) // _tuple__tuple_A_B__C_D_
((A, B, C), D) // _tuple__tuple_A_B_C__D_
```

tests/src/lowering.rs
```diff
@@ -69,6 +73,10 @@ fn test_lowering(fixture: &str) {
         replace_spans(to_ron_string_pretty(&actual_lowered_ast).unwrap()),
     );
 
-    fe_analyzer::analyze(&actual_lowered_ast, src_id)
-        .expect("analysis of the lowered module failed");
+    // TODO: the analyzer rejects lowered nested tuples, because
+    //  nested structs aren't supported yet. we should move the
+    //  not-yet-implemented error to the compiler.
+    //
+    // fe_analyzer::analyze(&actual_lowered_ast, src_id)
```
cburgdorf, line 80: Probably worth creating an issue for it and putting it on the active milestone unless you have that work already in progress anyway.

cburgdorf (approved):
Looks good and makes sense. Left a few comments inline.

.github/workflows/main.yml
```diff
@@ -7,7 +7,6 @@ on:
       - v*
 
   pull_request:
-    branches: [master]
```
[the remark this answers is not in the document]
sbillig: Yeah; that's in a separate commit. It enables gh actions for all PRs, regardless of the target branch. Sometimes I open a PR that depends on another that's pending review/merge; so I push the branch of the first to the main repo, and have the second PR target that branch (temporarily) so that the changes of the first don't show in the second's diff. (This one depends on #458)

tests/fixtures/lowering/base_tuple_lowered.fe
sbillig: Yeah, I was thinking that that's *less* readable in the case of a couple nestings, but it's better in the simpler/common cases, so I'll switch it back.

.github/workflows/main.yml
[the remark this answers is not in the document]
cburgdorf: Ah, makes sense :+1:

tests/fixtures/lowering/base_tuple_lowered.fe
cburgdorf: yes, I guess having to differentiate between two easy to confuse tuples will be rather rare so maybe it's better to keep it readable for the common cases.

tests/src/lowering.rs
sbillig: Opened #461

tests/fixtures/lowering/base_tuple_lowered.fe
sbillig: It's now 
```
(A, B) => tuple_A_B_
(A, (B, C), D) => tuple_A_tuple_B_C__D_
``` 
ie, replace parens and commas with underscores, and prefix with "tuple"

cburgdorf: Looks good :+1:

PR #460, opened 2021-06-21
Upgrade some dependencies and re-enable test

cburgdorf:
### What was wrong?

Our `ethereum_types`, `primitive_types` and `evm_runtime` dependencies were out of date. We also had one test disabled that was waiting on these upgrades

### How was it fixed?

- Modifications on some testing helpers (done by Grant)
- Re-enabled arithmetic test that was failing with previous eth_runtime versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,10 +30,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.41"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


cburgdorf:
@g-r-a-n-t Thanks for helping out with the upgrade :pray: 
Unfortunately the ethabi upgrade still doesn't address the issue with the new custom errors that I ran into so I have to dig a bit deeper here. However, I re-enabled the test that was waiting on the upstream `evm_runtime` fix. So at least there's that :)

cburgdorf:
I created an upstream issue to support the new `error` type of the ABI.

https://github.com/rust-ethereum/ethabi/issues/225

g-r-a-n-t (approved):

Issue #450, opened 2021-06-12
failed to analyze lowered AST: invalid assignment target at compiler/src/lib.rs:55:59


agroce:
Compiling this file:

```
contract o:
 pub def d(a:u256,b:u256)->u256:
        a-a&=b
        my_array:u256[0]
        return my_array[7]
```

```
thread 'main' panicked at 'failed to analyze lowered AST: AnalyzerError([Diagnostic { severity: Error, code: None, message: "invalid assignment target", labels: [Label { style: Primary, file_id: SourceFileId(290645383916162376144464712506526417610), range: 53..56, message: "" }], notes: ["The left side of an assignment can be a variable name, attribute, subscript, or tuple."] }])', compiler/src/lib.rs:55:59
stack backtrace:
   0:     0x55b5e6fd8a00 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55b5e6fd8a00 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55b5e6fd8a00 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55b5e6fd8a00 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x55b5e6ff1d3f - core::fmt::write::h3b84512577ca38a8
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/fmt/mod.rs:1092:17
   5:     0x55b5e6fd6492 - std::io::Write::write_fmt::h465f8feea02e2aa1
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/io/mod.rs:1572:15
[...]

PR #462, opened 2021-06-21
failed to analyze lowered AST: invalid assignment target at compiler/src/lib.rs:55:59 #450

mjobuda:
### What was wrong?

https://github.com/ethereum/fe/issues/450

Compiling this file:
```
contract o:
 pub def d(a:u256,b:u256)->u256:
        a-a&=b
        my_array:u256[0]
        return my_array[7]
```
`thread 'main' panicked at 'failed to analyze lowered AST: AnalyzerError`

I analyzed this and the result is:
1) This code shouldn't compile. &= is not found to be an operator as to the spec. So this is an assignment. But the left side of the assignment doesn't make any sense. But noncompiling code shouldn't lead to a crash! 

2) Anatomy of the bug: The analyzer is first run on the unlowered code. This analyzer run does not lead to an error. Next the analyzer is run on the lowered code. This is implemented to panic on error. Which happens here. The difference on both runs is that in the unlowered analyzer run fe_analyzer::traversal::assignments::aug_assign is called, in the lowered run fe_analyzer::traversal::assignments::assign is run.
### How was it fixed?

assign calls check_assign_target(context, target)? and finds the error(the left side of the assignment is not a valid target, which leads to the error triggering the  panic), aug_assign does not.
3) workaround/fix: call check_assign_target(context, target)?; in aug_assign

### Todo/question 

4) comments:
a) is it really necessary to call panic if the lowered code has errors?
b) After removing the problematic line the following code compiles:
```
contract o:
 pub def d(a:u256,b:u256)->u256:
        my_array:u256[0]
        return my_array[7]
```
Shouldn't this give something like ArrayIndex out of bound error?


sbillig:
@mjobuda Thanks for looking into this and fixing it.

> a) is it really necessary to call panic if the lowered code has errors?

The idea here is that a failure to analyze the lowered code is definitely indicative of a bug (in the Fe implementation), and panicking makes it easy for agroce's fuzzing effort to help us identify these weaknesses. This is still a pretty young project; as it matures and we feel it's ready for actual use, we'll probably convert these panics into something more user friendly.

>  Shouldn't this give something like ArrayIndex out of bound error?

Yeah, totally, we just haven't added a check for that to the analyzer yet. The analyzer is mostly just doing type checking so far, but we certainly plan to expand its powers over time. If this is something you're interested in working on, that would be awesome. Otherwise, it's definitely something we'll handle as the analyzer becomes more sophisticated.

sbillig (approved):

sbillig:
I opened #463 for the array out-of-bounds thing.

PR #464, opened 2021-06-22
Support revert with custom errors (EIP 838)

cburgdorf:
### What was wrong?

We don't support custom errors as specified by [EIP-838](https://github.com/ethereum/EIPs/issues/838) ([solidity docs have more info](https://docs.soliditylang.org/en/v0.8.4/abi-spec.html#errors))

### How was it fixed?

1. Added support for `revert <expression>` where `<expression>` has to be a `struct`. We might change that in the future to allow more flexibility with reverts. For now structs are the only way to return an EIP-838 compatible error.

E.g.

```
struct Error:
    msg: u256
    val: bool
    
pub def revert_custom_error():
    revert Error(msg=100, val=true)
```

The revert data for the above would be:

```
  --sig--|-----------------------------100------------------------------|---------------------------------true----------------------------
0x3253e3c700000000000000000000000000000000000000000000000000000000000000640000000000000000000000000000000000000000000000000000000000000001
```

2. This PR doesn't change anything about `assert` statements but as a recap we currently encode `assert false, "foo"` as if it were `assert false, Error(msg="foo")`. We could use the lowering pass to do the conversion but since we do not yet support `String<N>` in structs this continues to be a special case. But again, this PR doesn't change the existing encoding of `assert` it only refactors the underlying code to be more flexible and shared across `revert` and `assert` statements.

3. I added a bunch of new solidity tests that prove that the error coding on their end works similar to ours

### DIFF
--- a/crates/parser/src/grammar/functions.rs
+++ b/crates/parser/src/grammar/functions.rs
@@ -186,7 +186,6 @@ pub fn parse_single_word_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
         TokenKind::Continue => FuncStmt::Continue,
         TokenKind::Break => FuncStmt::Break,
         TokenKind::Pass => FuncStmt::Pass,
-        TokenKind::Revert => FuncStmt::Revert,
         _ => panic!(),
     };
     Ok(Node::new(stmt, tok.span))
@@ -203,7 +202,8 @@ pub fn parse_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
         While => parse_while_stmt(par),
         Return => parse_return_stmt(par),
         Assert => parse_assert_stmt(par),
-        Continue | Break | Pass | Revert => parse_single_word_stmt(par),
+        Revert => parse_revert_stmt(par),
+        Continue | Break | Pass => parse_single_word_stmt(par),
         Emit => parse_emit_statement(par),
         _ => parse_expr_stmt(par),
     }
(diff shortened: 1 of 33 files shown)


crates/test-utils/src/lib.rs
```diff
@@ -242,26 +257,49 @@ pub fn deploy_solidity_contract(
...
+}
 
-    // Length of the string padded to 32 bit hex
-    let string_len = format!("{:0>64x}", reason.len());
+#[allow(dead_code)]
+pub fn encode_error(selector: &str, input: &[ethabi::Token]) -> Vec<u8> {
+    let mut data = String::new();
+    for param in input {
+        let encoded = match param {
+            // FIXME: The double format feels hacky but how do I do that properly?!
```
cburgdorf, line 268: Make Github comment instead

crates/parser/src/grammar/functions.rs
```diff
@@ -406,6 +406,21 @@ pub fn parse_assert_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
...
+///
+/// # Panics
+/// Panics if the next token isn't `revert`.
+pub fn parse_revert_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
+    let revert_tok = par.assert(TokenKind::Revert);
+    let error = match par.peek() {
+        None | Some(TokenKind::Newline) => None,
+        Some(_) => Some(parse_expr(par)?),
+    };
+    par.expect_newline("assert statement")?;
```
cburgdorf, line 419: revert statement

crates/parser/tests/cases/snapshots/cases__parse_ast__stmt_revert.snap
```diff
@@ -4,7 +4,9 @@ expression: "ast_string(stringify!(stmt_revert), functions::parse_stmt, \"revert
 
 ---
 Node(
```
cburgdorf, line 6: remove file

crates/yulgen/src/mappers/functions.rs
```diff
@@ -150,7 +171,24 @@ fn assert(context: &mut Context, stmt: &Node<fe::FuncStmt>) -> yul::Statement {
         return match msg {
             Some(val) => {
                 let msg = expressions::expr(context, val);
-                statement! { if (iszero([test])) { (revert_with_reason_string([msg])) } }
+                let msg_expr = context
```
cburgdorf, line 174: rename to `msg_attributes`

g-r-a-n-t (approved):
very nice :sunglasses:

PR #471, opened 2021-06-29
The great reorg

### LINKED ISSUE #358: Consider switching back to MIT license
### What is wrong?

I believe the only reason why we migrated away from the MIT license was because the parser incorporated code from a project (nom) that isn't MIT licensed. With the parser being rewritten we could consider switching back.

### How can it be fixed

Discuss & change.

sbillig:
### What was wrong?

- The root directory has a lot of stuff in it. The crates (except for fuzz) now live in the crates/ directory.
- The compiler crate was a collection of mostly independent stages. This can be a hassle for development. Eg while working on a big analyzer refactor, I'd like to make the lowering stage work with the new analyzer changes and run the lowering tests, without having to fix the yulgen stage at the same time. This is now split into lowering, yulgen, yulc, abi; the main `compile` function has been moved to the driver crate.
- We moved most of the tests to the tests crate. This makes it hard to test each stage independently. The shared "fixture" files are now in a separate test-files crate (eg `let src = test_files::fixture("demos/erc20.fe");`). These files are embedded in the crate at compile time via https://crates.io/crates/include_dir, so they can be used easily in wasm. I made more of the tests run on wasm, but some of them need more work (I'll open an issue).
- Stuff that depends on solidity wasn't licensed properly (see https://github.com/ethereum/fe/issues/358#issuecomment-832046040). I updated the licenses accordingly (Apache2 for the frontend, GPL3 for the rest), updated the readme, and added license files.
- I changed the Cargo.toml author fields to "The Fe Developers <snake...>"
- I removed the (slightly outdated) readme contents from the main.rs doc comment.

There may be a few very small functional changes, but I tried to just move things around. Most notably, abi json generation failure is now a panic. This felt more appropriate, and I didn't want to bother mapping error types.

I recommend looking through the commits, rather than using gh's "Files changed" view.

closes #358 

### How was it fixed?
:computer:

### To-Do

- [x] Clean up commit history


g-r-a-n-t (approved):
This structure looks much better. My only concern is that we're wiping git blame info, is there a reason `git mv` wasn't used?

sbillig:
> My only concern is that we're wiping git blame info, is there a reason `git mv` wasn't used?

We aren't, and it was (though git recognizes most code moves even when not using `git mv`, eg I `git rm`ed the snapshots and regenerated them by running the tests, and git picked that up as a rename with modifications just fine. Afaict there's effectively no difference between mv and add+rm).

Github's 'files changed' view doesn't seem to pick up on the moves, but the individual commit view does. git blame works fine, and eg `git log --follow crates/yulgen/src/lib.rs` shows the full history.

g-r-a-n-t:
I see. Well, I'm ok with merging this.

cburgdorf:
Nice! I noticed a tiny breakage with the website deployment for which I send a fix in https://github.com/ethereum/fe/pull/473

PR #473, opened 2021-06-30
Fix path issue with cargp pkgid command

cburgdorf:
### What was wrong?

A tiny breakage that came out of the big repository reorganization. The website needs to read the current Fe version but the `cargo pkgid` command was confused now that the layout has changed.

### How was it fixed?

Added a `cd crates/fe` to the command. I tried specifying the command via `cargo pkgid -p crates/fe` but it seems that command only takes absolute file paths so I went for the `cd` instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build-website:
 	cp -r website/ target/
 
 	# Fill version marker with current version
-	sed -i "s/{{FE_VERSION}}/$$(cargo pkgid | cut -d# -f2 | cut -d: -f2)/g" target/website/index.html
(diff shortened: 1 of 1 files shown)


sbillig (approved):
:+1: 

Could/did you look at `cargo release` too? I *think* it's fine, but one of you seasoned release cutters should verify. I forgot to mention in the other PR that it says
> Cannot detect changes for fe-parser because tag fe-parser-v0.6.1-alpha is missing. Try setting `--prev-tag-name <TAG>`

(for all crates), them seems to carry on updating everything successfully (using `--dry-run` to test). Adding `--prev-tag-name v0.6.1-alpha` gets rid of the warning and also seems fine.

cburgdorf:
@sbillig I'll take a look at the release command

PR #474, opened 2021-06-30
Fix cargo pkgid invocation in release command

cburgdorf:
Similar to #473


sbillig (approved):
Turns out that `cargo pkgid fe` works fine (as opposed to `cargo pkgid crates/fe`) when in the root dir. Whichever is fine by me though.

cburgdorf:
Ah you're right. How did I not try that :D Fixed both invocations.

PR #481, opened 2021-07-13
Update readme hyperlink

byterose:
### What was wrong?
The hyperlink in [README.md](https://github.com/ethereum/fe/blob/master/README.md) that directs the user to the fe contract examples is not updated.

### How was it fixed?
I replaced the old hyperlink with the updated hyperlink.


g-r-a-n-t:
Thanks for this :slightly_smiling_face:

PR #482, opened 2021-07-14
Update file because of updated Syntax

bookteo:
### What was wrong?
string100 instead of String<100>
error message was outdated as well

### How was it fixed?
changed to String<100>
updated error message as well

### To-Do

- [ ] Clean up commit history


cburgdorf (approved):

cburgdorf:
Thanks for the fix!

PR #484, opened 2021-07-16
Fixes ICE 469

cburgdorf:
### What was wrong?

1. The following does currently lead to an ICE (see #469 found by the fuzzer)

```
contract Foo:
  pub def bar:
    (x,):u256[20]=[1,2,3,4,5,6,7,8,9,0,2,1,3,4,5,6,7,8,9,0]
```

The reason here is that the type `u256[20]` is wrong since we are destructuring into a tuple

2. The following also leads to to an ICE

``` 
contract Foo:
  pub def bar():
    (x, y):(u256,)=(1,)

```

The reason here is that the number of items between `(x, y)` (2 items) and the declared type `(u256,)` (1 item) doesn't match.

### How was it fixed?

Just added nice user messages. The error was already caught but we didn't have proper user errors yet which caused the ICE.


sbillig (approved):
Looks good

Issue #161, opened 2020-12-12
zero-cost abstraction of pointer location


g-r-a-n-t:
The EVM has two places where reference types may be stored (storage and memory). Storage is used for values that need to persist indefinitely, while memory is used for values that only need to persist for the duration of contract execution. The differences between these two locations can be confusing, so it would be nice if their usage was  tucked behind an abstraction with little or no runtime cost.

In short, we might be able to accomplish this by focusing on two things:

- limitations on where state can be mutated
- explicit variable mutability

**When you can mutate state in Fe**

In Fe, it should be clear when a contract's state is being mutated. This can be done by ensuring that all state-mutating operations access a mutable instance of `self` directly.

```python
# a function that mutates state
def my_mut_func():
  self.my_array[0] = 42

...

# calls a function that mutates a contract field
self.my_mut_func()

# mutates a contract field
self.my_map[0x00] = 26
```

In order to make this guarantee, though, we must ensure that all mutable storage pointers are expressed as attributes on `self`, which means that someone should not be able to assign a storage pointer (i.e. `self.attribute`) to some function variable. Instead, they must copy it into memory and then assign it.

```python
# invalid code
my_sto_array = self.my_array # ERROR: cannot assign a storage pointer to a mutable variable
my_sto_array[26] = 10 # This line would have mutated state without starting with `self.`

# also invalid
self.my_func(self.my_array) # ERROR: cannot assign storage pointer to a mutable function parameter

# fix
my_array = mem_copy(self.my_array)
my_array[26] = 10 # This line does not mutate state.

my_func(mem_copy(self.my_array))
```

*side note: this copy currently happens implicitly*

```python
my_array = self.my_array
#              ^ copied to memory
my_array[26] = 10 # mutates memory, not storage

self.my_func(self.my_array)
#                ^ copied to memory and passed in

# which leads to some confusing semantics
set_idx_3_to_42(my_array)
set_idx_3_to_42(self.my_array)
assert my_array[3] == self.my_array[3] # reverts: self.my_array has not been mutated
```

So for now, one must copy a value from storage and into memory before they can assign it to another value. This is somewhat expensive, but it eliminates the challenge of dealing with ambiguous expression locations, since everything is assumed to be in memory.

**Assigning and passing immutable storage pointers**

Currently in Fe, we assume that all values are mutable, so we need to take extra precautions when compiling code. If we were to introduced explicit mutability, we could make better inferences about code and eliminate the need for copying to memory in some cases.

[...]

PR #489, opened 2021-07-27
Overhaul spec content

cburgdorf:
### What was wrong?

- The layout of the syntax definitons is broken e.g:

image

- lots of broken links
- the Python indentation rules were not properly reflected in the syntax sections
- lots of missing content

### How was it fixed?

Still an ongoing effort but:

1. The spec files where all split into individual files (similar to how the [Rust reference is organized](https://doc.rust-lang.org/nightly/reference/) which makes it easier to maintain the links of each section
2. I removed the TOC preprocessor that we used for the release notes since it was causing lots of layout issues with the syntax definitons
image
3. Changed the syntax sections to properly use NEWLINE, INDENT, DEDENT tokens to describe the syntax
4. Added a bunch of new content as well as stubs for things that we will need to cover (e.g. all the expressions and statements)


cburgdorf:
@g-r-a-n-t This is far from finished but I think it may be good to get the first chunk in before it grows to large.

docs/src/spec/struct_types.md
```diff
@@ -0,0 +1,37 @@
...
+
+  def do_something():
+    # A rectangle in memory
+    square: Rectangle = Rectangle(width=10, length=10)
+```
+
+All fields of struct types are always initialized.
+
+The memory layout of a `struct` is undefined by default to allow for compiler
+optimizations like field reordering.
```
g-r-a-n-t, line 30: maybe this is a better note:

> The data layout of a `struct` is not part of its external API and may be changed between any release.

There are a couple things that influence the layout of struct fields: space efficiency and word alignment. Maybe we don't discuss either of these things in here and instead leave it to the data layout section.

docs/src/spec/to_mem_function.md
```diff
@@ -0,0 +1,9 @@
+# The `to_mem` function
+
+Reference type values can be copied from storage and into memory using the `to_mem` function.
```
g-r-a-n-t, line 3: In the future, I think there may be somewhat nuanced usage of  `to_mem()` on values in memory.

Example:

```
pub fn foo(self):
  sum(self.my_array)
  sum([1, 2, 3]) # this is not copied in `sum`, as `to_mem` operates on it passively

fn sum(some_array: u8[3]):
  let sum: u256 = 0
  for val in some_array.to_mem():
    sum += val

  return sum
```

So I think we should remove the check mandating that the value be in storage (if it exists). This would of course be a separate issue.

docs/src/spec/tuple_types.md
```diff
@@ -0,0 +1,42 @@
...
+The second field is `item1`.
+And so on.
+The type of each field is the type of the same position in the tuple's list of types.
+
+For convenience and historical reasons, the tuple type with no fields (`()`) is often called *unit* or *the unit type*.
+Its one value is also called *unit* or *the unit value*.
+
+Some examples of tuple types:
+
+* `()` (unit)
```
g-r-a-n-t, line 28: suggestion:

> `()` (aka the unit or zero-sized type)

g-r-a-n-t:
Looks great. I left a bunch of comments, some of which are suggestions and others are just general thoughts that probably warrant new issues.

g-r-a-n-t:
Also, I'd be happy to rewrite a bunch of stuff in the data layout section after this gets merged.

docs/src/spec/contracts.md
```diff
@@ -0,0 +1,56 @@
...
+> &nbsp;&nbsp; )
+>
+> _Visibility_ :\
+> &nbsp;&nbsp; `pub`<sup>?</sup>
+>
+> _ContractField_ :\
+> &nbsp;&nbsp; [IDENTIFIER] `:` [_Type_]
+
+
+A _contract_ in Fe is a collection of code that resides at a specific address on the Ethereum blockchain. It is defined with the keyword `contract`.
```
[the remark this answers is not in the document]
cburgdorf: :+1:

docs/src/spec/keywords.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+## Reserved keywords
+
+These keywords aren't used yet, but they are reserved for future use. They have
+the same restrictions as strict keywords. The reasoning behind this is to make
+current programs forward compatible with future versions of Fe by forbidding
+them to use these keywords.
+
+> **Lexer:**\
```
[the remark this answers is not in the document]
cburgdorf: No, we but we should definitely do that. This PR doesn't change the keywords (only moves the text into a dedicated file) but I will opened up an issue to ensure we sync spec and implementation in regard to keywords / reserved words: https://github.com/ethereum/fe/issues/504

docs/src/spec/string_type.md
```diff
@@ -0,0 +1,20 @@
+# String Type
+
+A value of type `String<N>` represents a sequence of unsigned bytes with the assumption that the data is valid UTF-8.
+
+Fe restricts strings to a maximum size to improve decidability and to make it easier to deliver precice gas cost estimations.
```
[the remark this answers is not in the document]
cburgdorf: >I think that, ideally, controls on decidability may be elided (like lifetimes in rust), but are easily available to those who would like to make use of them

Yes, I would love to see us going that direction, too :+1:

docs/src/spec/to_mem_function.md
cburgdorf: But why would we even write `some_array.to_mem()` here? Isn't the assumption that `some_array` is in memory anyway? Isn't it that we would only need to call `to_mem()` on something that is `self.<something>`? Or is this hypothetical code in which `some_array` may allow *readonly* storage pointers?

Also we might want to reconsider if `to_mem` can be merged with `clone`.

g-r-a-n-t: > Isn't the assumption that some_array is in memory anyway?

This is the current assumption, as we require that reference type parameters be in memory.

> Or is this hypothetical code in which some_array may allow readonly storage pointers?

This is what has been described in #161. `to_mem` would serve as a function that would guarantee that a readonly value is in memory. If the value is already in memory, then we would leave it alone, as to not incur unnecessary gas costs. If the value is in storage, then we would copy it.

PR #493, opened 2021-07-30
Using `value_expr` and `assignable_expr` analysis functions where appropriate.

### LINKED ISSUE #491: Fix places where `assignable_expr` should be used.
### What is wrong?

There are a few places in the analyzer where we use `expressions::expr` to analyze expressions, when we should be using `expressions::assignable_expr`.

The name `assignable_expr` is not very clear, so we should consider changing that too.

The purpose of `assignable_expr` is to make sure expressions are in the default location of their type. The yulgen phase is then able to assume the location of a given value. If we don't do this, then we will have all sorts of undefined behavior.

### How can it be fixed

- Fix the following places and double check that we are using these functions correctly every where else.
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/expressions.rs#L77
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/expressions.rs#L89
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/expressions.rs#L108
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/functions.rs#L415
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/functions.rs#L426
  - https://github.com/ethereum/fe/blob/52df65bce06a4e8197987af737b32f13fea39f3c/crates/analyzer/src/traversal/functions.rs#L449
- Add tests for most or all features that use `assignable_expr` where we use values in both storage and memory.

g-r-a-n-t:
### What was wrong?

closes #491 

### How was it fixed?

Searched usages of `expression::expr` and fixed places where `value_expr` or `assignable_expr` needed to be used.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -74,19 +74,24 @@ pub fn expr_list(
 
     let inner_type = if let Some(expected) = expected_type {
         for elt in elts {
-            let attr = expr(
+            let element_attributes = assignable_expr(
                 Rc::clone(&scope),
                 context,
                 elt,
                 Some(&Type::Base(expected.inner)),
             )?;
-            if attr.typ != Type::Base(expected.inner) {
-                context.type_error("type mismatch", elt.span, &expected.inner, attr.typ);
+            if element_attributes.typ != Type::Base(expected.inner) {
+                context.type_error(
+                    "type mismatch",
+                    elt.span,
+                    &expected.inner,
+                    element_attributes.typ,
+                );
             }
         }
         expected.inner
     } else {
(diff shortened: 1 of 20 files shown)


cburgdorf (approved):
Looks good :+1:

PR #500, opened 2021-08-06
Changed function definition "def" to "fn"

andy-ow:
### What was wrong?
We should use a symbol derived from the word "function" for function definitions. `fn` is concise and used in rust, so we should just go with that.

### How was it fixed?
Changed "def" to "fn" in all files and tests. Verified tests pass.
This is my first github pull request ever.

### To-Do

- [ ] Clean up commit history


cburgdorf:
Nice! Thanks for the PR. Do you mind adding a file `496.feature.md` to https://github.com/ethereum/fe/tree/master/newsfragments
with the content being something like:

>Function definitions are now denoted with the keyword `fn` instead of `def`. 

This is needed so that the change is highlighted in the next release notes.

cburgdorf:
Oh and I think there are a bunch of markdown files in `/docs` that need to be changed as well.

andy-ow:
> Do you mind adding a file 496.feature.md

> Oh and I think there are a bunch of markdown files in `/docs` that need to be changed as well.

Ok, I will take a look today or tomorrow.

andy-ow:
Done.

andy-ow:
I did a rebase so that I could change a new added file on the main branch. Which was a bit confusing as it was my first rebase with git, but I think it's ok now! And changed a few files which I missed the first time.

cburgdorf:
Hey Andrzej, happy to see you are landing your first Pull Request with Fe! There were a few conflicts to resolve and I also rolled back a few changes (changes to the past release notes as well as to the Python grammar file) but other than that it looks good! I'm waiting for the CI system to pass and then I'll merge this. Thank you very much for your work!

Oh and out of curiosity, am I correct to assume that you and @mjobuda are brothers?

andy-ow:
Hey Christoph, yes your assumption is correct!

Issue #497, opened 2021-08-04
Return type isn't always properly checked


cburgdorf:
### What is wrong?

We should expect the following to fail because the function will implicitly return `()` but its signature says it returns `bool`. It currently compiles but it shouldn't.

```
contract Foo:

  def bar() -> bool:
    if 1 == 2:
      return true
```

### How can it be fixed

The analyzer needs to reject this with a proper user error.

PR #502, opened 2021-08-06
Reject certain code that should return but doesn't

cburgdorf:
### What was wrong?

As described in #497 the following code does compile in master but should generate a user error instead. It needs to error because the code is missing a return statement since it is not guaranteed that the execution follows the arm of the `if` statement.

```
contract Foo:
    pub def bar(val: u256) -> u256:
        if val > 1:
            return 5
```

### How was it fixed?

The check in `all_paths_return_or_revert` is wrong because it counts a missing `else` branch as if the code returns when clearly a missing `else` branch shouldn't indicate such thing.

### DIFF
--- a/crates/analyzer/src/traversal/functions.rs
+++ b/crates/analyzer/src/traversal/functions.rs
@@ -174,7 +174,7 @@ fn all_paths_return_or_revert(block: &[Node<fe::FuncStmt>]) -> bool {
                 or_else,
             } => {
                 let body_returns = all_paths_return_or_revert(body);
-                let or_else_returns = or_else.is_empty() || all_paths_return_or_revert(or_else);
+                let or_else_returns = all_paths_return_or_revert(or_else);
                 if body_returns && or_else_returns {
                     return true;
                 }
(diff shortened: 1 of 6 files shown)


crates/test-files/fixtures/compile_errors/missing_return_in_else.fe
```diff
@@ -3,4 +3,4 @@ contract Foo:
         if val > 1:
             return 5
         else:
-            x = 1
```
cburgdorf, line 6: This was an extra error in the test file that I think we didn't intent to test for here

crates/analyzer/src/traversal/functions.rs
```diff
@@ -174,7 +174,7 @@ fn all_paths_return_or_revert(block: &[Node<fe::FuncStmt>]) -> bool {
                 or_else,
             } => {
                 let body_returns = all_paths_return_or_revert(body);
-                let or_else_returns = or_else.is_empty() || all_paths_return_or_revert(or_else);
+                let or_else_returns = !or_else.is_empty() && all_paths_return_or_revert(or_else);
```
g-r-a-n-t, line 177: we could just get rid of the `is_empty` check, as `all_paths_return_or_revert(or_else)` will return false for an empty vec

g-r-a-n-t (approved):

crates/analyzer/src/traversal/functions.rs
cburgdorf: Ah sure, I'll get that updated :+1:

Issue #501, opened 2021-08-06
Use let keyword for variable declarations


cburgdorf:
### What is wrong?

Currently variables are declared like this:

```
val: bool = true
```

In the future we can assume to add some form of type inference so that the above code would be shortened to:

```
val = true
```

In that form there is nothing to distinguish a variable declaration from other assignments. In other words, these two lines look identical even though the first line declares the variable `var` whereas the second line assigns it a new value.

```
val = true
val = false
```

Making variable declarations more explicit seems in line with the goals of the language. Furthermore, in the future we likely want to make variables immutable by default which is yet another reason to follow Rusts lead with `let val = true` and `let mut val = true` syntax.

### How can it be fixed

Require variable declarations to use the `let` keyword.

Example:

```
let val: bool = true
```

sbillig:
Implemented in #509

PR #509, opened 2021-08-11
Feat/let var decl

WilfredTA:
### What was wrong?

Described in this issue: https://github.com/ethereum/fe/issues/501

Briefly, there is no explicit syntactic distinction between variable declarations and reassignments.

### How was it fixed?

Added the requirement that variable declarations are preceded by `let` keyword, e.g., `let x: u8 = 1`. Updated all fixtures and demo contracts to be compatible with this change.

### To do

Update documentation & website examples to reflect this change.

### DIFF
--- a/crates/parser/src/grammar/functions.rs
+++ b/crates/parser/src/grammar/functions.rs
@@ -205,8 +205,46 @@ pub fn parse_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
         Revert => parse_revert_stmt(par),
         Continue | Break | Pass => parse_single_word_stmt(par),
         Emit => parse_emit_statement(par),
+        Let => parse_var_decl(par),
         _ => parse_expr_stmt(par),
     }
+}
+
+fn parse_var_decl(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
+
+    let let_tkn = par.assert(TokenKind::Let);
+    let expr = parse_expr(par)?;
+    let target = expr_to_vardecl_target(par, expr.clone())?;
+    let node = match par.peek() {
+        Some(TokenKind::Colon) => {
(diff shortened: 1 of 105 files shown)


crates/parser/src/grammar/functions.rs
```diff
@@ -205,10 +205,48 @@ pub fn parse_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
...
+        },
+        _ => {
+            par.fancy_error(
+                "failed to parse variable declaration",
+                vec![Label::primary(
+                    expr.span,
+                    "Must be followed by type annotation",
+                )],
+                vec![
+                    "The left side of a variable declaration can be either a name\nor a non-empty tuple."
```
sbillig, line 240: This "left side of a variable ..." note doesn't really describe the error here (missing colon); could you remove it or replace it with something like "Example: `let x: u8 = 1`"?

crates/parser/src/grammar/functions.rs
```diff
@@ -218,20 +256,34 @@ fn parse_expr_stmt(par: &mut Parser) -> ParseResult<Node<FuncStmt>> {
...
+            // let target = expr_to_vardecl_target(par, expr)?;
+            // let typ = parse_type_desc(par)?;
+            // let value = if par.peek() == Some(Eq) {
+            //     par.next()?;
+            //     Some(parse_expr(par)?)
+            // } else {
+            //     None
+            // };
+            // let span = target.span + typ.span + value.as_ref();
+            // Node::new(FuncStmt::VarDecl { target, typ, value }, span)
```
sbillig, line 284: Could you cut this block of commented-out code?

sbillig (commented):
Looks great! Added a couple nitpicky comments.

Issue #470, opened 2021-06-28
failed to analyze lowered AST: AnalyzerError([Diagnostic { severity: Error, code: None, message: "No field...  at compiler/src/lib.rs:55:59


agroce:
Compiling this file:

```
contract o:
 def m(my_tuple:(u256,bool,address))->u256:
  return my_tuple.item00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

```
thread 'main' panicked at 'failed to analyze lowered AST: AnalyzerError([Diagnostic { severity: Error, code: None, message: "No field `item00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000` exists on struct `tuple_u256_bool_address_`", labels: [Label { style: Primary, file_id: SourceFileId(226182870253612458078089501400146892043), range: 74..449, message: "undefined field" }], notes: [] }])', compiler/src/lib.rs:55:59
stack backtrace:
   0:     0x5571da8c29b0 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x5571da8c29b0 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5571da8c29b0 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
                               at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x5571da8c29b0 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
[...]

PR #510, opened 2021-08-11
Tuple item leading zeroes #470

andy-ow:
### What was wrong?
issue #470  
Leading zeros in tuple indexes should not be allowed.

### How was it fixed?
Check for leading zeros in tuple indexes (like item00 or item001). Small change in expressions.rs + tests.

### To-Do

- [x] Clean up commit history


sbillig:
@andy-ow Looks great, thanks! Could you add a simple `newsfragments/510.bugfix.md` describing what was fixed? (Add a trailing newline on that file, or the lint check will reject it, btw)

There's also a minor conflict with the master branch now that I've merged #509. Would you mind rebasing?

sbillig:
@andy-ow I handled the chores I mentioned above; you weren't taking too long, I'd just like to get this PR merged in prior to #468, which would have caused a lot more merge conflicts

Issue #517, opened 2021-08-19
Function calls don't properly check named arguments


cburgdorf:
### What is wrong?

The following compiles when clearly the named arguments are wrong:

```
contract Foo:

    pub fn baz():
        self.bar(doesn_exist=1, me_neither=2)
    
    pub fn bar(val1: u256, val2: u256):
        pass
```

### How can it be fixed

Add a check in the analyzer that ensures named arguments have to conform to the arguments of the function being called.

PR #518, opened 2021-08-19
Check labels in function calls on self and external contracts

cburgdorf:
### What was wrong?

As explained in #517, function calls on `self` or external contracts do not check the labels, allowing developers to apply any kind of label and in the worst case leave the user with the impression that they could use labels to call functions with the arguments ordered as they wish (which isn't the case).

### How was it fixed?

Added checks in the analyzer and added tests.

### DIFF
--- a/crates/analyzer/src/traversal/call_args.rs
+++ b/crates/analyzer/src/traversal/call_args.rs
@@ -55,9 +55,10 @@ pub fn validate_named_args(
     name_span: Span,
     args: &Node<Vec<Node<fe::CallArg>>>,
     params: &[impl LabeledParameter],
+    label_policy: LabelPolicy,
 ) -> Result<(), FatalError> {
     validate_arg_count(scope, name, name_span, args, params.len());
-    validate_arg_labels(scope, args, params);
+    validate_arg_labels(scope, args, params, label_policy);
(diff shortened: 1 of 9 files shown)


sbillig (approved):
Looks great! `validate_arg_types` and `validate_arg_labels` were only separate because I failed to think of a scheme like this. If you were so inclined (no pressure), you could combine the two so that we could not report a type error in the case that the label is incorrect (or maybe just when the labels are in the incorrect order).

cburgdorf:
>so that we could not report a type error in the case that the label is incorrect (or maybe just when the labels are in the incorrect order).

I think I don't follow.

This will only report label errors because the types are correct but mislabeled:

```
contract Foo:

    pub fn bar():
        self.baz(some=1, input=true)

    pub fn baz(input: u256, some: bool):
        pass
```

This will report label and type error because they are mislabled *and* in the wrong order:

```
contract Foo:

    pub fn bar():
        self.baz(some=true, input=1)

    pub fn baz(input: u256, some: bool):
        pass
```

Isn't that already what we want?

I'm going to merge this already but happy to follow up if I understand the intention of the proposed change. Or feel free to send a PR if it's less effort to explain it with the code directly :D

PR #520, opened 2021-08-23
`self` declarations

g-r-a-n-t:
### What was wrong?

`self` was implicitly defined in all function scopes, which made it impossible to write pure functions.

### How was it fixed?

1.) added `self` args to the function definition grammar
2.) added checking around the use of `self`
3.) fixed things that broke

I also went ahead and updated the Json ABI builder to label functions that do not take a `self` parameter as pure. This is something that I'm not sure we want right now though, as functions can still read msg and chain data (which conflicts with the definition of pure). For this reason, I'm considering dropping the [commit](https://github.com/ethereum/fe/pull/520/commits/1fedb01b96c0f981fe031ff7dfc2225da41af2de) responsible for this change, but have not 100% made up my mind. Feel free to provide input.

### To-Do

- [x] Update [Spec](https://github.com/ethereum/fe/blob/master/docs/src/spec/index.md) if applicable
- [x] Clean up commit history
- [x] Update grammar

### DIFF
--- a/crates/analyzer/src/namespace/scopes.rs
+++ b/crates/analyzer/src/namespace/scopes.rs
@@ -79,6 +79,14 @@ impl<'a> FunctionScope<'a> {
         self.function.parent(self.db).function(self.db, name)
     }
 
+    pub fn self_contract_function(&self, name: &str) -> Option<FunctionId> {
+        self.function.parent(self.db).self_function(self.db, name)
+    }
+
+    pub fn pure_contract_function(&self, name: &str) -> Option<FunctionId> {
+        self.function.parent(self.db).pure_function(self.db, name)
+    }
+
     pub fn resolve_event(&self, name: &str) -> Option<EventId> {
         self.function.parent(self.db).event(self.db, name)
     }
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -9,9 +9,7 @@ use crate::namespace::types::{
     Array, Base, Contract, FeString, Integer, SelfDecl, Struct, Tuple, Type, TypeDowncast, U256,
 };
 use crate::operations;
-use crate::traversal::call_args::{
-    validate_arg_count, validate_arg_types, validate_named_args, LabelPolicy,
-};
+use crate::traversal::call_args::{validate_arg_count, validate_named_args, LabelPolicy};
 use crate::traversal::utils::{add_bin_operations_errors, types_to_fixed_sizes};
 use fe_common::diagnostics::Label;
 use fe_common::numeric;
@@ -573,7 +571,7 @@ fn expr_attribute_self(
     value: &Node<fe::Expr>,
     attr: &Node<String>,
 ) -> Result<ExpressionAttributes, FatalError> {
-    check_for_self_param(scope, value.span)?;
+    check_for_self_param(scope, value.span);
 
     if let Ok(SelfField::Address) = SelfField::from_str(&attr.kind) {
         return Ok(ExpressionAttributes::new(
@@ -886,7 +884,7 @@ fn check_for_call_to_init_fn(
     }
 }
 
-fn check_for_self_param(scope: &mut BlockScope, use_span: Span) -> Result<(), FatalError> {
+fn check_for_self_param(scope: &mut BlockScope, use_span: Span) {
     if scope.root.function.signature(scope.db()).self_decl == SelfDecl::None {
         scope.fancy_error(
             "`self` is not defined",
@@ -896,9 +894,6 @@ fn check_for_self_param(scope: &mut BlockScope, use_span: Span) -> Result<(), Fa
                 "Example: `fn foo(self, bar: bool)`".to_string(),
             ],
         );
-        Err(FatalError::new())
-    } else {
-        Ok(())
     }
 }
 
@@ -910,12 +905,18 @@ fn expr_call_self_attribute(
     args: &Node<Vec<Node<fe::CallArg>>>,
 ) -> Result<ExpressionAttributes, FatalError> {
     check_for_call_to_init_fn(scope, func_name, name_span)?;
-    check_for_self_param(scope, self_span)?;
+    check_for_self_param(scope, self_span);
 
     if let Some(func) = scope.root.self_contract_function(func_name) {
         let sig = func.signature(scope.root.db);
-        validate_arg_count(scope, func_name, name_span, args, sig.params.len());
-        validate_arg_types(scope, func_name, args, &sig.params)?;
+        validate_named_args(
+            scope,
+            func_name,
+            name_span,
+            args,
+            &sig.params,
+            LabelPolicy::AllowAnyUnlabeled,
+        )?;
 
         let return_type = sig.return_type.clone()?;
         let return_location = Location::assign_location(&return_type);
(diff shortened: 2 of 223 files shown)


g-r-a-n-t:
@cburgdorf I saw your comment about using `Self::pure_func()`, but it appears to have been wiped in a rebase :thinking:.

This is what I was thinking too.

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -873,15 +879,65 @@ fn check_for_call_to_init_fn(
...
+fn expr_call_pure(
     scope: &mut BlockScope,
     func_name: &str,
     name_span: Span,
     args: &Node<Vec<Node<fe::CallArg>>>,
 ) -> Result<ExpressionAttributes, FatalError> {
     check_for_call_to_init_fn(scope, func_name, name_span)?;
 
-    if let Some(func) = scope.root.contract_function(func_name) {
+    if let Some(func) = scope.root.pure_contract_function(func_name) {
```
g-r-a-n-t, line 940: the error for attempting to call a `self` function as pure should be more helpful

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -873,15 +879,65 @@ fn check_for_call_to_init_fn(
...
+    scope: &mut BlockScope,
+    func_name: &str,
+    name_span: Span,
+    self_span: Span,
+    args: &Node<Vec<Node<fe::CallArg>>>,
+) -> Result<ExpressionAttributes, FatalError> {
+    check_for_call_to_init_fn(scope, func_name, name_span)?;
+    check_for_self_param(scope, self_span)?;
+
+    if let Some(func) = scope.root.self_contract_function(func_name) {
```
g-r-a-n-t, line 911: similar to the other comment, this error should be more helpful

crates/analyzer/src/namespace/scopes.rs
```diff
@@ -79,6 +79,14 @@ impl<'a> FunctionScope<'a> {
         self.function.parent(self.db).function(self.db, name)
     }
 
+    pub fn self_contract_function(&self, name: &str) -> Option<FunctionId> {
```
g-r-a-n-t, line 82: docs

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -873,15 +880,73 @@ fn check_for_call_to_init_fn(
...
+        let sig = func.signature(scope.root.db);
+        validate_arg_count(scope, func_name, name_span, args, sig.params.len());
+        validate_arg_types(scope, func_name, args, &sig.params)?;
```
cburgdorf, line 912: This is recreating a bug that was fixed recently in #518

Here's a snippet that doesn't cause a compile error with this change but should:

```
contract Foo:

    pub fn baz(self):
        self.bar(doesnt_exist=1, me_neither=2)
    
    pub fn bar(self, val1: u256, val2: u256):
        pass
```

We should add this as a test case variant of the `mislabeled_call_args.fe` fixture that we already have (but which doesn't apply anymore since it doesn't use `self`)

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -873,15 +880,73 @@ fn check_for_call_to_init_fn(
...
+    if scope.root.function.signature(scope.db()).self_decl == SelfDecl::None {
+        scope.fancy_error(
+            "`self` is not defined",
+            vec![Label::primary(use_span, "undefined value")],
+            vec![
+                "add `self` to the scope by including it in the function signature".to_string(),
+                "Example: `fn foo(self, bar: bool)`".to_string(),
+            ],
+        );
+        Err(FatalError::new())
```
cburgdorf, line 893: I think just adding the error should be enough and the function doesn't need to return anything. That way we could get rid of the `FatalError` which is preventing us from collecting more errors.

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -568,8 +570,11 @@ fn tuple_item_index(item: &str) -> Option<usize> {
 
 fn expr_attribute_self(
     scope: &mut BlockScope,
+    value: &Node<fe::Expr>,
     attr: &Node<String>,
 ) -> Result<ExpressionAttributes, FatalError> {
+    check_for_self_param(scope, value.span)?;
```
cburgdorf, line 576: According to my comment a few lines further down, I think we could call with without `?` and simply not return anything (to keep collecting errors).

cburgdorf (approved):
Nice work! Left a few comments inline. There's a regression introduced that should be addressed.

Regarding:

>I also went ahead and updated the Json ABI builder to label functions that do not take a `self` parameter as pure

I think we shouldn't mark them as `pure` in the ABI then imho. We could still mark it as `view` though even if it is technically a bit purer than `view` :)

I kinda dislike the current syntax of calling pure functions on the contract by just their name (e.g. `foo()`). You seem to be on board with demanding those to be prefixed with `Self::` (e.g. `Self::foo()`) but we should probably create an issue for that. Also paging in @sbillig for visibility. It's also not just an aesthetically question since the current syntax would become ambiguous with imported functions or functions that are defined outside of the contract but within the same file.

Since today is the last day of August and we should cut our monthly release, I'm a bit hesitant if we just want to leave this PR for the next release so that we should be able to bundle it with the `Self::` change?

docs/src/spec/functions.md
```diff
@@ -59,6 +59,22 @@ fn answer_to_life_the_universe_and_everything() -> u256:
     return 42
 ```
 
+A function may accept `self` as a parameter. This gives the function the ability
```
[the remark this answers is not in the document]
g-r-a-n-t: This is something that crossed my mind, but I ultimately decided that checking the position of `self` is better left to semantic analysis. By doing it this way, we simplify the grammar/parser and still enforce the rules mentioned above.

This is how the [Rust grammar](https://github.com/rust-lang/wg-grammar/blob/7a1bd645fd50d8783da7e2010da8bcc226d656ea/grammar/item.lyg#L88-L100) works and from what I understand, the [Python grammar](https://docs.python.org/3/reference/grammar.html) doesn't even mention `self`.

On a side note, correct grammar does not necessarily mean that a program is valid. Where errors are handled is at our own discretion, and in this case, I thought it made sense to split the validation of `self` between the parser and analysis in such a way that kept the grammar simple and made it possible to return a helpful error.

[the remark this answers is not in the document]
cburgdorf: Ok, that might be true for the grammar file but in their rust reference they don't seem to care about at which stage the syntax become illegal, just that it is illegal: https://doc.rust-lang.org/nightly/reference/items/functions.html

sbillig:
> I kinda dislike the current syntax of calling pure functions on the contract by just their name (e.g. foo()). 

I'm not sure how I feel about this; it seems like `Self::` would would be unnecessary noise here.  If `fn foo()` is defined in the current contract scope, why require the scope to be specified? We aren't going to require a scope prefix for functions defined at the module level and called within the same module; why should functions defined in a contract be treated differently?

> It's also not just an aesthetically question since the current syntax would become ambiguous with imported functions or functions that are defined outside of the contract but within the same file.

The analyzer can check for ambiguity, and require the user to disambiguate.

docs/src/spec/functions.md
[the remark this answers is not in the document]
g-r-a-n-t: I see. Spec has been updated and grammar has been left as is.

sbillig (approved):
Looks great. I guess I agree non-self fns should be called views until we can be sure that they don't access state (but also think that it doesn't really matter at this stage of development).

cburgdorf:
>The analyzer can check for ambiguity, and require the user to disambiguate.

That's true but I kinda like the added clarity of avoiding the ambiguity in the first place :)

Issue #521, opened 2021-08-24
Implement unary ~ (bitwise NOT) operator


cburgdorf:
### What is wrong?

The `~` operator isn't yet supported in the code generation phase.

### How can it be fixed

Probably just implementing it as `expression! { not([yul_operand]) }` here

https://github.com/ethereum/fe/blob/a0821aa5aa08ab656be3424de4f2bed67775f2b3/crates/yulgen/src/mappers/expressions.rs#L281

Plus adding some tests.

PR #526, opened 2021-08-30
Add support for numeric invert operator

cburgdorf:
### What was wrong?

As noted in #521 we needed to add support for the `~` operator which is currently recognized on the parser level but not analyzed and generates no code.

### How was it fixed?

Implementation is pretty straight forward. The more interesting part is that this PR lays the ground work for a bunch of new `adjust_numeric_*` runtime functions that we need to use at a bunch more places to ensure numerics smaller than 256 bit are handled correctly. This PR only uses these methods for the unary invert operator.


crates/tests/src/features.rs
```diff
@@ -161,6 +161,19 @@ fn test_assert() {
     case("return_msg_sig.fe", &[], uint_token(4273672062)),
     case("return_sum_list_expression_1.fe", &[], uint_token(210)),
     case("return_sum_list_expression_2.fe", &[], uint_token(210)),
+    // unary invert
+    case("return_invert_i256.fe", &[int_token(1)], int_token(-2)),
+    case("return_invert_i128.fe", &[int_token(1)], int_token(-2)),
+    case("return_invert_i64.fe", &[int_token(4000000000)], int_token(-4000000001)),
+    case("return_invert_i32.fe", &[int_token(30000)], int_token(-30001)),
+    case("return_invert_i16.fe", &[int_token(2000)], int_token(-2001)),
+    case("return_invert_i8.fe", &[int_token(1)], int_token(-2)),
```
g-r-a-n-t, line 170: Is there a special use for signed ints that I'm missing?

Since this is a bitwise operation, I think use should be restricted to unsigned ints, just like the binary bitwise operations.

crates/yulgen/src/runtime/functions/math.rs
```diff
@@ -287,6 +304,31 @@ fn _checked_exp_unsigned(size: Integer) -> yul::Statement {
...
+}
+
+fn adjust_numeric_signed(size: Integer, base: usize) -> yul::Statement {
+    if !size.is_signed() {
+        panic!("Expected signed integer")
+    }
+    let fn_name = names::adjust_numeric_size(&size);
+    function_definition! {
+        function [fn_name](value) -> cleaned {
+            (cleaned := signextend([literal_expression! { (base) }], value))
```
g-r-a-n-t, line 327: Nice, this solution looks good.

g-r-a-n-t (approved):

crates/yulgen/src/mappers/expressions.rs
```diff
@@ -272,13 +272,26 @@ pub fn expr_unary_operation(context: &mut FnContext, exp: &Node<fe::Expr>) -> yu
...
         return match &op.kind {
             fe::UnaryOperator::USub => {
                 let zero = literal_expression! {0};
                 expression! { sub([zero], [yul_operand]) }
             }
             fe::UnaryOperator::Not => expression! { iszero([yul_operand]) },
-            _ => todo!(),
+            fe::UnaryOperator::Invert => match typ {
+                Type::Base(Base::Numeric(integer)) if integer.size() < 32 => {
+                    expression! { [names::adjust_numeric_size(integer)]((not([yul_operand]))) }
```
g-r-a-n-t, line 288: Oh, we should also have operations for these adjustment functions. This way we are constructing calls to these functions in a single place.

e.g. 

replace

```rust
{
  expression! { [names::adjust_numeric_size(integer)]((not([yul_operand]))) }
}
```

with

```
numeric_operations::adjust_numeric_size(integer, yul_operand)
```

crates/tests/src/features.rs
cburgdorf: I haven't thought much about use cases but Solidity, Python, Rust (and probably other languages) all let you do this so I think we shouldn't be overly restrictive.

Solidity
```solidity
int8 y = -10;
assert(~y == 9);
```

Python:
```
>>> ~-10
9
```

Rust:
```rust
    assert_eq!(!-10i8, 9);
```

crates/yulgen/src/runtime/functions/math.rs
cburgdorf: Just copied what Solidity is doing. We still need to apply that to a bunch of other places (which I will follow up on after this PR)

crates/yulgen/src/mappers/expressions.rs
cburgdorf: :+1: done

PR #536, opened 2021-09-06
Module level const support

cburgdorf:
### To-Do

- [ ] Clean up commit history


crates/lowering/tests/snapshots/lowering__module_const.snap
```diff
@@ -0,0 +1,21 @@
+---
+source: crates/lowering/tests/lowering.rs
+expression: lowered_code
+
+---
+const SOME_THING: u256 = 1 + 2
```
g-r-a-n-t, line 6: can we drop these definitions from the AST?

cburgdorf: Makes sense :+1:

PR #538, opened 2021-09-06
Report all parser errors

g-r-a-n-t:
### What was wrong?

This is similar to #535. @sbillig I also happened to be looking at this stuff late last week, but was pretty busy and didn't get around to making a PR. It looks like the parser and doc snippet tests are missing things with the way we silently emit non-fatal errors, so I'm fixing those in this PR.

### To-Do

- [ ] Clean up commit history


g-r-a-n-t:
Oh, wait I see how these tests are supposed to work now. `should_fail` indicates whether or not the error is fatal.

I guess we still want the doc snippet tests to fail if there are any diagnostics.

g-r-a-n-t:
well, I guess the docs validation has been fixed too. closing this pointless PR :laughing:

PR #539, opened 2021-09-06
Add analyzer checks for shadowing of built-ins and generic arg lists where they don't belong

sbillig:
### What was wrong?

The analyzer allowed things like:

```
type u8 = u256
let u256: u8 = 10
fn u8(u8: u256) -> u8:
let keccak256: u256 = keccak256(x)
let x: u256 = u256<1, 2, 3>(10)
```

### How was it fixed?

Added more checks to type, function, and var defs. This implementation isn't super satisfying; I was able to consolidate some duplicate logic, but we still have to perform this kind of check in 3 places, which doesn't seem ideal.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/src/builtins.rs
+++ b/crates/analyzer/src/builtins.rs
@@ -8,7 +8,7 @@ pub enum ValueMethod {
     AbiEncode,
 }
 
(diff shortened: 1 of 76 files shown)


g-r-a-n-t (approved):
:fire: :100:

PR #540, opened 2021-09-07
Pretty printing of Fe AST

### LINKED ISSUE #531: failed to analyze lowered AST (duplicate type name)
Compiling this file:

```
struct tuple_bool_address_:
 m:bool
contract o:
 e:(bool,address)
```

```
thread 'main' panicked at 'failed to analyze lowered AST: [Diagnostic { severity: Error, message: "duplicate type name", labels: [Label { style: Primary, span: Span { start: 0, end: 0 }, message: "`tuple_bool_address_` first defined here" }, Label { style: Secondary, span: Span { start: 0, end: 35 }, message: "`tuple_bool_address_` redefined here" }], notes: [] }]', crates/driver/src/lib.rs:62:51
stack backtrace:
   0:     0x5558642d6100 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
   1:     0x5558642d6100 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
   2:     0x5558642d6100 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
   3:     0x5558642d6100 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
   4:     0x5558642eee4f - core::fmt::write::h3b84512577ca38a8
   5:     0x5558642d3a22 - std::io::Write::write_fmt::h465f8feea02e2aa1
   6:     0x5558642d8295 - std::sys_common::backtrace::_print::h525280ee0d29bdde
   7:     0x5558642d8295 - std::sys_common::backtrace::print::h1f0f5b9f3ef8fb78
   8:     0x5558642d8295 - std::panicking::default_hook::{{closure}}::ha5838f6faa4a5a8f
   9:     0x5558642d7d43 - std::panicking::default_hook::hfb9fe98acb0dcb3b
  10:     0x555864221b66 - fe_common::panic::DEFAULT_PANIC_HOOK::{{closure}}::{{closure}}::hfa0d7480b545d3da

Aborted
```

Using

```
[profile.dev]
debug = true
panic = "abort"
```

g-r-a-n-t:
### What was wrong?

We can't print the AST.

I wanted to fix #531 by adding a `$` to the front of lowered tuple struct names (e.g. `$tuple_bool_u256`), but wasn't able to do so without breaking the lowering tests. This is because the lowering tests compare the RON serialized AST of a lowered source file to that of an AST from a parsed file. So by adding a character to the lowered AST that is illegal to the parser, we are unable to validate that part of the lowering phase.

By making the AST printable, we can replace the existing lowering tests with insta snapshots and not have to be bound by grammar rules when lowering.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/parser/src/ast.rs
+++ b/crates/parser/src/ast.rs
@@ -1,7 +1,10 @@
 use crate::node::Node;
 use fe_common::{Span, Spanned};
+use indenter::indented;
 use serde::{Deserialize, Serialize};
 use std::fmt;
+use std::fmt::Formatter;
+use std::fmt::Write;
 use vec1::Vec1;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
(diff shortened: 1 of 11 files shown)


sbillig:
@g-r-a-n-t Does Display give us enough functionality to indent the tree so that it's pretty?

Also, if I had my druthers, the parser ast tests would compare asts with spans and the text content of the span, but other ast printing would ignore the spans. Eg from rowan https://crates.io/crates/rowan
```
Root@0..7
  BinaryExpr@0..7
    BinaryExpr@0..5
      BinaryExpr@0..3
        Literal@0..1
          Number@0..1 "1"
        Plus@1..2 "+"
        Literal@2..3
          Number@2..3 "2"
      Plus@3..4 "+"
      Literal@4..5
        Number@4..5 "3"
    Plus@5..6 "+"
    Literal@6..7
      Number@6..7 "4"
```
Not suggesting you have to implement this; just something to consider as you ponder the design.

g-r-a-n-t:
> Does Display give us enough functionality to indent the tree so that it's pretty?

I'm using [indenter](https://docs.rs/indenter/0.3.3/indenter/) the same way it's [used](https://github.com/g-r-a-n-t/yultsur/blob/ae854702e90b2c70e5c19961167f2b6f2aff32d2/src/yul.rs#L193-L198) in Yultsur's AST Display impl, so indents shouldn't be a problem. It may be reasonable to not use Display for other design reasons, but I can't think of any, so I just started doing it this way.

crates/lowering/src/names.rs
```diff
@@ -9,7 +9,7 @@ pub fn list_expr_generator_fn_name(list_expr_type: &Array) -> String {
 
 /// The name of a lowered tuple struct definition.
 pub fn tuple_struct_name(tuple: &Tuple) -> String {
-    tuple.lower_snake()
+    format!("${}", tuple.lower_snake())
```
g-r-a-n-t, line 12: do this in another PR

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,327 @@ impl Spanned for ContractStmt {
...
+impl fmt::Display for GenericArg {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            GenericArg::TypeDesc(node) => write!(f, "{}", node.kind),
+            GenericArg::Int(node) => write!(f, "{}", node.kind),
+        }
+    }
+}
+
+impl fmt::Display for SimpleImportName {
```
g-r-a-n-t, line 476: I just left the imports out for now, since we may not be settled on their design.

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,327 @@ impl Spanned for ContractStmt {
...
+                }
+            }
+            FuncStmt::Emit { name, args } => {
+                write!(f, "emit {}({})", name.kind, node_comma_joined(&args.kind))
+            }
+            FuncStmt::Expr { value } => write!(f, "{}", value.kind),
+            FuncStmt::Pass => write!(f, "pass"),
+            FuncStmt::Break => write!(f, "break"),
+            FuncStmt::Continue => write!(f, "continue"),
+            FuncStmt::Revert { .. } => write!(f, "revert"),
```
g-r-a-n-t, line 632: finish this

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,327 @@ impl Spanned for ContractStmt {
...
+            }
+            Expr::UnaryOperation { op, operand } => {
+                let operand = maybe_fmt_inner_with_parens(self, &operand.kind);
+                if op.kind == UnaryOperator::Not {
+                    write!(f, "{} {}", op.kind, operand)
+                } else {
+                    write!(f, "{}{}", op.kind, operand)
+                }
+            },
+            Expr::CompOperation { left, op, right } => {
```
g-r-a-n-t, line 676: do these need parens?

crates/parser/src/ast.rs
```diff
@@ -439,3 +763,85 @@ impl fmt::Display for CompOperator {
...
+
+fn maybe_fmt_inner_with_parens(outer: &Expr, inner: &Expr) -> String {
+    if inner_expr_needs_parens(outer, inner) {
+        format!("({})", inner)
+    } else {
+        format!("{}", inner)
+    }
+}
+
+fn inner_expr_needs_parens(outer: &Expr, inner: &Expr) -> bool {
```
g-r-a-n-t, line 799: comment

crates/parser/src/ast.rs
```diff
@@ -439,3 +763,85 @@ impl fmt::Display for CompOperator {
...
+    expr_power(outer) > expr_power(inner)
+}
+
+fn expr_power(expr: &Expr) -> usize {
```
g-r-a-n-t, line 803: comment

crates/parser/src/ast.rs
g-r-a-n-t: I dont think so

crates/parser/src/ast.rs
```diff
@@ -439,3 +769,93 @@ impl fmt::Display for CompOperator {
...
+/// e.g.
+/// In the expression `2 * (3 + 4)`, the power of an `*` operation is greater than the power of
+/// an `+` operation. If we do not account for this by wrapping the `+` operation in parentheses,
+/// then the expression changes meaning.
+fn inner_expr_needs_parens(outer: &Expr, inner: &Expr) -> bool {
+    expr_power(outer) > expr_power(inner)
+}
+
+/// Similar to binding power, but in terms of a whole expressions, as opposed to a token.
+fn expr_power(expr: &Expr) -> usize {
```
g-r-a-n-t, line 817: Im not sure if there is better terminology to use here..

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,333 @@ impl Spanned for ContractStmt {
...
+                    write!(f, "<{}>", comma_joined(&generic_args.kind))?;
+                }
+                write!(f, "({})", node_comma_joined(&args.kind))
+            }
+            Expr::List { elts } => write!(f, "[{}]", node_comma_joined(elts)),
+            Expr::Tuple { elts } => write!(f, "({})", node_comma_joined(elts)),
+            Expr::Bool(bool) => write!(f, "{}", bool),
+            Expr::Name(name) => write!(f, "{}", name),
+            Expr::Num(num) => write!(f, "{}", num),
+            Expr::Str(str) => write!(f, "{}", str),
```
g-r-a-n-t, line 703: this might need quotes

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,333 @@ impl Spanned for ContractStmt {
     }
 }
 
+impl fmt::Display for Module {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        for stmt in self.body.iter() {
```
g-r-a-n-t, line 394: use line join function

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,333 @@ impl Spanned for ContractStmt {
...
+        }
+
+        Ok(())
+    }
+}
+
+impl fmt::Display for Struct {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        writeln!(f, "struct {}:", self.name.kind)?;
+        writeln!(indented(f), "{}", node_comma_joined(&self.fields))
```
g-r-a-n-t, line 449: should be line joined

crates/parser/src/ast.rs
```diff
@@ -386,6 +389,333 @@ impl Spanned for ContractStmt {
...
+            ContractStmt::Event(node) => write!(f, "{}", node.kind),
+            ContractStmt::Function(node) => write!(f, "{}", node.kind),
+        }
+    }
+}
+
+impl fmt::Display for Event {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        writeln!(f, "event {}:", self.name.kind)?;
+        for field in self.fields.iter() {
```
g-r-a-n-t, line 524: use line join function

g-r-a-n-t (commented):
noticed a few more things while working on #546

crates/parser/src/ast.rs
```diff
@@ -439,3 +769,93 @@ impl fmt::Display for CompOperator {
...
+/// An inner expression needs parentheses if the power of the outer expression is greater than
+/// that of the inner.
+///
+/// e.g.
+/// In the expression `2 * (3 + 4)`, the power of an `*` operation is greater than the power of
+/// an `+` operation. If we do not account for this by wrapping the `+` operation in parentheses,
+/// then the expression changes meaning.
+fn inner_expr_needs_parens(outer: &Expr, inner: &Expr) -> bool {
+    expr_power(outer) > expr_power(inner)
+}
```
sbillig, line 814: The paren insertion logic isn't right. Eg the parens of `2 * (3 / 4)` are dropped. For left-associative ops, using `>` for the lhs of a binop, and `>=` for the rhs would suffice, but `**` is right-associative, so `(2 ** 3) ** 4` needs to keep its parens but `2 ** (3 ** 4)` doesn't.

Maybe there's a more elegant way, but I think of it as left and right binding power, ie the amount of attraction power an op has on its left and right operands, as in `infix_binding_power`. A single `expr_power` fn isn't sufficient, as the power an outer expr exerts on its lhs and rhs might be different, and the power an inner expr   responds with may differ depending on whether it's on the left or right.

crates/parser/src/ast.rs
sbillig: Yeah, eg `(1 == 2) == (3 == 4)`

crates/parser/src/ast.rs
g-r-a-n-t: Ok, updated it to use left and right binding power. I think it's correct now.

crates/parser/src/ast.rs
g-r-a-n-t: `(a == b) == (c == d)` is equivalent to  `1 == 2 == (3 == 4)`, right? this is at least what the test are telling me, as the first set of parens are dropped when printing

g-r-a-n-t:
I'm merging this @sbillig.

I'm fairly confident the paren stuff is correct now. It's also only going to be used for testing, so it's not something that needs to be perfect.

PR #544, opened 2021-09-10
Improved articulation of #overview

zora89:
### What was wrong?

The articulation of the overview heading was wobbly. Not reflecting precision and grace. 'and' was used twice in succession 'python and rust and' , making it awkward. 

### How was it fixed?

I rephrased while keeping the meaning precisely the same, and avoiding incorrect articulation methods. Minimal tweaking for a far better impact. Read it and see for yourself :)

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Fe is an emerging smart contract language for the Ethereum blockchain.
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


g-r-a-n-t:
Thank you! looks great.

Will merge once the build completes.

PR #547, opened 2021-09-20
Basic module system

g-r-a-n-t:
### What was wrong?

We can't use things defined in other modules.

### How was it fixed?

*Note: This is not going to be a complete module system implementation. I think starting with neighboring files and definition specific uses is enough for this. By definitions specific uses, I mean `use my_mod::my_func` would add `my_func` to the scope, while `use my_mod` would give a `not yet implemented` error.*

Here is what I think the implementation involves:

- [ ] replace pythonic `import` syntax with rusty `use` syntax
- [ ] update namespacing in the analyzer
- [ ] introduce module resolution phase that replaces use statements with full defintions

### To-Do

- [ ] Clean up commit history


crates/parser/src/grammar/module.rs
```diff
@@ -116,6 +116,124 @@ pub fn parse_from_import(par: &mut Parser) -> ParseResult<Node<Import>> {
...
+                    _ => panic!(),
+                }
+            }
+
+            UseTree::Nested {
+                prefix: base_path,
+                children: sub_trees,
+            }
+        }
+        Some(TokenKind::Star) => {
```
g-r-a-n-t, line 224: again, we can branch into based on the trailing delim

crates/parser/src/grammar/module.rs
```diff
@@ -116,6 +116,124 @@ pub fn parse_from_import(par: &mut Parser) -> ParseResult<Node<Import>> {
...
+            par.next()?;
+            UseTree::Glob { prefix: base_path }
+        }
+        Some(_) => UseTree::Simple {
+            path: base_path,
+            rename: None,
+        },
+    };
+
+    Ok(Node::new(uze, Span::zero()))
```
g-r-a-n-t, line 234: correct span

crates/test-files/fixtures/printing/defs.fe
```diff
@@ -1,3 +1,11 @@
+use foo::bar as baz
```
g-r-a-n-t, line 1: the double lines are a bit ugly here, but Im inclined to keep things simple and leave it alone

PR #549, opened 2021-09-24
Remove minor typo

jacobkaufmann:
### What was wrong?

Dangling "python" language specifier above code block

### How was it fixed?

Remove unnecessary "python"


cburgdorf:
Thanks for the fix :pray:

PR #563, opened 2021-10-08
Ensure expressions in list expressions are lowered

cburgdorf:
### What was wrong?

We missed to lower expressions inside list expressions.

### How was it fixed?

Call to `expr`

I didn't create a release notes entry for this because I don't think that in previous releases there was a way for an user to hit this bug. But they can hit it now with the `const` support.

### DIFF
--- a/crates/lowering/tests/snapshots/lowering__module_const.snap
+++ b/crates/lowering/tests/snapshots/lowering__module_const.snap
@@ -3,16 +3,47 @@ source: crates/lowering/tests/lowering.rs
 expression: lowered_code
 
 ---
+struct $tuple_u256_u256_:
+    item0: u256
+    item1: u256
+
+struct Bar:
+    val: u256
+
 contract Foo:
-    pub fn bar() -> u256:
-        if true:
-            return 3 * 10
-        else true:
-            return 10
-
-
-    pub fn baz() -> u256:
-        return 3 * 3
+    table: Map<u256, u256>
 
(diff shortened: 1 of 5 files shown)


crates/test-files/fixtures/lowering/module_const.fe
```diff
@@ -3,16 +3,38 @@ const TEN: u256 = 10
...
-  pub fn baz() -> u256:
-    return THREE * 3
+  table: Map<u256, u256>
 
   pub fn return_unit():
-    return UNIT
\ No newline at end of file
+    return UNIT
+
+  pub fn usage(self) ->  u256:
```
cburgdorf, line 16: The bug inspired me to test a few more use cases

crates/lowering/tests/snapshots/lowering__module_const.snap
```diff
@@ -3,16 +3,47 @@ source: crates/lowering/tests/lowering.rs
...
+        while 3 > 4:
+            pass
+
+        for x in list_expr_array_u256_2(3, 10):
+            pass
+
+        self.table[10] = 3
+        if THREE == TEN:
+            pass
+        else:
```
cburgdorf, line 34: It looks like from our AST there's no way to reconstruct the actual `elif` when we print the AST. Doesn't seem like a problem today but would become a problem when we want to have something like `fe fmt`.

sbillig:
:+1:

PR #569, opened 2021-10-12
Basic `unsafe` support

sbillig:
This adds basic support for unsafe functions and unsafe blocks. This doesn't add any built-in unsafe functions; someday soon there will be some fun unsafe fns in a `std::evm` module, but that'll come in a separate PR. So there's currently no point in using any of this.

I've decided to enforce the rule that contract functions can be either pub or unsafe, but not both. I don't have strong feelings about this, it just seemed reasonable.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/src/db/queries/functions.rs
+++ b/crates/analyzer/src/db/queries/functions.rs
@@ -9,6 +9,7 @@ use crate::traversal::types::type_desc;
 use fe_common::diagnostics::Label;
 use fe_parser::ast;
 use fe_parser::node::Node;
+use if_chain::if_chain;
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::rc::Rc;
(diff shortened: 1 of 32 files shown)


crates/analyzer/src/db/queries/functions.rs
```diff
@@ -25,6 +26,17 @@ pub fn function_signature(
     let mut scope = ItemScope::new(db, function.module(db));
     let contract = function.contract(db);
 
+    if_chain! {
```
cburgdorf, line 29: Sweet helper :)

crates/parser/src/grammar/functions.rs
```diff
@@ -9,14 +9,26 @@ use crate::node::{Node, Span};
...
+    let unsafe_qual = par.optional(TokenKind::Unsafe).map(|tok| tok.span);
+    if let Some(pub_) = par.optional(TokenKind::Pub) {
+        let unsafe_span =
+            unsafe_qual.expect("caller must verify that next token is `unsafe` or `fn`");
+
+        par.fancy_error(
+            "`pub` visibility modifier must come before `unsafe`",
+            vec![Label::primary(
+                unsafe_span + pub_.span,
+                "use `pub unsafe` here",
```
cburgdorf, line 23: >I've decided to enforce the rule that contract functions can be either pub or unsafe, but not both

As I understand `pub unsafe` is a compile time error so the error message of `"use pub unsafe here"` seems to be misleading.

cburgdorf: Ah, understood :+1:

PR #571, opened 2021-10-14
Change array type syntax to Array<u8, 100>

Maltby:
### What was wrong?

https://github.com/ethereum/fe/issues/273

### To-Do

- [ ] Clean up commit history


sbillig:
@Maltby looks great! The failing `features::test_method_return::case_029` test has revealed a small weakness in the old-array-syntax detecting code.

```
error: Outdated array syntax
  ┌─ features/return_from_memory_i8.fe:6:24
  │
6 │         let in_memory: i8[2] = [val, val]
  │                        ^^^^^
  │
  = Hint: Use `Array<i8, 2>`

error: type mismatch
  ┌─ features/return_from_memory_i8.fe:6:32
  │
6 │         let in_memory: i8[2] = [val, val]
  │                                ^^^^^^^^^^ this has type `Array<i8, 2>`; expected type `i8`

error: `i8` type is not subscriptable
   ┌─ features/return_from_memory_i8.fe:12:26
   │
12 │         let result: i8 = in_memory[0] + in_memory[1]
   │                          ^^^^^^^^^ unsubscriptable type
   │
   = Note: Only arrays and maps are subscriptable
```

Some background: if the parser emits an error but doesn't fail (which it should do if it's able to construct a valid ast despite the error), then the analyzer will run on the resulting ast. The analyzer is producing the type errors above.

So, there are two ways to solve this:
1) just make the old-array-syntax a fatal error (by returning `Err(ParseFailed)`), or, slightly fancier:
2) produce a correct `TypeDesc` generic array node for the old syntax. To do this successfully, you'd have to change the `if par.peek() == Some(BracketOpen)` back to `while par.peek() ...`, so nested old-syntax-arrays are handled correctly, eg `u8[4][4][4]`

I'd be fine with option 1. Sometime in the future, when the old array syntax is sufficiently old, we may just delete the code that tries to detect it anyway. (Same goes for the code that checks for the old `def` keyword and suggests `fn`). Option 2 would result in nicer user experience, because when the parser fatally fails, no further errors can be reported, because the whole thing just aborts. Totally up to you and your appetite for continuing to tweak this. Either way, would you a parser error test or two to test this new error handling code (`parser/tests/cases/errors.rs`)?

Regarding the "lint" failure, that's a weakness in our `docs/validate_doc_examples.py` script. The failure is happening in a doc example that should be skipped, but the skipping logic doesn't work if the example code is changed.

Change this line https://github.com/ethereum/fe/blob/d010ada567dfd60829791f986e4a463b7cbe4018/docs/validate_doc_examples.py#L26 to `clone_function.md_b5723756c9f8ecc9364a6a871a161dcb4607f2a3.fe` and that error will be cleared up. Someone can make this more robust in a future PR.

Maltby:
@sbillig I think I'll give option 2 a whirl, seems doable enough. I'll take care of that, add some tests, fix the lint failure, and let you know when that's all done. Thanks for all your help with this!

sbillig:
@Maltby Looks great! I pushed a couple commits (`--force`d) to your branch:
- made validate_doc_examples.py's skip list a bit smarter (turns out a few other files needed their hashes updated as well; the hashes have been replaced with simple per-file indexes)
- added a "newsfragment" file for the release notes, and fixed a minor build issue due to the latest change to the master branch

Thanks for doing all of this! I'll merge when the checks finish running.

Issue #524, opened 2021-08-26
Sign extend loaded signed ints 


g-r-a-n-t:
### What is wrong?

We are not extending the sign on loaded integers that are 128 bits or smaller.

e.g. a negative 8 bit integer currently gets loaded into a 256 bit stack value like so:

```
...00000000000   11010110
-------------- | --------
      ^               ^
 248 zeros       the 8 bit value (-42)
```

when it should be loaded like this

```
...11111111111   11010110
-------------- | --------
      ^               ^
 248 ones        the 8 bit value (-42)
```

### How can it be fixed

Call the `signextend` Yul function when loading signed integers.

PR #575, opened 2021-10-19
Signextend and friends

cburgdorf:
### What was wrong?

There were a few cases where we needed to call `adjust_numeric_size` to "clean" a `256` bit stack value that is holding a numeric value <= 128 bits (See #524)

### How was it fixed?

Changed `move_expression` to call `adjust_numeric_size` when loading integer values from memory or storage.

### DIFF
--- a/crates/test-files/fixtures/features/return_from_memory_i8.fe
+++ b/crates/test-files/fixtures/features/return_from_memory_i8.fe
@@ -1,0 +1,13 @@
+
+
+contract Foo:
+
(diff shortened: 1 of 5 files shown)


g-r-a-n-t:
>  I know that Solidity performs way more of these cleanup operations

I've noticed that too. It could be precautionary, but I'm not sure.

The PR looks good, but I'll spend some more time reviewing and approve it when I feel comfortable with it.

PR #577, opened 2021-10-21
Struct functions

sbillig:
This required a good bit'o refactoring of the way we handle `self` in the analyzer, because `self` structs can be used anywhere an expression can, which is not the case for the mystical `self` contract. It also "required" refactoring the `CallType` stuff. A better person than I could have done much of this refactoring in a separate PR, but they didn't.

Just for funsies, I also reworked the analysis test snapshot output, because the current snapshot printing was hard to use for debugging. Stuff is mostly printed in order now. Where possible, labels are combined into a single diagnostic to cut down on the noise and line count a bit (hence the -3k lines).

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/src/context.rs
+++ b/crates/analyzer/src/context.rs
@@ -1,7 +1,7 @@
-use crate::builtins::GlobalMethod;
+use crate::builtins::{ContractTypeMethod, GlobalFunction, ValueMethod};
 use crate::errors::{self, CannotMove, TypeError};
-use crate::namespace::items::{DiagnosticSink, EventId, FunctionId, Item};
-use crate::namespace::types::{FixedSize, Type};
+use crate::namespace::items::{Class, ContractId, DiagnosticSink, EventId, FunctionId, Item};
+use crate::namespace::types::{FixedSize, SelfDecl, Type};
 use crate::AnalyzerDb;
 use fe_common::diagnostics::Diagnostic;
 pub use fe_common::diagnostics::Label;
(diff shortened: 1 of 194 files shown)


crates/analyzer/src/context.rs
```diff
@@ -233,21 +245,77 @@ impl ExpressionAttributes {
...
 /// The type of a function call.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum CallType {
-    BuiltinFunction(GlobalMethod),
-    TypeConstructor { typ: Type },
-    SelfAttribute { func_name: String, self_span: Span },
+    BuiltinFunction(GlobalFunction),
+    BuiltinValueMethod(ValueMethod),
+
+    // create, create2 (will be methodss of the contract struct soon)
```
g-r-a-n-t, line 262:
```suggestion
    // create, create2 (will be methods of the contract struct soon)
```

crates/analyzer/src/namespace/scopes.rs
```diff
@@ -236,14 +244,14 @@ impl<'a, 'b> BlockScope<'a, 'b> {
...
-        )
-    }
+    // pub fn contract_name(&self) -> Option<String> {
+    //     Some(
+    //         self.root
+    //             .function
+    //             .contract(self.root.db)?
+    //             .name(self.root.db),
+    //     )
+    // }
```
g-r-a-n-t, line 255:
```suggestion
```

crates/parser/src/lexer/token.rs
```diff
@@ -209,107 +209,177 @@ pub enum TokenKind {
...
+    //         SlashEq => "`/=`",
+    //         PercentEq => "`%=`",
+    //         AmperEq => "`&=`",
+    //         PipeEq => "`|=`",
+    //         HatEq => "`^=`",
+    //         LtLtEq => "`<<=`",
+    //         GtGtEq => "`>>=`",
+    //         Arrow => "`->`",
+    //     }
+    // }
```
g-r-a-n-t, line 384: got some more commented code here

g-r-a-n-t (approved):
LGTM

cburgdorf:
Hot stuff :hot_pepper:

PR #580, opened 2021-10-27
Add test that sets and reads to and from storage

cburgdorf:
### What was wrong?

We haven't had a test that calls a method with a struct, sets it to storage and reads it again.

### How was it fixed?

Added such test.


sbillig (approved):
Sorry, I missed this one. Looks good to me.

PR #582, opened 2021-10-29
Don't coerce type of right hand side of binary operation

### LINKED ISSUE #550: failed to analyze lowered AST: The right hand side of the `>>` operation must be unsigned
Compiling this file:

```
contract o:
 pub fn o():
  [-1>>0]
```

```
thread 'main' panicked at 'failed to analyze lowered AST: [Diagnostic { severity: Error, message: "The right hand side of the `>>` operation must be unsigned", labels: [Label { style: Primary, span: Span { start: 32, end: 33 }, message: "this has signed type `i256`" }], notes: [] }]', crates/driver/src/lib.rs:67:51
stack backtrace:
   0:     0x555776868a90 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
   1:     0x555776868a90 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
   2:     0x555776868a90 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
   3:     0x555776868a90 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
   4:     0x5557768817df - core::fmt::write::h3b84512577ca38a8
   5:     0x5557768663b2 - std::io::Write::write_fmt::h465f8feea02e2aa1
   6:     0x55577686ac25 - std::sys_common::backtrace::_print::h525280ee0d29bdde
   7:     0x55577686ac25 - std::sys_common::backtrace::print::h1f0f5b9f3ef8fb78
   8:     0x55577686ac25 - std::panicking::default_hook::{{closure}}::ha5838f6faa4a5a8f
   9:     0x55577686a6d3 - std::panicking::default_hook::hfb9fe98acb0dcb3b
  10:     0x5557767b44f6 - fe_common::panic::DEFAULT_PANIC_HOOK::{{closure}}::{{closure}}::hfa0d7480b545d3da

Aborted
```

Using

```
[profile.dev]
debug = true
panic = "abort"
```

cburgdorf:
### What was wrong?

Fixes #550

The following code crashes:

```
let my_array: i256[1] = [-1 << 1] 
```

The interesting part is that this crashes **after lowering** with the following error:

>The right hand side of the `>>` operation must be unsigned"

Clearly, the right hand side **is** an unsigned integer so what's the problem here and why isn't it detected in the first analyzer pass?

The issue is that this gets lowered to something like this:

```
fn list_expr_array_i256_1(val0: i256) -> i256[1]:
    let generated_array: i256[1]
    generated_array[0] = val0
    return generated_array

contract Foo:
    pub fn o() -> ():
        list_expr_array_i256_1(-1 >> 0)
        return ()
```

Now that `-1 >> 0` is passed into the function `list_expr_array_i256_1` we try to coerce the types of both the left and the right hand side.

However, since only the left hand side determines the outcome of the type for the expression, I believe the fix is to not try to coerce the right hand side.

### How was it fixed?

Removed coercion for right hand side.

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -658,9 +658,18 @@ fn expr_bin_operation(
     expected_type: Option<Integer>,
 ) -> Result<ExpressionAttributes, FatalError> {
     if let fe::Expr::BinOperation { left, op, right } = &exp.kind {
-        let expected = expected_type.map(Type::int);
-        let left_attributes = value_expr(scope, left, expected.as_ref())?;
-        let right_attributes = value_expr(scope, right, expected.as_ref())?;
+        let (left_expected, right_expected) = match &op.kind {
+            // In shift operations, the right hand side may have a different type than the left hand side
+            // because the right hand side needs to be unsigned. The type of the entire expression is
+            // determined by the left hand side anyway so we don't try to coerce the right hand side in this case.
+            fe::BinOperator::LShift | fe::BinOperator::RShift => {
+                (expected_type.map(Type::int), None)
+            }
+            _ => (expected_type.map(Type::int), expected_type.map(Type::int)),
+        };
+
+        let left_attributes = value_expr(scope, left, left_expected.as_ref())?;
+        let right_attributes = value_expr(scope, right, right_expected.as_ref())?;
 
         let typ = match operations::bin(&left_attributes.typ, &op.kind, &right_attributes.typ) {
             Err(err) => {
--- a/crates/test-files/fixtures/crashes/agroce550.fe
+++ b/crates/test-files/fixtures/crashes/agroce550.fe
@@ -1,0 +1,3 @@
+contract Foo:
+  pub fn o():
+    [-1>>0]
--- a/crates/test-files/fixtures/features/return_bitwiseshl_i64_coerced.fe
+++ b/crates/test-files/fixtures/features/return_bitwiseshl_i64_coerced.fe
@@ -1,0 +1,4 @@
+contract Foo:
+    pub fn bar() -> i64:
+        let x: i64 = -1 << 0
+        return x
--- a/crates/tests/src/crashes.rs
+++ b/crates/tests/src/crashes.rs
@@ -16,3 +16,4 @@ macro_rules! test_file {
 }
 
 test_file! { agroce531 }
+test_file! { agroce550 }
--- a/crates/tests/src/features.rs
+++ b/crates/tests/src/features.rs
@@ -326,6 +326,7 @@ fn test_assert() {
     case("return_bitwiseshr_u256.fe", &[uint_token(212), uint_token(1)], uint_token(106)),
     case("return_bitwiseshr_i256.fe", &[int_token(212), uint_token(0)], int_token(212)),
     case("return_bitwiseshr_i256.fe", &[int_token(212), uint_token(1)], int_token(106)),
+    case("return_bitwiseshl_i64_coerced.fe", &[], int_token(-1)),
     // comparison operators
     case("return_eq_u256.fe", &[uint_token(1), uint_token(1)], bool_token(true)),
     case("return_eq_u256.fe", &[uint_token(1), uint_token(2)], bool_token(false)),
--- a/newsfragments/550.bugfix.md
+++ b/newsfragments/550.bugfix.md
@@ -1,0 +1,9 @@
+Fixed a rare compiler crash.
+
+Example:
+
+```
+let my_array: i256[1] = [-1 << 1] 
+```
+
+Previous to this fix, the given example would lead to an ICE.


cburgdorf:
Oops, need to look into failing tests

cburgdorf:
Ok, fixed the failing test. No idea if that is the right approach to fix it but it does pass the tests :sweat_smile:

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -623,7 +623,14 @@ fn expr_bin_operation(
...
+        let expected = match &op.kind {
+            // In shift operations, the right hand side may have a different type than the left hand side
+            // because the right hand side needs to be unsigned. The type of the entire expression is
+            // determined by the left hand side anyway so we don't try to coerce the right hand side in this case.
+            fe::BinOperator::LShift | fe::BinOperator::RShift => None,
+            _ => expected_type.map(Type::int),
+        };
+
         let left_attributes = value_expr(scope, left, expected.as_ref())?;
         let right_attributes = value_expr(scope, right, expected.as_ref())?;
```
sbillig, line 635: I agree that the RHS shouldn't be coerced, but shouldn't the LHS be? With this change:
```
error: type mismatch
  ┌─ /home/sean/code/eth/shift.fe:4:18
  │
4 │     let x: i64 = -1 << 0
  │                  ^^^^^^^ this has type `i256`; expected type `i64`
```

sbillig (changes requested):

crates/analyzer/src/traversal/expressions.rs
cburgdorf: Mmh, that's right (side note, it doesn't work on master either) but I wonder if it is safe to simply coerce the return type into `i64` or if an actual type cast would be needed.

When you do `let foo: i8 = -1` the generated yul is `let $foo := sub(0, 1)` but when you do `let foo: i8 = -1` the generated yul is  `let $foo := adjust_numeric_i8(sub(0, 1))`. *Maybe*[1] the call to `adjust_numeric_*` isn't needed when we are dealing with a literal because the range of the literal is checked at compile time but I would expect it is important for all other cases. So, I guess what we really want is `let x: i64 = i64(-1 << 0)` which **does work** and produces the following yul: `let $x := adjust_numeric_i64(shl(0, sub(0, 1)))`

I have a hard time to decide what to do here:

- should we just coerce the return type of the binary operation to the expected type?
  - if yes then in all cases or just within boundaries?
- should we continue to error and have the user add a manual type cast
- should we use lowering to add type casts (and maybe even replace the current coercing mechanic which sidesteps the numeric adjustment calls)

As you can see, I'm desperately in need of a professional to help me out here :joy_cat: 
Also paging in @g-r-a-n-t because he already spend some time thinking about the cleanup stuff, too. 

[1] I have a hard time telling if that is actually true. The more I think about it, the less I know.

sbillig: I think in the case of values that are known at compile-time, we should error if the result of the operation overflows the target type, and simply coerce into the target type if it doesn't. Ideally, we'd just emit the resulting value rather than a call to `shl`(. For now, "known at compile-time" will just be literals, but someday the compiler will be sufficiently smart to propagate constant values from previous lines of code and could detect overflow on something like `x << y` if x and y have known values at compile-time.)

If the values *aren't* known at compile time:

1) if the type of the LHS matches the target type: check for overflow at runtime and revert?
2) if the type of the LHS doesn't match the target type: require a manual type cast

For `let foo: i8 = i8(-1)`, I think if the value is a constant, this should be exactly the same as `let foo: i8 = -1`, and ideally the yul would be `let $foo := 0x11111...` (though maybe the yul optimizer would replace `sub(0, 1)` with a constant for us anyway). It seems to me that we don't need or want `adjust_numeric` calls when the value is known to fit within the target type, right? The value will already be sign-extended.

cburgdorf: Thanks for your answer. So, I agree with everything you wrote about dealing with values that are known at compile time, that eventually we should be able to map them into a literal.

So, let's focus on the case where they *aren't* known at compile time:

>if the type of the LHS matches the target type: check for overflow at runtime and revert

This is exactly what we do right now. Also note that the overflow check is currently always based on the type of the LHS.

>if the type of the LHS doesn't match the target type: require a manual type cast

Yes, that's why I was thinking that `let x: i64 = -1 << 0` is ok to error because `-1` is of type `i256` and so the user needs to manually cast it: `let x: i64 = i64(-1 << 0)`. Because when I was looking at ``let x: i64 = -1 << 0` what I was actually thinking about was ``let x: i64 = some_i256 << 0` and not a literal. But now I realized that the type coercion is only applied to a literal so that the following continues to be an error (rightfully so I think)

```
contract Foo:

  pub fn bar(z: i256):
    let x: i64 = z << 0
```

But this has revealed to me that I don't really understand how the type coercion based on passing `expected_type` works. How is it that the compiler doesn't try to coerce `z` into an `i64`? I tried to answer that myself but it isn't obvious to me how that works :grimacing: 

Apart from my open question, I believe the PR is now doing the right thing, doesn't it?

sbillig: > How is it that the coercion is only applied to literals and the compiler doesn't try to coerce z into an i64?

The expected type is only used when the type of an expression isn't already known (we consider `10` to be of some unknown integer type, and `[]` to be some unknown array type, etc). `expr_name` doesn't use the expected type because the types of all variables are explicitly declared; it returns the declared type and lets the caller decide if the types are compatible or not.

> Apart from my open question, I believe the PR is now doing the right thing, doesn't it?

Yep, looks good to me. As far as I can tell, we aren't yet evaluating operations and checking for overflow on the resulting value (for constants, at compile time I mean). Eg `let x: u8 = 1 << 9` could/should be an analyzer error, but isn't. A job for a future PR though.

sbillig (approved):

crates/analyzer/src/traversal/expressions.rs
cburgdorf: Ah, that's how the magic works. Thanks!

>checking for overflow

That's true but afaik Solidity isn't doing that either (Rust does though afaik). I *think* it's intentional and that one can look at it from two different angles.

Let's consider the value `1` as an `u8` which is represented as
the following 256 bit item on the EVM stack `00..|00000001|`. A left shift of `8` bits (`val << 8`) turns that into `00..01|00000000|`. We can say that this overflows but we could also simply consider it all zeros now since the `1` has moved out of the 8 bit range. We do cleanup the value so that it eventually turns into `00..00|00000000|` then.

sbillig: I think we should definitely call that an error at compile time, and perhaps suggest `let x: u8 = u8(1 << 8)` as a way to explicitly truncate and silence the error.

PR #583, opened 2021-10-29
Make differential proptests opt-in (via --include-ingored)

sbillig:
### What was wrong?

`cargo test --workspace --all-features` takes too long when running locally.

### How was it fixed?

I `#[ignore]`d the proptest-based tests, and added a `--include-ignored` flag to the gh workflow so they still run there.

They can be run separately with something like `cargo test --workspace --all-features -- --ignored`


sbillig:
Gonna unilaterally merge this :shipit:

PR #584, opened 2021-11-01
Re-enable differential testing as a cronjob

cburgdorf:
### What was wrong?

We recently added differential tests that aim to find bugs through fuzzing certain contracts with random inputs. These tests do not work well for the PR CI runs because:

- They may discover bugs which then causes a PR to fail for an unrelated reason
- They take a long time to run
- They should run as often as possible

### How was it fixed?

- Moved it into a cron job to run every 30 minutes. This was chosen as a sweet spot to run it as often as possible without clogging our CI queue (leaving room for PR jobs etc)
- When a job fails, the `differential.txt` is read and presented which give us a chance to add the seed to version control as a regression test.
- Ideally, we would like to create a github issue automatically but that's a job for another day because we will need to come up with something that first checks if the issue was already reported to not clog our issue log with duplicates

### DIFF
--- a/.github/workflows/expensive.yml
+++ b/.github/workflows/expensive.yml
@@ -1,0 +1,51 @@
+name: Expensive
+
+on:
+  schedule:
+    - cron:  '*/30 * * * *'
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test:
+    # Build & Test runs on all platforms
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache Rust dependencies
+        uses: actions/cache@v1.1.2
+        with:
+          # There's a problem with caching serde, hence we exclude it here
+          path: |
+            target
+            !target/**/*serde*
+          key: ${{ runner.OS }}-build-v2-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install Mac System dependencies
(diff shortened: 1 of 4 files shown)


.github/workflows/expensive.yml
```diff
@@ -0,0 +1,52 @@
...
+          override: true
+      - name: Build
+        run: cargo build --all-features --verbose
+      - name: Run expensive tests
+        id: expensive_tests
+        run: cargo test --workspace --features solc-backend --verbose -- --ignored
+      - name: Report
+        if: failure() && steps.expensive_tests.outcome == 'failure'
+        run: |
+          cat ./crates/tests/proptest-regressions/differential.txt
```
cburgdorf, line 51: I tested this on my local fork and it seems to do the job for now.

https://github.com/cburgdorf/fe/runs/4066729973?check_suite_focus=true

image

crates/test-files/fixtures/differential/math_i8.sol
```diff
@@ -16,10 +16,6 @@ contract Foo {
     return val1 / val2;
   }
 
-  function pow(int8 val1, uint8 val2) public pure returns (int8) {
-    return int8(uint8(val1) ** val2);
```
cburgdorf, line 20: I removed this because we can not really test it. Solidity doesn't allow the base to be negative whereas in Fe `-2 **3 = -8` is fine. I initially thought I could cast the problem away but that turned out to be silly and doesn't work. We still perform that test on the `math_u8` tests though.

sbillig (approved):
lgtm

PR #587, opened 2021-11-02
Attach file IDs to `Span`s

g-r-a-n-t:
### What was wrong?

https://github.com/ethereum/fe/blob/bfeffe82b4f8623330c1ebb1950ca981449c9e86/crates/common/src/diagnostics.rs#L8-L11

We also can't get a list of diagnostics with labels pointing to different source files.

### How was it fixed?

Refactored things.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/tests/snapshots/analysis__call_statement_with_args.snap
+++ b/crates/analyzer/tests/snapshots/analysis__call_statement_with_args.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/analyzer/tests/analysis.rs
-expression: "build_snapshot(\"features/call_statement_with_args.fe\", &src, module, &db)"
+expression: "build_snapshot(files, module, &db)"
 
 ---
 note: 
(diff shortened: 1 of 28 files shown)


crates/parser/src/ast.rs
```diff
@@ -18,7 +18,7 @@ pub enum ModuleStmt {
     Use(Node<Use>),
     TypeAlias(Node<TypeAlias>),
     Contract(Node<Contract>),
-    Constant(Node<ConstantDecl>),
+    Constant(Box<Node<ConstantDecl>>),
```
g-r-a-n-t, line 21: clippy was complaining about this variant being ~2x the size of the next largest

crates/analyzer/tests/snapshots/analysis__call_statement_with_args.snap
```diff
@@ -115,14 +115,11 @@ note:
...
-  │         ^^^^^^^^^^^ attributes hash: 8420354620523897173
+  │         ^^^^^^^^^^^ attributes hash: 8522453005002335674
   │
   = SelfAttribute {
         func_name: "assign",
-        self_span: Span {
-            start: 137,
-            end: 141,
-        },
+        self_span: 137..141,
```
sbillig, line 122: Yes! Why didn't we think of this earlier?

sbillig (approved):
Looks good to me.

crates/analyzer/tests/snapshots/analysis__call_statement_with_args.snap
g-r-a-n-t: :man_shrugging:

PR #589, opened 2021-11-03
Add out of bound check for array element access

cburgdorf:
### What was wrong?

Another thing found by the differential contract fuzzing is that we are missing *out of bounds* checks for when we access array elements.

### How was it fixed?

Not yet, just reporting.

### To-Do

- [ ] Clean up commit history


cburgdorf:
This landed with a different PR

Issue #593, opened 2021-11-14
ast in .json


hacker-DOM:
Hi folks!

We're building a static analyzer and would love to have support for Fe from Day 1!

I was wondering, by any chance are you planning to add support for json output of the AST?

sbillig:
Hi @hacker-DOM! That would be sweet. We certainly can add support for json; here's a quick-and-dirty PR that does just that: #594, and a sample of what the output currently looks like: https://gist.github.com/sbillig/82f0f4db6c58d1b6075cd1de5c0a3ce5 (for this contract: https://github.com/ethereum/fe/blob/b2e06d42713a3b50a62648ff0a5a104332881817/crates/test-files/fixtures/demos/erc20_token.fe)

Does that format suffice? Note that this is just a direct encoding of the internal AST structures, and is guaranteed to change as we continue adding features to the language, and refining the guts of the implementation. We could try to define some stable-ish encoding of the AST someday, but that wouldn't make it to the top of our priority list for quite a while, I reckon.

Does this static analyzer exist in a public repo somewhere? What language are you implementing it in?

hacker-DOM:
@sbillig Sorry for the long delay! In the end we decided to start with Solidity and add other langs later. Here's the repo https://github.com/Ackee-Blockchain/woke

PR #594, opened 2021-11-15
Output ast as json

sbillig:
### What was wrong?

#593 

### How was it fixed?

This is just a quick swap of the default `Debug` encoding of the ast with `serde_json::to_string_pretty`. Anyone have any feelings about what the ast output should look like? Should json be a separate cli option, with `Debug` (or RON?) the default? I have no preference. I also have no attachment to this PR (it was like 30 seconds of effort), so if someone wants to do other things with our cli options and outputs.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -486,6 +486,7 @@ dependencies = [
  "fe-common",
  "fe-driver",
  "fe-parser",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


cburgdorf:
>Should json be a separate cli option

That would be my preference I think

sbillig:
Defunct. Anyone interested is welcome and encouraged to add a separate cli arg for this.

Issue #488, opened 2021-07-21
Short circuited behavior of if-else, and, or


formalize:
### What is wrong?

The ternary if-else operator evaluates both branches and discards one of the two results, but it shouldn't evaluate the wrong branch at all. Moreover, the logical 'and' and 'or' operators are usually expected to behave like if-elses ("short circuited") rather than like bitwise operators.

For example, this Python program never calls f():

```python
def f() -> bool:
    return True

False and f()
True or f()
f() if False else False
False if True else f()
```

But this corresponding Fe program, as currently implemented, calls f() all 4 times:

```python
contract ShortCircuit:
    def f() -> bool:
        return true

    pub def main():
        false and self.f()
        true or self.f()
        self.f() if false else false
        false if true else self.f()
```

### How can it be fixed

Unfortunately there's no way to neatly compile Fe expressions into Yul expressions. Temporary variables and switch statements must be used, and those switch statements cannot be tucked inside a ternary() function as they are now.

cburgdorf:
Ran into this today again and added the *beta-required*  flag. Looking into it.

cburgdorf:
Ok, so far, we've always tried to stricly map expressions to expressions and statements to statements at both the lowering and the yulgen pass. It seems like we can not continue this pattern here because the only way to do so is to do what we do now and call `ternary(test, if_branch, else_branch)` but this causes all function arguments to be evaluated which prevents us from short circuting.

I propose that we tackle this at the lowering pass instead of the yulgen pass and that we turn a single ternary expression into a bigger code block as demonstrated with the following example:

Before lowering:

```
fn foo():
  some_fn(1 if some_condition else bar())
```

After lowering:

```
fn foo():
  let result: u256
  if some_condition:
    result = 1
  else:
    result = bar()
  some_fn(result)
```

This is more complex as it breaks with typical expression -> expression mapping convention but I don't see away around it.

sbillig:
Yeah, I think this is a good approach to take. In my opinion, we should, over time, move more and more of the complexity out of the yulgen phase into some earlier phase, either our current lowering phase or some imaginary future phase involving an intermediate representation that then requires only a very simple translation into yul.

PR #595, opened 2021-11-19
Lower ternary and boolean expressions

cburgdorf:
### What was wrong?

As described in #488 our current ternary expression doesn't short circuit but instead evaluates both branches and then discards one side of it.

### How was it fixed?

1. Create some helpers to manipulate parts of the AST better
2. Used these helpers to lower ternary expressions in a way that maps them to an if/else followed by a simple name expression

### DIFF
--- a/crates/lowering/src/ast_utils.rs
+++ b/crates/lowering/src/ast_utils.rs
@@ -1,0 +1,858 @@
+use fe_analyzer::namespace::types::FixedSize;
+use fe_parser::ast::{CallArg, Expr, FuncStmt, VarDeclTarget};
+use fe_parser::node::{Node, NodeId};
+
+use crate::names;
+use crate::utils::ZeroSpanNode;
+
+#[derive(Debug)]
+pub enum StmtOrExpr {
+    Stmt(Node<FuncStmt>),
+    Expr(Node<Expr>),
+}
+
+impl From<Node<FuncStmt>> for StmtOrExpr {
+    fn from(stmt: Node<FuncStmt>) -> Self {
+        StmtOrExpr::Stmt(stmt)
+    }
+}
+
+impl From<Node<Expr>> for StmtOrExpr {
+    fn from(expr: Node<Expr>) -> Self {
+        StmtOrExpr::Expr(expr)
+    }
+}
+
+impl StmtOrExpr {
+    pub fn as_stmt(&self) -> Node<FuncStmt> {
+        match self {
+            StmtOrExpr::Stmt(stmt) => stmt.clone(),
+            _ => panic!("not a statement"),
+        }
+    }
+
+    pub fn as_expr(&self) -> Node<Expr> {
+        match self {
+            StmtOrExpr::Expr(expr) => expr.clone(),
+            _ => panic!("not an expression"),
+        }
+    }
+}
+
+/// Recursively map the given `node` by applying the given `map_fn`
+pub fn map_ast_node<M>(node: StmtOrExpr, map_fn: &mut M) -> StmtOrExpr
+where
+    M: FnMut(StmtOrExpr) -> StmtOrExpr,
+{
+    match node {
+        StmtOrExpr::Stmt(stmt) => {
+            let node = match stmt.kind {
+                FuncStmt::Assert { test, msg } => FuncStmt::Assert {
+                    test: map_ast_node(test.into(), map_fn).as_expr(),
+                    msg: msg.map(|val| map_ast_node(val.into(), map_fn).as_expr()),
+                },
+                FuncStmt::Assign { target, value } => FuncStmt::Assign {
+                    target: map_ast_node(target.into(), map_fn).as_expr(),
+                    value: map_ast_node(value.into(), map_fn).as_expr(),
+                },
+                FuncStmt::AugAssign { target, op, value } => FuncStmt::AugAssign {
+                    target: map_ast_node(target.into(), map_fn).as_expr(),
+                    value: map_ast_node(value.into(), map_fn).as_expr(),
+                    op,
+                },
+                FuncStmt::Emit { name, args } => FuncStmt::Emit {
+                    name,
+                    args: map_call_args(args, map_fn),
+                },
+                FuncStmt::Expr { value } => FuncStmt::Expr {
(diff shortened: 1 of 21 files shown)


crates/lowering/src/ast_utils.rs
```diff
@@ -0,0 +1,344 @@
...
+            _ => panic!("meh"),
+        }
+    }
+}
+
+pub fn map_ast_node<S>(thingy: StmtOrExpr, map_fn: &mut S) -> StmtOrExpr
+where
+    S: FnMut(StmtOrExpr) -> StmtOrExpr,
+{
+    // TODO: This needs to walk the entire tree. Currently only walks enough to get my tests passing
```
cburgdorf, line 43: @sbillig warning, the code is very WIP and it doesn't yet walk the entire tree (only enough to get my tests passing).

So, the function that you asked about was `fn function_static_strings(db, function: FunctionId) -> Rc<IndexSet<String>>`. I think that you *could* build this on top of this if you call `map_ast_node` for each function statement and if the passed closure accumulates the `IndexSet<String>` as a side effect. That said, it's not particular elegant to have a mapping function leak out a side effect into a variable that's wrapped into the closure...I think it would actually be nicer to have another function `scan` that can traverse the tree and accumulate information into a single return value..but I haven't gotten that far yet :grimacing:

crates/lowering/src/mappers/functions.rs
```diff
@@ -101,6 +108,116 @@ pub fn func_def(context: &mut ModuleContext, function: FunctionId) -> Node<fe::F
...
+fn lower_ternary(context: &mut FnContext, statements: Vec<Node<FuncStmt>>) -> Vec<Node<FuncStmt>> {
+    let mut lowered_statements = vec![];
+    for stmt in statements.clone() {
+        // Not ideal, we are producing some side effects from the map.
+        // For each transformed ternary expression we want to collect
+        //  - the original ternary expression before it was lowered
+        //  - the name for the name expression that it was replaced with
+        //  - the statement that came right before it
+        let artifacts = RefCell::new(vec![]);
+        let last_stmt = RefCell::new(None);
```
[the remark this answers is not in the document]
cburgdorf: I've spend a bit trying to implement something like `scan_map` (`fold_map` ?) where we would not just map but also accumulate a custom value. It didn't went that well though...

cburgdorf:
I have another PR coming that lowers `and` and `or` since they suffer from the same problem. We might be able to just lower these to ternary expressions which then get lowered away. It's 99 % ready but apparently there's a problem with the unswap demo so I keep that out from this PR for now.

cburgdorf:
Ok, I just noticed this doesn't yet work well with nested ternary statements.

This lowering is wrong:

From:

```
    pub fn nested_ternary():
        let a: u256 = 10
        let b: u256 = 20
        let x: u256 = 10
        let y: u256 = 20
        return baz(a if true else (x if true else y))
```

To:

```
pub fn nested_ternary() -> ():
    let a: u256 = 10
    let b: u256 = 20
    let x: u256 = 10
    let y: u256 = 20
    let $ternary_result_0: u256
    if true:
        $ternary_result_0 = x
    else:
        $ternary_result_0 = y

    let $ternary_result_1: u256
    if true:
        $ternary_result_1 = a
    else:
        $ternary_result_1 = $ternary_result_0

     return baz($ternary_result_1)
```


But it actually should be:

```
pub fn nested_ternary() -> ():
    let a: u256 = 10
    let b: u256 = 20
    let x: u256 = 10
    let y: u256 = 20
    let $ternary_result_0: u256
    if true:
        $ternary_result_0 = a
    else:
        let $ternary_result_1: u256
        if true:
            $ternary_result_1 = x
        else:
            $ternary_result_1 = y

        $ternary_result_0 = $ternary_result_1

     return baz($ternary_result_0)
```

This is a bit more complicated because the nested ternary needs to be projected into the `else` branch of code that has yet to be generated.

I have ideas how to overcome that but I just wanted to make a quick brain dump of the bug and probably ask withdraw the requested review since the PR needs more work :grimacing:

cburgdorf:
Short progress update: I solved the problem of nested ternary expressions so they will now unroll into nested if/else statements accordingly. The code is still ugly which is why I haven't pushed it yet. The mechanism actually much more straight forward though which is good. Nothing to see yet, just a quick update that it is moving forward :D

cburgdorf:
@sbillig This can be reviewed now. It doesn't yet contain the `and` and `or` handling because I noticed I can not simply lower them into ternary expressions because I will face the same issue with nested `and` or `or` expressions. But we can handle them in the same way as the ternary expressions and I think most of the existing code can be re-used.

cburgdorf:
The PR now also contains the lowering of `and` and `or` expressions.

crates/lowering/src/ast_utils.rs
```diff
@@ -0,0 +1,969 @@
...
+                        .into_iter()
+                        .map(|val| map_ast_node(val.into(), map_fn).as_expr())
+                        .collect(),
+                },
+                Expr::UnaryOperation { op, operand } => Expr::UnaryOperation {
+                    op,
+                    operand: Box::new(map_ast_node((*operand).into(), map_fn).as_expr()),
+                },
+                // The following *could* be covered via catch all. However, that would turn into a footgun if we add
+                // more expressions in the future that need to be walked. It's better to not use a catch all here.
```
g-r-a-n-t, line 170: agreed. we can shorten this up with `|` separates, tho. still breaks the match statement when the enum is updated and saves us from having to repeat `expr.kind`

crates/lowering/tests/snapshots/lowering__and_or.snap
```diff
@@ -0,0 +1,182 @@
+---
+source: crates/lowering/tests/lowering.rs
```
g-r-a-n-t, line 2: general note on these snapshots: I think we should just include the unlowered code in the snapshot

g-r-a-n-t (approved):
looks good :+1:

crates/lowering/src/ast_utils.rs
cburgdorf: Ah, of course!

PR #596, opened 2021-11-24
Item dependency graphs and yulgen salsafication

### LINKED ISSUE #299: Can not use subsequently defined contract type in function signature
### What is wrong?

The following code doesn't compile because `Foo` can't use `Bar` in a function definition.

```
contract Foo:

    pub def yay() -> Bar:
        revert

contract Bar:

    pub def bar() -> u256:
        return 26
```

### How can it be fixed

We probably need a way to represent types that are only partially resolved (yet). This is because at the time where the signature of `yay` is analyzed, we haven't analyzed contract `Bar` yet.

PR #294 can be seen as a proof of concept but the way it represented partially resolved contracts wasn't as clean as we would like.

### LINKED ISSUE #305: Unused data objects included in initialization code.
### What is wrong?

From #304:

> The Yul optimizer does not currently remove these objects if they are unused. Meaning, created contracts will be included in the initialization code twice no matter what. This does not increase the contract size, but increases the initialization code size.

> The cost of this extra tx data is 16 for non-zero bytes and 4 for zero bytes (see appendix g).

> For example, say we write a contract that deploys an ERC-20 contract and has an __init__ that does not create an ERC-20 contract. A compiled Fe ERC-20 contract is about 2600 bytes. So, the extra cost would be slightly less than 42000, which is roughly equivalent to the cost of 2 sstore operations. This is a non-trivial expense and we should find a way to remove these unused data object. It might be the case that the Yul team plans on implementing this, so we don't have to worry about.

### How can it be fixed

It's possible Yul could optimize this stuff out, which would be ideal. So, asking the Yul team about this would be a good first step.

sbillig:
Functions, contracts, and struct items now have dependency graphs. A function depends on the functions it calls and any types that are present in the function signature or body (and on anything in the dependency graphs of those items). A type depends on the types of its fields (but not its functions; those are only included if they're called).

A contract runtime object is built by traversing the dependency graph of its (as yet imaginary) `__call__` function, which is the union of the dependency graphs of the contract's public functions (implemented in `contract_runtime_dependency_graph`).

The constructor object is built by traversing the graph of the contract's `__init__` fn, which means we no longer have to include the entire world in the constructor object "just in case".

The (directed) edges in the dependency graphs are labeled with `DepLocality::Local` or `::External`. An "external" edge between items means that the items will exist at different addresses when the code is deployed; eg calls to external contract functions result in "external" edges. (Note that this is orthogonal to whether the items are in the same "ingot"). When compiling a contract, we include all items reachable by traversing only "local" edges. "External" edges will be used for future 'unused item' warnings.

Inclusion of contract objects that are `create/create2`d is handled by checking the `CallType` of every call by every function in the dependency graph. This should probably be reworked; maybe an explicit "creates" relationship should be modeled in the dependency graph, or in a separate graph.

This is only a partial re-organisation of the yulgen code; some functions became salsa queries but others didn't, somewhat arbitrarily. More cleanup could be done, but not today.

closes #299
closes #305

### To-Do
- [x] Add entry to the [release notes]
- [x] Clean up commit history


g-r-a-n-t (approved):
LGTM :confetti_ball:

PR #609, opened 2021-12-14
More differential contract fuzzing

cburgdorf:
### What was wrong?

Need more differential contract fuzzing

### How was it fixed?

I had these piled up for a while but they were waiting on the fix for array index bounds checking which landed recently. 
Mostly just writing and reading things to and from memory and storage.
It also adds gas tracking (reporting overuse as an error). Our gas usage is astronomic but at least these checks should ensure that we notice if it gets worse. We should screw the bounds tighter when we optimize our code-.


g-r-a-n-t (approved):
:+1:

PR #610, opened 2021-12-14
Update solc to 0.8.10

### LINKED ISSUE #108: Update Solidity backend
### What is wrong?

We are using a version of solc-rust that compiles with Solidity version 0.6.2, we should instead be using Solidity version 0.7.4 or maybe even the 0.8.x preview.

See [this](https://github.com/axic/solc-rust/pull/2) for relevant info.

### How can it be fixed

Upgrade the Solidity submodule in solc-rust and and fix any issues that arise.

sbillig:
### What was wrong?

Closes #108

We were using solc 0.8.4, which doesn't support the (London) basefee opcode.

### How was it fixed?

Updated solc to 0.8.10

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,7 +1839,7 @@ dependencies = [
 [[package]]
 name = "solc"
 version = "0.1.0"
-source = "git+https://github.com/g-r-a-n-t/solc-rust#fcb910596c2b22d607970b2ba64229fe02e98e54"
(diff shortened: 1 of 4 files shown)


g-r-a-n-t (approved):

sbillig:
closing in favor of #613

PR #612, opened 2021-12-16
Fix feature flags in test & test-utils

WilfredTA:
### What was wrong?
Running only the `fe-compiler-tests` with `cargo test -p fe-compiler-tests --features solc-backend` fails because the `solc-backend` feature did not enable that feature in `test-utils`. Further, `test-utils` depends on `solc-backend` being enabled in `fe-driver` in the situation that `test-utils/solc-backend` is enabled. 

### How was it fixed?
Added relevant features in `tests` and `test-utils` crate to enable the same conditional features in appropriate dependencies.


sbillig (approved):
:+1:

PR #615, opened 2021-12-17
Rust 2021, update (most) dependencies

sbillig:
### What was wrong?

I didn't feel like doing anything challenging.

### How was it fixed?
Did some mindless chores. Find and replace, deleted now-unnecessary `TryInto` `use`s, etc, ran `cargo outdated` a couple times. I didn't update a few things that would have required code changes that I didn't want to make.

@g-r-a-n-t Note the `include_dir` change in test-files. You might need to do the same for the new library crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,15 +13,6 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.11.0"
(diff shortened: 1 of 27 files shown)


g-r-a-n-t (approved):
:partying_face:

PR #616, opened 2021-12-17
CI speedup

sbillig:
### What was wrong?

Github CI builds are slow.

### How was it fixed?

The rust-cache action is doing most of the work here, but I made a few other tweaks too. The test jobs now take 2-3 minutes instead of 20-25. Installing boost is about a third of this. I had been using an apt cache action, but then the coverage build failed mysteriously on a failure to find boost, so I scrapped that. I'll either revisit that someday, or maybe just install the few boost libs that solc actually uses.

The coverage job is still slow (~20min now instead of 35). I'm tempted to remove it from the PR workflow so that finishes quickly, and just let it run on merges to master or something. We aren't using coverage stats to judge the test cases of PRs anyway.

Oh, I bumped the proptest job to run every 60 minutes, just to reduce our resource consumption a bit. I'm not attached to this, it just felt like a reasonable thing to do. They are now running much more quickly overall (~6min iirc), thanks to the build speedup. We should, however, speed up the proptest code. Grant demonstrated that 95% of the time is spent building/deploying the contracts over and over again, so a 20x perf improvement should be possible with some refactoring of the test utils.

### DIFF
--- a/.github/workflows/expensive.yml
+++ b/.github/workflows/expensive.yml
@@ -2,7 +2,7 @@ name: Expensive
 
 on:
   schedule:
-    - cron:  '*/30 * * * *'
+    - cron:  '*/60 * * * *'
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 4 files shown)


cburgdorf (approved):
Great speedup! :racehorse:

PR #617, opened 2021-12-23
Module Item Visibility

WilfredTA:
# What was wrong?

`pub` qualifiers for modules could not be parsed for events, structs, contracts, and type aliases

# How was it fixed?

Added parsing logic to handle when any of the above-listed types were preceded by `pub` keyword.
# To-Do

- [x] OPTIONAL: Update Spec if applicable
- [ ] Add entry to the release notes (may forgo for trivial changes)
- [ ]  Clean up commit history


g-r-a-n-t (approved):
:+1: LGTM

Issue #236, opened 2021-02-05
Full support for pub qualifier on structs


cburgdorf:
### What is wrong?

When #203 lands, we'll have basic support for structs.

```
struct House:
    price: u256
    size: u256
    pub vacant: bool
```

Notice that `vacant` uses `pub` to declare that the field is public. We know that we want to allow `pub` qualifier which will become most relevant for when structs are allowed to have methods. This is why the parser does already recognize it. However, it doesn't have any effect yet.

### How can it be fixed

1. Define how visibility of struct fields should work exactly
2. Make the relevant changes to analyzer and mapper pass

PR #620, opened 2022-01-03
private struct fields

cburgdorf:
### What was wrong?

As described in #236 we need support for private fields in structs.

### How was it fixed?

Public fields now need to be declared with the `pub` modifier, otherwise they default to private fields.
If a struct contains private fields it can not be constructed directly except from within the
struct itself. The recommended way is to implement a method `new(...)` as demonstrated in the
following example.

```
struct House:
    pub price: u256
    pub size: u256
    vacant: bool

    pub fn new(price: u256, size: u256) -> House
      return House(price=price, size=size, vacant=true)

contract Manager:

  house: House

  pub fn create_house(price: u256, size: u256):
    self.house = House::new(price, size)
    let can_access_price: u256 = self.house.price
    # can not access `self.house.vacant` because the field is private
```

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -619,9 +619,20 @@ fn expr_attribute(
         // If the value is a struct, we return the type of the struct field. The location stays the
         // same and can be memory or storage.
         Type::Struct(struct_) => {
-            if let Some(field) = struct_.id.field(scope.db(), &field.kind) {
+            if let Some(struct_field) = struct_.id.field(scope.db(), &field.kind) {
+                if !scope.root_item().is_struct(&struct_.id) && !struct_field.is_public(scope.db())
+                {
+                    scope.fancy_error(
+                        &format!(
+                            "Can not access private field `{}` on struct `{}`",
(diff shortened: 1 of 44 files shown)


crates/analyzer/src/traversal/expressions.rs
```diff
@@ -619,11 +619,23 @@ fn expr_attribute(
...
+                if !scope.root_item().is_struct(&struct_.id) && !struct_field.is_public(scope.db())
+                {
+                    Err(FatalError::new(scope.fancy_error(
+                        &format!(
+                            "Can not access private field `{}` on struct `{}`",
+                            &field.kind, struct_.name
+                        ),
+                        vec![Label::primary(field.span, "private field")],
+                        vec![],
+                    )))
```
sbillig, line 632: This doesn't have to be fatal.

sbillig (approved):
Looks great! :cowboy_hat_face:

crates/analyzer/src/traversal/expressions.rs
cburgdorf: eagle eyes! :eagle:

PR #622, opened 2022-01-04
contract `__call__` function  

### LINKED ISSUE #559: Add a special `__call__` function
Users can define a contract function named `__init__` that will execute at initialization time.

Continuing with this pattern, we might consider adding a function named `__call__` that runs at *call time*. By adding such a function, the language would be more flexible with regard to contract interfaces.

e.g.

```python
contract Foo:
  # the __call__ function is defined with full access here. not sure if that should be required..?
  pub fn __call__(mut self, ctx: mut Context):
    # grab the calldata from the Context object
    let calldata = ctx.calldata()

    # the function is selected using the first byte in calldata
    if calldata[0] == 0:
      # pass the second byte into foo
      self.foo(calldata[1])
      return
    else if calldata[0] == 1:
      # pass the second and third byte into bar
      self.bar(calldata[1], calldata[2])
      return
   
    revert "invalid calldata"

  fn foo(self, my_byte: u8):
    # do stuff

  fn bar(self, my_byte: u8, my_extra_byte: u8):
    # do stuff
```

In the example above, we have defined a super simple and efficient contract interface. Writing a custom interface such as this one might be useful for people looking to reduce fees. The other benefit is that we could move the Solidity ABI decoding/encoding implementation into Fe-lang as opposed to having that code generated in the Yul gen phase.

In the future, it might be possible to change the interface of a contract using a procedural macro that writes a `__call__` function for you.

e.g.

```python
#[interface(SomeAbi)]
contract Foo:
   ...
```

One thing to note is that the `__init__` function takes ABI encoded parameters, which does not align well with this proposal.

g-r-a-n-t:
### What was wrong?

closes #559 

### How was it fixed?

:keyboard: 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/tests/errors.rs
+++ b/crates/analyzer/tests/errors.rs
@@ -297,6 +297,11 @@ test_file! { init_wrong_return_type }
 test_file! { init_duplicate_def }
 test_file! { init_call_on_self }
 test_file! { init_call_on_external_contract }
+test_file! { call_wrong_return_type }
+test_file! { call_duplicate_def }
+test_file! { call_call_on_self }
+test_file! { call_call_on_external_contract }
+test_file! { call_with_pub_fns }
 test_file! { abi_encode_u256 }
 test_file! { abi_encode_from_storage }
 test_file! { assert_sto_msg_no_copy }
--- a/crates/tests/src/features.rs
+++ b/crates/tests/src/features.rs
@@ -1771,3 +1771,18 @@ fn intrinsics() {
         );
     });
 }
+
+#[test]
+fn call_fn() {
+    with_executor(&|mut executor| {
+        let harness = deploy_contract(&mut executor, "call_fn.fe", "Foo", &[]);
+
+        let mut calldata = vec![0; 32];
+
+        calldata[31] = 1;
+        harness.test_call_reverts(&mut executor, calldata.clone(), &[]);
+
+        calldata[31] = 0;
+        harness.test_call_returns(&mut executor, calldata, &[])
+    });
+}
--- a/crates/yulgen/src/db/queries/contracts.rs
+++ b/crates/yulgen/src/db/queries/contracts.rs
@@ -1,5 +1,7 @@
 use crate::constructor;
+use crate::context::FnContext;
 use crate::db::YulgenDb;
+use crate::mappers::functions::multiple_func_stmt;
 use crate::runtime::{abi_dispatcher, functions};
 use crate::types::{AbiDecodeLocation, AsAbiType};
 use fe_analyzer::builtins::ValueMethod;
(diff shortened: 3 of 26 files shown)


crates/analyzer/tests/snapshots/errors__init_wrong_return_type.snap
```diff
@@ -10,7 +10,7 @@ error: `__init__` function has incorrect return type
   │                          ^^^ return type should be `()`
   │
   = Hint: Remove the return type specification.
-  = Example: `pub fn __init__():`
+  = Example: `pub fn __init__(self):`
```
g-r-a-n-t, line 13: remove self

crates/tests/src/features.rs
```diff
@@ -1771,3 +1771,16 @@ fn intrinsics() {
...
+#[test]
+fn call_fn() {
+    with_executor(&|mut executor| {
+        let harness = deploy_contract(&mut executor, "call_fn.fe", "Foo", &[]);
+
+        let mut calldata = vec![0; 32];
+        calldata[31] = 1;
+
+        harness.test_call_reverts(&mut executor, calldata, &[]);
+        harness.test_call_reverts(&mut executor, calldata, &[])
```
g-r-a-n-t, line 1784: fix this

crates/yulgen/src/db/queries/contracts.rs
```diff
@@ -21,7 +23,23 @@ pub fn contract_object(db: &dyn YulgenDb, contract: ContractId) -> yul::Object {
             Item::Type(TypeDef::Contract(contract)),
         );
 
-        functions.extend(db.contract_abi_dispatcher(contract));
+        let call_fn_ident = identifier! { ("$$__call__") };
```
g-r-a-n-t, line 26: add comment explaining that `__call__` will eventually be expected at yulgen

crates/analyzer/tests/errors.rs
```diff
@@ -297,6 +297,10 @@ test_file! { init_wrong_return_type }
 test_file! { init_duplicate_def }
 test_file! { init_call_on_self }
 test_file! { init_call_on_external_contract }
+test_file! { call_wrong_return_type }
```
g-r-a-n-t, line 300: check that there are no pub fns in a contract when `__call__` is defined

crates/yulgen/src/db/queries/contracts.rs
```diff
@@ -21,7 +23,31 @@ pub fn contract_object(db: &dyn YulgenDb, contract: ContractId) -> yul::Object {
             Item::Type(TypeDef::Contract(contract)),
         );
 
-        functions.extend(db.contract_abi_dispatcher(contract));
+        // This can all be replaced with a call to the contract's `__call__` function once the
+        // dispatching code has been moved into lowering.
```
cburgdorf, line 27: Realy cool, looking forward to it!

cburgdorf (approved):
Looks great!

WilfredTA:
I like this approach. It is a natural way to accommodate the ABI compatibility concerns and is foundational enough that we can always build higher level language features that use this internally.

LGTM

WilfredTA (approved):

PR #625, opened 2022-01-06
Properly resolve generic type desc

cburgdorf:
### What was wrong?

Empty list expressions lead to an ICE

### How was it fixed?

Lowering code produced wrong `TypeDesc` for generic arrays.


sbillig (approved):
:+1:

PR #628, opened 2022-01-07
file salsa db and other hot refactorings

sbillig:
- `FileStore` has been replaced by (salsa) `SourceDb`. This is in the common crate for now, but should probably be moved to a new crate (and the diagnostics stuff should probably be generic, so it doesn't have to depend on SourceDb)
- a `SourceFileId` is now the salsa-interned path
- file content is a (salsa) input to the `SourceDb`. This is mostly so that the future language server can update file content as the user types, which will trigger the rerunning of any necessary downstream queries
- ingot module list and ingot dependencies are also a salsa inputs, for the language server (user adds/removes a file or dependency, analysis is rerun). Ideally ingot *file* list would be the input rather than modules, but that makes supporting analysis of lowered ingots tricky.
- standalone modules (eg a module compiled with `fe fee.fe`) now have a fake ingot parent. `Ingot` has an `IngotMode` (Lib, Main, StandaloneModule), which is used to find the root module and disallow `ingot::` in the case of standalone modules.
- `parse_module` now always returns an `ast::Module`, and thus a `Module` item will always exist for a source file, even if it contains fatal parse errors. If the parsing 
fails, the body will end with a `ModuleStmt::ParseError` node. This is a modest first step toward making the parser return as complete an ast as possible, so that we can analyze whatever code doesn't have errors, and the language server can be useful while the user types
- `ModuleId::ast(db)` is now a query that parses the module's file on demand, rather than the ast being interned into salsa. This makes handling parse diagnostics cleaner, and removes the up-front parsing of every module at ingot creation time.

Many of the changes in this pr feel like fairly pointless rewriting to get everything to work in this new structure; a lot will need to be changed again as the ingot system matures.

### To-Do

- [x] Clean up commit history


sbillig:
This may have introduced a bug in the lowering of const values in revert statements; need to investigate before it's merged.

g-r-a-n-t (approved):
LGTM

PR #629, opened 2022-01-07
std::evm lib, port send_value to fe

sbillig:
Depends on #628

This adds simple wrappers around the intrinsics to `std::evm`, deletes the global `balance`, `balance_of`, and `send_value` (which is now in `std`)

The functions that access chian/msg/tx context are not currently marked as unsafe, but should be when the Context object is ready.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/library/std/src/evm.fe
+++ b/crates/library/std/src/evm.fe
@@ -1,0 +1,293 @@
+# Basic context accessor functions.
+# These should be unsafe when the Context object is introduced.
+
+pub fn chain_id() -> u256:
+  unsafe:
+    return __chainid()
+
+pub fn base_fee() -> u256:
+  unsafe:
+    return __basefee()
+
+pub fn origin() -> address:
+  unsafe:
+    return address(__origin())
+
+pub fn gas_price() -> u256:
+  unsafe:
+    return __gasprice()
+
+pub fn block_hash(b: u256) -> u256:
+  unsafe:
+    return __blockhash(b)
+
+pub fn coin_base() -> address:
+  unsafe:
+    return address(__coinbase())
+
+pub fn timestamp() -> u256:
+  unsafe:
+    return __timestamp()
+
+pub fn block_number() -> u256:
+  unsafe:
+    return __number()
+
+pub fn difficulty() -> u256:
+  unsafe:
+    return __difficulty()
+
+pub fn gas_limit() -> u256:
+  unsafe:
+    return __gaslimit()
+
+pub fn gas() -> u256:
+  unsafe:
+    return __gas()
+
+pub fn self_address() -> address:
+  unsafe:
+    return address(__address())
+
+pub fn balance_of(addr: address) -> u256:
+  unsafe:
+    return __balance(u256(addr))
+
+pub fn balance() -> u256:
+  unsafe:
+    return __selfbalance()
+
+pub fn caller() -> address:
+  unsafe:
+    return address(__caller())
+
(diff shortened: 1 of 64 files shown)


crates/library/std/src/evm.fe
```diff
@@ -0,0 +1,293 @@
...
+  unsafe:
+    return address(__caller())
+
+pub fn call_value() -> u256:
+  unsafe:
+    return __callvalue()
+
+
+# Overflowing math ops. Should these be unsafe or named
+# `overflowing_add`, etc?
```
cburgdorf, line 70: Without taking very long to think about it I'm in favor of making these `unsafe`

crates/library/std/src/evm.fe
```diff
@@ -0,0 +1,293 @@
...
+
+
+# Evm state access and control
+pub unsafe fn return_mem(location: u256, len: u256):
+    return __return(location, len)
+
+pub unsafe fn revert_mem(location: u256, len: u256):
+    return __revert(location, len)
+
+pub unsafe fn self_destruct(addr: address):
```
cburgdorf, line 185: The underscore between `self` and `destruct` stands out odd with me.

crates/library/std/src/evm.fe
```diff
@@ -0,0 +1,293 @@
...
+pub unsafe fn invalid():
+    return __invalid()
+
+pub unsafe fn stop():
+    return __stop()
+
+pub unsafe fn pc() -> u256:
+    return __pc()
+
+# TODO: dunno if we should enable this
```
cburgdorf, line 200: What's the counter argument to not enable it?

cburgdorf (approved):
Hot stuff :fire:

crates/library/std/src/evm.fe
sbillig: Ok, I'll remove.

crates/library/std/src/evm.fe
sbillig: I'm not sure that I understand the semantics of yul's `pop`. It seems to just ignore the return value of an expression. Fe (and rust, without a `must_use` annotation) will happily ignore a return value that you don't use, but in yul that's an error. So fe's `call(...)` becomes `pop(call(...))` in yul.

Otherwise, it doesn't seem to do anything. Eg.
```
        let x := 5
        pop(x)
        mstore(0, x)
        return(0, 32)
```
becomes:
```
  0x05
  0x00
  mstore
  0x20
  0x00
  return
```
The `pop(x)` is apparently just 'ignoring the "return value" of the expression `x`', and thus a no-op.

So as far as I can tell, there's no point in including it.

If yul exposed raw pushN/pop/dupN etc stack manipulation operations, I'd certainly be in favor of including those in std::evm, but alas. It's a bummer that there's apparently no way we can allow advanced unsafe fe code to mess around with the stack.

crates/library/std/src/evm.fe
sbillig: `unsafe` feels a little strong to me for arithmetic; like perhaps shouldn't "weaken" it by including wrapping ops in that category. I ended up just leaving it as-is; anyone with strong feelings is welcome to change these of course. Btw, we had a brief discussion on discord while you weren't around to stop us about making the operations on the base types wrap without revert (gasp!), and introducing `SafeU32` etc types that `impl Add` with revert-on-overflow behavior. A discussion to rehash at some point.

PR #633, opened 2022-01-15
signext unpacked ints

g-r-a-n-t:
### What was wrong?

We weren't `signext`ing int array elements when encoding (see #581).

### How was it fixed?

Added a `signed` parameter to the `abi_unpack` Yul method. If this value is set true, each element will be `signext`ed when unpacked.

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
Looks good!

Issue #343, opened 2021-03-27
Support non base type items in structs


cburgdorf:
### What is wrong?

Currently structs are limited to just base types. E.g the following is not possible yet.

```
struct TodoItem:
    content: String<100>
```

### How can it be fixed

:hocho:

cburgdorf:
This happened, I'll open up follow up issues for the remaining parts.

PR #634, opened 2022-01-17
Support non-base type fields in structs

cburgdorf:
### What was wrong?

As described in #343 structs can currently only use fields with base types. That limitation should be lifted.

### How was it fixed?

Structs can now use any fixed size type (except contract types for now) as field types. However, support is currently limited in two major ways:
  - Structs with complex fields can not be returned from public functions
  - Structs with complex fields can not be stored in storage

That functionality is expected to arrive with follow up PRs.

I should also say that the implementation based on `get_x_ptr` / `get_x_ptr_raw` feels a bit dirty (especially the part where we derive the `_raw` API from the non `raw` one). But this seems like something that would naturally evolve in another direction if we add explicit reference semantics so it may be ok for now.

### DIFF
--- a/crates/analyzer/src/db/queries/contracts.rs
+++ b/crates/analyzer/src/db/queries/contracts.rs
@@ -346,6 +346,13 @@ pub fn contract_field_type(
         scope.not_yet_implemented("contract field initial value assignment", value_node.span);
     }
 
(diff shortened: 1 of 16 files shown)


sbillig (approved):
Sweet, looks good :1st_place_medal:

PR #640, opened 2022-01-25
Refactor complex struct handling

cburgdorf:
### What was wrong?

When I wrote the handling of complex structs I changed the getter handling to create two different kind of getters. One with auto-deref and one without. I think it is better to keep the deref step out of the getters but for some reason the solution to it didn't occur to me.

### How was it fixed?

We add derefs (`mload`) to the get_attribute calls and remove them from those where the attribute is used assign a reference type in a struct field.


cburgdorf:
This landed as part of #642

PR #643, opened 2022-01-29
Handle parser panic on unary plus

Maltby:
### What was wrong?
Parser panicked on unary plus https://github.com/ethereum/fe/issues/443

### How was it fixed?
Added error message when unary plus is hit.

### To-Do

- [ ] Clean up commit history


g-r-a-n-t:
:100:

PR #644, opened 2022-01-29
[library] Change cargo rerun condition

Y-Nak:
### What was wrong?
All crates in Fe were recompiled every time due to the wrong path in `library/bulid.rs`.

### How was it fixed?
Just changed cargo rerun condition in `library/bulid.rs`

- [x] Clean up commit history


Y-Nak:
I found rust edition of `library` is still 2018, so updated it in this PR.

sbillig (approved):
:+1:

Issue #641, opened 2022-01-27
A wrong path to a non-existence src file causes ICE


Y-Nak:
### What is wrong?

Please include information like:

* full output of the error you received
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { depth: 0, inner: Io { path: Some("wrong_path"), err: Os { code: 2, kind: NotFound, message: "No such file or directory" } } }', crates/fe/src/main.rs:160:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8cdb3cd94efece1e17cbd8f6edb1dc1a482779a0/library/std/src/panicking.rs:577:5
   1: core::panicking::panic_fmt
             at /rustc/8cdb3cd94efece1e17cbd8f6edb1dc1a482779a0/library/core/src/panicking.rs:135:14
   2: core::result::unwrap_failed
             at /rustc/8cdb3cd94efece1e17cbd8f6edb1dc1a482779a0/library/core/src/result.rs:1737:5
   3: core::result::Result<T,E>::unwrap
   4: fe::build_ingot_filestore_for_dir
   5: fe::main
   6: core::ops::function::FnOnce::call_once
```

* what command you ran
`RUST_BACKTRACE=1 cargo run --features solc-backend wrong_path`

Narasimha1997:
I can work on fixing this.

Narasimha1997:
#645

Y-Nak:
This was fixed in #645, thanks @Narasimha1997!

PR #645, opened 2022-01-30
Fix: Prevent solc-backend panic when non-existent path is passed to it as argument. (Fixes issue #641)

Narasimha1997:
### What was wrong?
As stated in #641 , the `solc-backend` panics when a non-existent path is passed to it as an argument because `unwrap` was called directly on `entry` assuming that it is always `Ok` and not an `Err`. (in other words, directory errors were unhandled)

### How was it fixed?
Just added a checker before calling `unwrap`, this would exit the process with code `1` and avoids panic.
```rust
for entry in walker {
        if entry.is_err() {
            eprintln!("Error: {}", entry.unwrap_err());
            std::process::exit(1)
        }
        let entry = entry.unwrap();
      .......
```
Every item returned by the iterator is of type `Result<walkdir::DirEntry, walkdir::Error>`, so `entry.is_err()` check  needs to be called for each item. An alternative solution would be to check if the path exists before the for loop, but that would not guarantee that every item in the iteration is always of type `Ok(walkdir::DirEntry)`, because an `Err` can happen for so many reasons.  Also the error message obtained from `entry.unwrap_err()` is self-descriptive by itself, so no need to print anything extra from our side: 
```
Error: IO error for operation on wrong_path: No such file or directory (os error 2)
```

### DIFF
--- a/crates/fe/src/main.rs
+++ b/crates/fe/src/main.rs
@@ -157,6 +157,10 @@ fn build_ingot_filestore_for_dir(path: &str) -> FileStore {
     let mut files = FileStore::new();
 
     for entry in walker {
+        if entry.is_err() {
+            eprintln!("Error: {}", entry.unwrap_err());
+            std::process::exit(1)
+        }
         let entry = entry.unwrap();
         let file_path = &entry.path().to_string_lossy().to_string();
 


g-r-a-n-t:
looks good! appreciate the contribution

PR #649, opened 2022-02-01
Constant evaluation

Y-Nak:
This PR adds some features and fixes some ICE related to constant evaluation.

### 1. Support local constant

Example:
```python
contract Foo:
    pub fn bar():
        const LOCAL_CONST: i32 = 1
```

### 2. Support constant expression

Example:
```python
const GLOBAL: i32 = 8

contract Foo:
    pub fn bar():
        const LOCAL: i32 = GLOBAL * 8
```

### 3. Support constant generics expression

Example:
```python
const GLOBAL: u256= 8
const USE_GLOBAL: bool = false
type MY_ARRAY = Array<i32, { GLOBAL / 4 }>

contract Foo:
    pub fn bar():
        let my_array: Array<i32, { GLOBAL if USE_GLOBAL else 4 }>
```

### Fix ICE when constant type is a mismatch

Example:
```python
const GLOBAL: i32 = "FOO"

contract Foo:
    pub fn bar():
        let FOO: i32 = GLOBAL
```

### Fix ICE when assigning value to constant twice

Example:
```python
const BAR: i32 = 1

contract FOO:
    pub fn bar():
        BAR = 10
```

### TODO
I opened a follow-up issue [here](#652)


g-r-a-n-t (approved):
great work!

Y-Nak:
Fixed an invalid example in news fragments.

newsfragments/649.feature.md
```diff
@@ -0,0 +1,58 @@
...
+const GLOBAL: u256= 8
+const USE_GLOBAL: bool = false
+type MY_ARRAY = Array<i32, { GLOBAL / 4 }>
+
+contract Foo:
+    pub fn bar():
+        let my_array: Array<i32, { GLOBAL if USE_GLOBAL else 4 }>
+```
+
+# Bug fixes
```
cburgdorf, line 36: Just a nitpick but next time it would be better to create another file `649.bugfix.md` because our changelog generator picks up these file name keywords to group the items accordingly. No big deal though, we can always apply changes to the changelog by hand.

Y-Nak: Oh, Ok. I understand.

PR #659, opened 2022-02-08
Fix expensive tests

Y-Nak:
### What was wrong?
I broke expensive tests due to inappropriate annotations on example code blocks. Sorry.

### How was it fixed?
Changed annotations from `ignore` to `fe`.


sbillig (approved):
👍  No problem

PR #638, opened 2022-01-24
`std::context::Context` and usage checks

g-r-a-n-t:
### What was wrong?

see #558 

### How was it fixed?

Created an `std::context` module and added a `Context` struct. Usage of this struct is guided by the compiler. To see how usage is guided, I recommend looking at the `ctx_` compiler error snapshots.

There are a few things missing in this PR:

- Function access in the Json ABI should take `ctx` into account. This would be a good first issue, so I'll write it up and see if someone picks it up.
- There should be a `std::prelude` module that reexports items like `Context`. I tried to implement this, but ran into some cycle errors and decided to set it aside for now. For now, `Context` must be imported by the user. I'll write an issue for this also.

Note: `ctx` can only be used safely in contract functions. This might be too strict..

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] OPTIONAL: Update [Spec](https://github.com/ethereum/fe/blob/master/docs/src/spec/index.md) if applicable
- [x] Add entry to the [release notes](https://github.com/ethereum/fe/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [x] Clean up commit history

PR #661, opened 2022-02-09
Implement `mir`

Y-Nak:
# MIR
## MIR structure
Fe's MIR has a similar structure as rustc's [MIR](https://blog.rust-lang.org/2016/04/19/MIR.html).
MIR is a low-level layer that represents the information required to generate an EVM (or in the future, WASM) in as small a structure as possible, and enables various optimizations and semantic analysis that are difficult to perform on AST.

A MIR function is represented as a series of [Basic Block](https://en.wikipedia.org/wiki/Basic_block) that contains a series of instructions.
Now, let's see how the following `add` function is represented in MIR.
```python
fe add(x: i32, y:i32) -> i32:
    return x + y
````

add
Actually, this example is not so interesting, but there are a few things worth noting.

First, all local variables and arguments in the function body are represented as a number prefixed with `_`.  Below the function signature, you can see a map from these lowered variables to user-defined variables.

Second, expressions are lowered into [Three-address code](https://en.wikipedia.org/wiki/Three-address_code#:~:text=In%20computer%20science%2C%20three%2Daddress,assignment%20and%20a%20binary%20operator.) and stored in a temporary local which is introduced while lowering.

## Control flow
MIR doesn't have a high-level control flow, so all control flow in AST is lowered into low-level `jump` and `branch` instruction.

#### If statement
```python
pub fn abs(x:i32) -> i32:
    if x > 0:   
        return x
    else:
        return -x
```

abs

```python
pub fn abs2(x:i32) -> i32:
    if x < 0:   
        x = -x
    return x
```
abs2

#### for statement
```python
pub fn sum(x: Array<i32, 3>) -> i32:
    let result: i32 = 0
    for i in x:
        result += i
    return result
```
loop

```python
pub fn for_with_break(x: Array<u256, 3>) -> u256:
    let result: u256 = 0
    for i in x:
        result = result + i
        if result == 15:
            break
    return result
```
loop_with_break

```python
pub fn for_with_continue(x: Array<u256, 3>) -> u256:
    let result: u256 = 0
    for i in x:
        if i % 2 == 0:
            continue
        result = result + i
    return result
```
loop_with_continue

#### Nested control flow
```python
pub fn sqrt(val: u256) -> u256:
    let z: u256
    if (val > 3):
        z = val
        let x: u256 = val / 2 + 1
        while (x < z):
            z = x
            x = (val / x + x) / 2
    elif (val != 0):
        z = 1
    return z
```
while

### Types
Types are not lowered except for some cases.
1. External contract type is lowered into `address`.
2. `String<N>` type is lowered into `Array<u8, N>`.

#### `If`, `for`, `break` and `continue`
We`ve already seen how to lower them.

#### `declaration`, `assign`, `emit`, and `assign`
Simply lower them into corresponding `inst`.

#### `assert`
Lowered into `branch` and `revert`
```python
pub fn bar(baz: i256):
    assert baz > 0 , "baz must be positive"
```
assert

#### `pass`
lowered into `revert`

#### Aggregate type construction
`tuple`, `struct`, `event`, and `array` constructions are lowered into the same instruction, `AggregateConstruct`.

```python
pub struct Foo:
    pub x: i32
    pub y: i32

pub fn aggregate():
    (1, 2)
    Foo(x:1, y:2)
    [1, 2]
```
aggregate

#### Aggregate access
attribute access and array index are lowered into the same instruction, `AggregateAccess`.
```python
pub struct Foo:
    pub x: i32
    pub y: i32

pub fn aggregate(arr: Array<i32, 1>, foo: Foo):
    arr[0]
    foo.y
```
aggregate_access
NOTE: attributes names are no longer used in MIR.

#### Bool literal
`true` is lowered into 1, `false` is lowered into 0.

#### String literal
String literal is lifted to a constant and stored in db. This information is collected and properly handled in the `codegen` phase.

#### Method call
A method call is lowered into normal call expression, method receiver is explicitly placed in the first argument.
```python
pub struct Foo:
   pub x: i32
    
   pub fn add(self, rhs: i32) -> i32:
       return self.x + rhs

pub fn method_call(foo: Foo):
    foo.add(1)
```
method_call

#### Ternary
Lowered into branch and jump instructions.
```python
pub fn ternary(b: bool):
    1 if b else 2
```
ternary

#### LogicalAnd/LogicalOr
Lowered into branch and jump instructions to reflect short circuit property.
```python
pub fn logical_and(x: i32, y: i32):
    (x > 0) and (y < 0)
```
logical_and

## MIR visualization
`fe  --mir path.fe` dumps [`dot`](https://graphviz.org/doc/info/lang.html).

## Misc
Currently, mir panics when file uses `ctx` related features like `msg.sender` etc, I think this would be resolved when https://github.com/ethereum/fe/pull/638 is merged.


Y-Nak:
Oops, I found some bugs related to `if` lowering and fixed them in the last commit.

sbillig (approved):
Epic work @Y-Nak, this looks really nice! I look forward to using it.

PR #663, opened 2022-02-11
Allow assignments from smaller string type to larger string type

Y-Nak:
### What was wrong?
Semantic analysis for string is too strict.
```python
let s: String<10> = "FOO"
```
should be accepted.


sbillig (approved):
:dancers:

PR #666, opened 2022-02-14
Require arguments to function calls to be labeled

### LINKED ISSUE #397: Named arguments
### What is wrong?

The parser and ast support python-style named arguments, but the names aren't checked in the rest of the compiler.
```
def f(x: u256, y: u256) -> u256:
  return x - y

def g() -> u256:
  return self.f(y=10, a=5)  # same as f(10, 5), so x = 10, y = 5
```

Furthermore, functions can be called without specifying argument names, which can be a source of bugs.
```
def _update(balance0: u256, balance1: u256, reserve0: u256, reserve1: u256):
   ...
self._update(token0.balanceOf(self.address), self.reserve0, token1.balanceOf(self.address), self.reserve1)
```

### How can it be fixed

I think we should require argument names, and require that arguments always be provided in the correct order. I also think we should use `:` instead of `=`. Eg `self.transfer(from: me, to: you)`

We could consider adopting swift's syntax and semantics here https://docs.swift.org/swift-book/LanguageGuide/Functions.html

In swift, function parameters have labels (which default to the parameter name). Labels must be included when calling the function, and arguments must be provided in the correct order. In cases where labels aren't warranted, they can be disabled for a given parameter.

Default case:
```
def transfer(sender: address, recipient: address):
  ...
self.transfer(sender: me, recipient: you)
```

Specify different label names, so calls are convenient but the function body can use long names (which can change without breaking the calling code):
```
def transfer(from sender_addr: address, to recipient_addr: address):
  ...
self.transfer(from: me, to: you)
```

Disable labels:
```
def max(_ val1: u256, _ val2: u256):
  ...
self.max(x, y) # instead of self.max(val1: x, val2: y)
```
note that some args can have labels while others don't

```
def new_account(_ addr: address, balance: u256 = 0):
  ...
new_account(sender)
new_account(sender, balance: 100)

```

sbillig:
Closes #397

```
pub fn transfer(self, to: address, _ x: u256):
  self.do_transfer(from: msg.sender, to, x + 1)

fn do_transfer(self, from: address, to: address, _ val: u256):
  pass
```

### To-Do

- [x] Clean up commit history


g-r-a-n-t:
:imp:

g-r-a-n-t (approved):
s/o to the man downstairs

Y-Nak (approved):
👍

PR #669, opened 2022-02-16
Some lints from clippy:pedantic

cburgdorf:
### What was wrong?

Nothing really, but I needed to find something useful to do that could be finished within a matter of hours. So I went through `cargo clippy -- -W clippy::pedantic` and looked for possible improvements to the code base.

### How was it fixed?

Either by just using `--fix` or manually applying the suggested changes if that wasn't possible.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-targets --all-features -- -D warnings -A clippy::upper-case-acronyms -A clippy::large-enum-variant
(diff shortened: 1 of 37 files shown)


.github/workflows/main.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-targets --all-features -- -D warnings -A clippy::upper-case-acronyms -A clippy::large-enum-variant
+          args: --workspace --all-targets --all-features -- -D warnings -A clippy::upper-case-acronyms -A clippy::large-enum-variant -W clippy::redundant_closure_for_method_calls
```
cburgdorf, line 70: I didn't want to switch on `pedantic` altogether but that one seems very useful because I can not think of a place where I  would rather like to have a redundant closure

Y-Nak (approved):
:+1:

PR #671, opened 2022-02-18
ABI Encode Unit Type

### LINKED ISSUE #442: unit type is not abi encodable at analyzer/src/namespace/types.rs:542:27
Compiling this file:

```
contract o
 event t
  _:()
```

```
thread 'main' panicked at 'unit type is not abi encodable', analyzer/src/namespace/types.rs:542:27
stack backtrace:
   0:     0x55bdbfeff890 - std::backtrace_rs::backtrace::libunwind::trace::h63b7a90188ab5fb3
   1:     0x55bdbfeff890 - std::backtrace_rs::backtrace::trace_unsynchronized::h80aefbf9b851eca7
   2:     0x55bdbfeff890 - std::sys_common::backtrace::_print_fmt::hbef05ae4237a4d72
   3:     0x55bdbfeff890 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h28abce2fdb9884c2
   4:     0x55bdbff18c0f - core::fmt::write::h3b84512577ca38a8
   5:     0x55bdbfefd2f2 - std::io::Write::write_fmt::h465f8feea02e2aa1
   6:     0x55bdbff01a05 - std::sys_common::backtrace::_print::h525280ee0d29bdde
   7:     0x55bdbff01a05 - std::sys_common::backtrace::print::h1f0f5b9f3ef8fb78
   8:     0x55bdbff01a05 - std::panicking::default_hook::{{closure}}::ha5838f6faa4a5a8f
   9:     0x55bdbff014b3 - std::panicking::default_hook::hfb9fe98acb0dcb3b
  10:     0x55bdbfcb0fef - fe_compiler::errors::DEFAULT_PANIC_HOOK::{{closure}}::{{closure}}::h8f119a8f95ba7581

Aborted
```

Using

```
[profile.dev]
debug = true
panic = "abort"
```

Maltby:
### What was wrong?
closes #442

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/abi/src/elements.rs
+++ b/crates/abi/src/elements.rs
@@ -106,7 +106,7 @@ impl JsonAbi for Base {
             Base::Numeric(Integer::I8) => "int8".to_string(),
             Base::Address => "address".to_string(),
             Base::Bool => "bool".to_string(),
-            Base::Unit => panic!("unit type is not abi encodable"),
+            Base::Unit => "".to_string(),
         }
     }
 }
--- a/crates/yulgen/src/db/queries/functions.rs
+++ b/crates/yulgen/src/db/queries/functions.rs
@@ -39,6 +39,13 @@ pub fn function_def(db: &dyn YulgenDb, function: FunctionId) -> yul::Statement {
 
     let function_name = identifier! { (db.function_yul_name(function)) };
     // all user-defined functions are given a return value during lowering
+
+    // ^ this is no longer true. functions that return the unit type do not
+    // get a return statement. Need a way to identify what functions don't have a
+    // return statement from here, so we can remove the '-> return_val' in those cases.
+
+    // let has_return_value = function_statements.iter().any(|x| x);
+
     function_definition! {
         function [function_name]([param_names...]) -> return_val {
             [function_statements...]
--- a/crates/yulgen/src/mappers/expressions.rs
+++ b/crates/yulgen/src/mappers/expressions.rs
@@ -39,7 +39,13 @@ pub fn expr(context: &mut FnContext, exp: &Node<fe::Expr>) -> yul::Expression {
         fe::Expr::List { .. } => panic!("list expressions should be lowered"),
         fe::Expr::Tuple { .. } => panic!("tuple expressions should be lowered"),
         fe::Expr::Str(_) => expr_str(exp),
-        fe::Expr::Unit => expression! { 0x0 },
+        // fe::Expr::Unit => expression! { 0x0 },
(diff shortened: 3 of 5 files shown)


sbillig:
@Maltby I went down this road once, and abandoned it because our compiler pipeline wasn't powerful enough to solve this in a satisfactory way. In yul, functions that don't return anything *really* don't return anything, so they can't be used in a position that expects some kind of value. There's no concept of a zero-sized value (everything is a u256).

Fe code like this:
```
fn f(a: u256, b: ()) -> u256:  return a
fn g():  pass
f(10, g())
```
is currently compiled to something like this in yul:
```
function f($a, $b) -> return_val {
    return_val := $a
    leave
}
function g() -> return_val {
    return_val := 0x0
    leave
}
pop(f(10, g()))
```
which only works because `g` returns a value. If we get rid of `g`'s return value, we can't use `g()` as an argument to `f`. We could modify the current yulgen so that it calls `g()` first, then calls `f(10, 0)` or something, but I think that the ideal solution here would be for all zero-sized types (including unit) to not exist at all by the time we reach the code generation (yulgen) phase. In a pre-yul optimization, we could eliminate all unit-type fn args, all unit-type struct fields, and zero-sized structs altogether.

The "middle" of the compiler is currently being revamped (#661, and a follow-up pr that'll generate yul from mir). My tentative suggestion is that unit types could be eliminated in an optimization pass on mir, and the mir-to-yul codegen could assume that unit types don't exist. Thoughts, @Y-Nak?

Y-Nak:
Yeah, I agree with the opinion. We should/could remove unit types from both function call-sites and def-sites. 
Also, I assume the removal would be done in the [legalization pass](https://llvm.org/docs/GlobalISel/Legalizer.html) because its removal would be specific for EVM and should be independent of MIR. 
So the compilation path in my mind currently would look like the below.
graphviz (2)

Y-Nak:
@Maltby I think it'd be better to fix this issue in `codegen` crate, which will be introduced in #676.
I'm still working on it,  and I could finish it within a week if there are not so many problems.  So could you wait for its completion, please?

Maltby:
@Y-Nak Ok, I'll wait for #676 to be merged. No rush at all! Thanks again for the guidance.

Y-Nak:
Hi @Maltby, I'm so sorry for the late response.
I've struggled so much through #676. Also, I fixed this issue in #676. 
I should have let you know about this, I'm really sorry for my rudeness.

PR #676, opened 2022-02-23
 Codegen

### LINKED ISSUE #249: Optimize access of struct fields in storage
### What is wrong?

When accessing struct fields in storage we are currently calculating storage pointers in YUL (which the optimizer doesn't optimize away)

https://github.com/ethereum/fe/blob/fa7c3c1bbd120959a3a474190b4a73eee623072f/compiler/src/yul/mappers/expressions.rs#L472

### How can it be fixed

1. We could pre-calculate them
2. Alternatively we might end up optimizing this another way as soon as we start storing structs tightly packed

### LINKED ISSUE #442: unit type is not abi encodable at analyzer/src/namespace/types.rs:542:27

### LINKED ISSUE #590: Memory-to-memory assignment ICE
### What is wrong?

Code like
```
somestruct.x = somestruct.y
# or
x[0] = x[1]
```
causes a panic in `yulgen::mappers::assignments::assign`.

### How can it be fixed

The code for the `(Location::Memory, Location::Memory)` assignment case should be replaced with an `mcopym(typ, target, value)`, with mcopym redefined to take two pointers (to match scopys), and the existing mcopym that returns a new ptr renamed to mclone or something.

### LINKED ISSUE #681: Panic when passing a struct containing an array to a function
### What is wrong?
The below code causes ICE.

```python
struct MyArray:
    pub x: Array<i32, 2>
    
    
contract Foo:
    pub fn bar(my_arr: MyArray):
        pass
```

Error:

```
Error: DeclarationError: Function "abi_decode_component_int32_calldata" not found.
  --> input.yul:33:40:
   |
33 |                     let decoded_val := abi_decode_component_int32_calldata(head_start, inner_offset)
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Error: DeclarationError: Variable count mismatch for declaration of "decoded_val": 1 variables and 0 values.
  --> input.yul:33:21:
   |
33 |                     let decoded_val := abi_decode_component_int32_calldata(head_start, inner_offset)
   |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


thread 'main' panicked at 'Yul compilation failed with the above errors', crates/driver/src/lib.rs:179:17

```

### LINKED ISSUE #709: Struct field called `number` leads to ICE
### What is wrong?

The following leads to a compiler error:

```
struct Foo:
  pub number: u256

contract Meh:

  pub fn yay() -> Foo:
    return Foo(number:2)
```

Leads to ICE

```
Error: ParserError: Cannot use builtin function name "number" as identifier name.
  --> input.yul:18:36:
   |
18 |             function $$foo$Foo.new(number) -> return_val {
   |                                    ^^^^^^


Error: ParserError: Expected keyword "data" or "object" or "}".
  --> input.yul:18:36:
   |
18 |             function $$foo$Foo.new(number) -> return_val {
   |                                    ^^^^^^


thread 'main' panicked at 'Yul compilation failed with the above errors', crates/driver/src/lib.rs:179:17

```

### How can it be fixed

Needs to be investivated

Y-Nak:
close #249, close #442, close #590, close #671, close #681, close #709

## Overview
1. Implement `fe-codegen` which lower `mir` into `yul`, also it will be used to lower `mir` into `sonatina-ir`.
2. Remove `fe-lowering` and `fe-yulgen`

## Storage layout
The new storage layout is almost the same as the solidity's one, which is described in [here](https://docs.soliditylang.org/en/v0.8.14/internals/layout_in_storage.html)

```
* The first item in a storage slot is stored lower-order aligned.
* Value types use only as many bytes as are necessary to store them.
* If a value type does not fit the remaining part of a storage slot, it is stored in the next storage slot.
```

## Memory layout
Unlike solidity, all primitive and aggregate types are stored in the same manner with the storage layout.
Also, the initial 96 bytes in memory are reserved as solidity does.

```
* 0x00 - 0x3f (64 bytes): scratch space for hashing methods
* 0x40 - 0x5f (32 bytes): currently allocated memory size (aka. free memory pointer)
```

## Event type
`event` types are treated in the same way as "normal" struct types to remove ad-hoc implementation specialized for event types. By regarding event types as normal struct types, we could easily allow event types to contain aggregate types for example.

But this abstraction makes gas efficiency worse because we should allocate memory every time when events appear.
e.g.

```python
 event MyEvent:
    x: u256
    y: u256
    
    
contract Foo:
    pub fn bar(ctx: Context):
        emit MyEvent(ctx, x: 1, y: 2)
```

In this case 64 bytes are allocated for `MyEvent`, while old `yulgen` doesn't allocate it.

## Followup
#716 

- [x] Clean up commit history


crates/fe/src/main.rs
```diff
@@ -77,6 +77,12 @@ pub fn main() {
                 .help("dump mir dot file")
                 .takes_value(false),
         )
+        .arg(
+            Arg::with_name("codegen")
+                .long("codegen")
+                .help("todo")
```
cburgdorf, line 83: That's a user facing `todo` so it should be replaced with something meaningful.

crates/fe/src/main.rs
```diff
@@ -296,3 +307,30 @@ fn mir_dump(input_path: &str) {
...
+                println!("{}", contract)
+            }
+            Err(err) => {
+                eprintln!("Unable to dump mir `{}", input_path);
+                print_diagnostics(&db, &err.0);
+                std::process::exit(1)
+            }
+        }
+    } else {
+        eprintln!("mir doesn't support ingot yet");
```
cburgdorf, line 333: It's not supported as far as the print out of yul is concerned or in general? Because the ingot tests seem to be working so I'm a bit confused what exactly it is about ingots that `mir` doesn't support yet.

crates/tests/src/snapshots/fe_compiler_tests__stress__tuple_stress.snap
```diff
@@ -1,15 +1,14 @@
...
-encode_my_tuple([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 1794 gas
+build_my_tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)]) used 605 gas
+read_my_tuple_item0([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 464 gas
+read_my_tuple_item1([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 522 gas
+read_my_tuple_item2([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 556 gas
+set_my_sto_tuple([Uint(42), Int(115792089237316195423570985008687907853269984665640564039457584007913129639910)]) used 44808 gas
+get_my_sto_tuple([]) used 922 gas
+emit_my_event([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 2484 gas
+build_tuple_and_emit([]) used 2537 gas
+encode_my_tuple([Tuple([Uint(26), Bool(true), Address(0x0000000000000000000000000000000000000042)])]) used 1205 gas
```
cburgdorf, line 13: Looks like we are getting some nice gas savings out of this :heart:

cburgdorf (approved):
Epic work! I took about two hours to go over this and the code looks very well written and clean. :+1: :ninja: 

But to be honest I would probably need days to digest it better. :exploding_head: So this is more like a rubber stamp approval. Chances are that @sbillig or @g-r-a-n-t will be able to provide better feedback.

crates/fe/src/main.rs
Y-Nak: Now `--codegen` is no longer necessary, so I just removed the argument.

crates/fe/src/main.rs
Y-Nak: If the input is an ingot, then the printing out of `MIR` is not supported yet. All functionality except for the pretty-printing works fine. I made the message clearer.

sbillig (approved):
:man_dancing: looks great. Sorry it took me a while. I'd say merge whenever you're ready.

I noticed a couple minor things that could be cleaned up now that the old lowering pass is gone, namely the `Ingot` item's `original` field, `ModuleSource::Lowered` (and maybe more?), but someone can do a small cleanup pr later of course.

Y-Nak:
@sbillig I opened an issue for cleaning up the leftovers.

PR #677, opened 2022-02-27
Add check for module item visibility

### LINKED ISSUE #598: Module item visibility control (`pub` modifier)
### What is wrong?

All items in a module can currently be `use`d in another module; there's no way to differentiate between `pub`lic and private items at the module level.

### How can it be fixed

Allow all module-level items to be declared as `pub` (eg `pub contract Foo: ...`, `pub struct Bar: ...`, etc), and don't allow anything not declared as pub to be imported by other modules.

Y-Nak:
closes #598 
### What was wrong?
Item visibility is not checked, so the ingot below is not rejected properly.

```python
# foo.fe
const FOO: i32 = 1

# main.fe
use foo::FOO
contract Main:
    fn foo() -> i32:
       return FOO
```

### How was it fixed?
Add check for visibility in `resolve_name` and `resolve_path`.

### To-Do

- [x] Clean up commit history


sbillig (approved):
Looks good. I think it would be ideal to emit an error on the `use` statement that imports the private item instead of emitting an error every time the item is referenced. I'd be happy to see this merged as-is though, as it's obviously much better than not checking at all.

Y-Nak:
Yeah, ideally it would be best.
Some difficulty here is that items can be referenced even if they don't appear in `use` statements. i.e., through wildcard or path.
In that case, a visibility check must be performed on a ref point. I guess that's why rustc and rust-analyzer emit the error at all reference points.
This error is easy to understand, so one option would be to reduce the error message to improve the user experience a little.

Y-Nak:
Anyway, I'll merge as is, this would fix the broken CI.

PR #684, opened 2022-03-04
fixed memory bug in storage maps

g-r-a-n-t:
### What was wrong?

The region of memory used to compute the slot of a storage map value was not being allocated.

### How was it fixed?

:keyboard: 

### To-Do

- [x] Clean up commit history


Y-Nak (approved):
~Just a question. When is memory used before initialization in current yulgen?~
Oh, an array is the one example I guess.

PR #692, opened 2022-03-29
mut

sbillig:
This adds the `mut` keyword, used in a fashion similar to rust (with some differences caused by the fact that we don't yet have explicit references in fe).

For example:
```
fn mutate_array(mut array: Array<u8, 3>) {  # this would be `array: &mut Array<u8, 3>`
  array[0] = 10
}
let mut foo: Array<u8, 3> = [1, 2, 3]
mutate_array(foo)
```

This is an important language feature, and this pr is a "better than nothing" (I hope) implementation, which should ultimately be replaced or improved. There are some problems caused by the fact that all compound types are implicitly references, and this pr doesn't really attempt to solve them.

Consider:
```
struct X {
  my_array: Array<u8, 3>
  pub fn foo(self) -> Array<u8, 3> { ... }
}
let x: X = ...
let mut y: Array<u8, 3> = x.foo()
```
Does `X::get_array` return a (non-mut) reference to the `my_array` field? Or does it return a new array? The return type in the current fe syntax is ambiguous, so there's no way to tell from that. We could attempt to answer this question by doing some dataflow analysis on the body of `foo` to figure out the *actual* return type and represent it as a `Type::Ref { inner: Type::Array { .. }}` or whatever, but this pr does no such thing. Instead, mutability is sort of tacked onto the side, in a way that mostly works, but can't possibly be perfect.

The correct solution is probably to copy rust (of course), where the return type would be either `&Array<u8, 3>` or `Array<u8, 3>` and would thus be unambiguous. Alternatively, if we want to keep the all-compound-types-are-implicit-references thing (like in python&co), we could use `-> mut Array<u8, 3>` to signify that the returned array can be mutated by the recipient. If the body of foo was `return self.my_array`, it would be an obvious error. I prefer the ownership semantics of rust, where `-> Array<u8, 3>` unambiguously means that the recipient is getting its own array that isn't shared with anyone else, but of course that brings with it a pile of complexity around references.

Note that this pr also definitely doesn't do anything about aliasing. There can be multiple mutable references to the same thing. It shouldn't, however, be possible to create a mutable reference to a thing that's declared as immutable.

~~Because I was in there messing with `mut ctx` stuff, I changed how `ctx` works in a couple cases. I could split this out into a separate pr, but knowing me I probably won't.~~ I did! #703 

### To-Do

- [ ] Clean up commit history


sbillig:
Closed in favor of #777

PR #694, opened 2022-04-23
Return instead of revert when contract is called without data

cburgdorf:
Otherwise, there would currently be no way to send ETH
to a contract.


sbillig (approved):
:+1:

Issue #687, opened 2022-03-11
Allow `pub unsafe fn __call__`, and `pub unsafe fn` for contract fns in general


sbillig:
### What is wrong?

> a contract function can be either `pub` or `unsafe`, but not both

^ This seems like a restriction that we can drop. :shrug:

dalmoveras:
Hey @sbillig  could you provide more details please?

sbillig:
@dalmoveras There's currently a check that disallows public contract functions from being unsafe; for example:
```
contract Foo:
  pub unsafe fn bar():
    std::evm::mstore(0, 100)
```

IIRC, I chose to do this because `unsafe` functions should only be callable from an `unsafe` context (another unsafe fn or an unsafe block), but public contract functions can be called from any external contract, and thus we can't restrict how they're called.

In hindsight, it doesn't seem useful to disallow `pub unsafe`.

Changing this is probably as simple as removing this block of code: https://github.com/ethereum/fe/blob/master/crates/analyzer/src/db/queries/functions.rs#L30-L39 and updating or removing a test or two.

PR #695, opened 2022-04-24
Allow public unsafe smart contract functions to be called from outside

Narasimha1997:
### What was wrong?
`unsafe` functions weren't allowed to be public (`pub`) as described in #687 

### How was it fixed?
The issue was fixed by removing the block of code that checked whether the function is both `pub` and `unsafe`. i.e by removing the [following check](https://github.com/ethereum/fe/blob/master/crates/analyzer/src/db/queries/functions.rs#L30-L39).

Later, the tests were updated to respect the same by updating the snapshots using `cargo insta review`.


g-r-a-n-t (approved):
looks good!

g-r-a-n-t:
just need to run `cargo fmt`

Narasimha1997:
@g-r-a-n-t , yup. Formatting issues are fixed in the latest commit.

PR #701, opened 2022-05-06
feat/support underscore in numberic

### LINKED ISSUE #149: Support underscores in numeric literals
### Abstract

We should support underscores `_` in numeric literals so that `1000000000000000000` can be written in the more readable form of  `1_000_000_000_000_000_000`.

### Motivation

Fe is intended to be used in smart contract development and variables and constants holding large numerals is a very well expected normal use case.  Especially since many variables are going to represent `Wei` and e.g. `1_000_000_000_000_000_000` Wei represent `1` Ether.

vuvoth:
### Proposed Changes

Support underscore for readable in numberic type.
Close #149 

### Implementation 
- Add regex rule to parser/lexer/token.rs

Please review my PR! Thanks!

### To-Do

- [ ] Clean up commit history


cburgdorf (commented):
Great!

Could you also add another test case here to prove that something like `100_000_` or `_1_0` is still rejected.

https://github.com/ethereum/fe/blob/c6edbe8e5d8793977bfeb6b6aa0dc3d224352857/crates/parser/tests/cases/errors.rs#L84

Lastly, can you add a file `newsfragments/149.feature.md` with the contents being something like

```
Support for underscores in numbers to improve readability e.g. `100_000`
```

This is to ensure the feature will be included in the upcoming release notes.

vuvoth:
> Could you also add another test case here to prove that something like `100_000_` or `_1_0` is still rejected.

I have problem is when I write "fail tests" case error but snap file body is empty. How I can handle it? 

> Lastly, can you add a file `newsfragments/149.feature.md` with the contents being something like

I don't know format of the `newfragments` files.  Have any source for me to reference ?

sbillig:
> I have problem is when I write "fail tests" case error but snap file body is empty. How I can handle it?

I guess you probably figured this out, but the snapshot testing stuff uses [insta](https://crates.io/crates/insta). You can just add a new test and run it. The test will initially fail, because the new snapshot won't match the "current" (non-existent) snapshot. To accept the snapshot, it's easiest to use `cargo insta review` (which requires a `cargo install cargo-insta` first).

> I don't know format of the `newfragments` files. Have any source for me to reference ?

Usually there are some examples in there, but they're removed on every release and added to the changelog. What you have is good, just drop the leading `#` and add a trailing newline.

cburgdorf (approved):

cburgdorf:
Thank you very much for your work @vuvoth 
Merged! :tada:

PR #702, opened 2022-05-06
Get rid of FixedSize

cburgdorf:
### What was wrong?

FixedSize makes the type system more complex which makes it harder to add new features.

### How was it fixed?

- Removed `FixedSize` type
- Introduced a `has_fixed_size()` API directly on Type
- Using this in a bunch of places where we would previously *try*  to convert a `Type` into a `FixedSize`

I'm not really happy with this. It does remove complexity from our type system but it also introduces new weaknesses. There's probably a better approach hiding somewhere that I haven't found yet...

### DIFF
--- a/crates/analyzer/src/namespace/types.rs
+++ b/crates/analyzer/src/namespace/types.rs
@@ -1,5 +1,5 @@
 use crate::context::AnalyzerContext;
-use crate::errors::{NotFixedSize, TypeError};
+use crate::errors::TypeError;
 use crate::namespace::items::{Class, ContractId, StructId};
 use crate::AnalyzerDb;
 
@@ -46,16 +46,6 @@ pub enum Type {
     /// The type of a contract while it's being executed. Ie. the type
     /// of `self` within a contract function.
     SelfContract(Contract),
-    Struct(Struct),
-}
-
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub enum FixedSize {
-    Base(Base),
-    Array(Array),
-    Tuple(Tuple),
-    String(FeString),
-    Contract(Contract),
     Struct(Struct),
 }
 
@@ -116,7 +106,7 @@ pub struct Map {
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Tuple {
-    pub items: Vec1<FixedSize>,
+    pub items: Vec1<Type>,
 }
 
 #[derive(Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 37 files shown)


crates/analyzer/src/namespace/types.rs
```diff
@@ -489,6 +505,18 @@ impl Type {
             _ => None,
         }
     }
+
+    pub fn has_fixed_size(&self) -> bool {
```
cburgdorf, line 509: I'm not really happy with this. Ideally we would have some trait (e.g `FixedSize` or `HasFixedSize`) that `Base`, `Array` and friends would implement and then this method would look like the following pseudo code:

```
pub fn has_fixed_size(&self) -> bool {
   self.as_that_sweet_trait().has_fixed_size()
}
```
But that kind of setup does't seem to be a natural Rust pattern so even if we had an `HasFixedSize` trait we would still end up breaking the `Type` into a match arm for each type which isn't ideal (violating the open-closed principle).

So, just pointing out that I'm not happy with this API and maybe someone has a better idea how to solve that.

crates/yulgen/src/mappers/assignments.rs
```diff
@@ -20,8 +21,7 @@ pub fn assign(context: &mut FnContext, stmt: &Node<fe::FuncStmt>) -> yul::Statem
         let value = expressions::expr(context, value_node);
 
         let target_attributes = context.expression_attributes(target_node);
-        let typ = FixedSize::try_from(target_attributes.typ.clone()).expect("invalid attributes");
-
+        let typ = target_attributes.typ.as_evm_sized();
```
cburgdorf, line 24: `as_evm_sized()` will panic if a wrong type makes it here but it's been like that before with that `FixedSize::try_from(..).expec(..)` pattern.

crates/yulgen/src/types.rs
```diff
@@ -166,15 +190,16 @@ pub trait AsAbiType {
...
-            FixedSize::String(string) => string.as_abi_type(db),
-            FixedSize::Contract(_) => AbiType::Address,
-            FixedSize::Struct(val) => val.as_abi_type(db),
+            Type::Base(base) => base.as_abi_type(db),
+            Type::Array(array) => array.as_abi_type(db),
+            Type::Tuple(tuple) => tuple.as_abi_type(db),
+            Type::String(string) => string.as_abi_type(db),
+            Type::Contract(_) => AbiType::Address,
+            Type::Struct(val) => val.as_abi_type(db),
+            _ => panic!("Can not force {self} into an AbiType"),
```
cburgdorf, line 202: Not ideal

sbillig (approved):
:+1: I think this is improvement.

PR #703, opened 2022-05-07
ctx tweak

sbillig:
This makes minor changes to the rules of the context object.

- ctx *is not* required when casting an address to a contract type. Eg `let foo: Foo = Foo(address(0))`
- ctx *is* required when calling an external contract function that requires ctx

Example:

```
contract Foo {
  pub fn emit_stuff(ctx: Context) {
    emit Stuff(ctx) // will be changed to ctx.emit(Stuff{}) someday
  }
}
contract Bar {
  pub fn call_foo_emit_stuff(ctx: Context) {
    Foo(address(0)).emit_stuff(ctx)
  }
}
event Stuff {}
```

### To-Do

- [x] Clean up commit history


g-r-a-n-t (approved):
looks good :+1:

Issue #682, opened 2022-03-02
Reject infinite size struct definition


Y-Nak:
### What is wrong?
Currently, the below codes are not appropriately rejected and lead to ICE when it is instantiated.

#### Self recursion
```python
struct Foo:
    foo: Foo

contract MyContract:
    pub fn func(foo: Foo):
       pass
```

#### Mutual recursion
```python
struct Foo:
    x: Bar
    
struct Bar:
    x: Foo

contract MyContract:
    pub fn func(foo: Foo):
       pass
```

### How can it be fixed
Reject recursive struct definition in `analyzer`.
This could be done by checking recursions in `analyzer::items::StructId::sink_diagnostics` using `analyzer::items::StructId::struct_depency_graph`.

Narasimha1997:
I can work on this.

Narasimha1997:
@Y-Nak 

`salsa` create panics when it sees a cycle in the dependency graph.

Here is the error message:

```
---- struct_recursive_cycles stdout ----
thread 'struct_recursive_cycles' panicked at 'Internal error, cycle detected:

DatabaseKeyIndex { group_index: 0, query_index: 57, key_index: 0 }
DatabaseKeyIndex { group_index: 0, query_index: 57, key_index: 1 }
', /home/narasimha/.cargo/registry/src/github.com-1ecc6299db9ec823/salsa-0.16.1/src/lib.rs:490:48
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Is there any way we can catch this Panic and handle it?

 Check out the related issue [here](Cycle recovery · Issue #6 · salsa-rs/salsa).

Y-Nak:
@Narasimha1997 
Hi! Thanks for working on the issue!

One way to catch the cycle would be to use `salsa::cycle`. We use it [here](https://github.com/ethereum/fe/blob/6d9c722afeca32334c41422d8daddbee429fb1f9/crates/analyzer/src/db.rs#L83) for example.

Narasimha1997:
@Y-Nak thanks for reverting back.
Also if possible can you suggest me a suitable error message for this?

PR #704, opened 2022-05-08
Detect cyclic field definitions in struct dependency graph

Narasimha1997:
### What was wrong?
As stated by @Y-Nak in the issue #682 
### How was it fixed?
By adding `struct_cycles` as salsa cycle callback that throws an error when cycles are detected.


Y-Nak (approved):
Thanks for opening the PR.
The implementation itself is perfect 💯 

nits:
1. Could you add `close #682` to the PR comment to close the issue automatically when the PR is merged?
2. Could you add `newsfragments/682.bugfix.md` to ensure the fix will be included in the next release note?

Narasimha1997:
@Y-Nak 
Yup, did the changes you suggested.

Y-Nak:
Thanks so much!

Issue #558, opened 2021-10-04
Gate certain features behind a `context` object.


cburgdorf:
### What is wrong?

One of the main objectives of Fe is to be explicit and easy to audit. The EVM exposes certain features such as *writing logs*, *creating  contracts* or obtaining the current block number and it should be easy to determine from the *signature* of a function whether the function has the capabilities to use such features or not.
 
This is currently not the case, e.g we might currently have a function such as the following which - from it's signature - looks pure when in reality it does modify the blockchain (by emitting a log).

```
pub def looks_pure_but_isnt():
  emit SomeEvent()
```

#### The `Context` object

We propose to introduce a `Context` object which gates access to features such as:

- emitting logs
- creating contracts
- transferring Ether
- reading message info
- reading block info

The `Context` object needs to be passed as a parameter to the function e.g. the example above would be rewritten as:

```
pub fn emits_event(ctx: mut Context):
  ctx.emit(SomeEvent())
```

Similar, creating a contract via `create`/`create2` would also require to have access to the `Context` object:

```
pub fn creates_contract(ctx: mut Context):
  ctx.create2(...)
```

As well does *reading*  block chain information such as the current block number.

```
pub fn retrieves_blocknumber(ctx: Context):
  ctx.blocknumber()
```

There are a few special rules about the context object:

1. The `Context` object has a defined location in the parameter list. It either is the *first* parameter if the function does not take `self` (e.g. `pub fn foo(context:Context)`) or it comes second if the function does take `self` (e.g. `pub fn foo(self, context:Context)`). Any other index in the parameter list is a compile time error.

2. The `Context` object is automatically injected when a function is called externally but it has to be passed **explicitly** when the function is called from another Fe function e.g.

```
# The context object is automatically injected when this is called externally
pub fn amplified_block(ctx: Context) -> u256:
  # but it has to be passed along in this function call
  return retrieves_blocknumber(ctx) * 1000

fn retrieves_blocknumber(ctx: Context):
  return ctx.blocknumber()
```

#### Taking `mut Context` vs `Context`

As the eagle-eyed reader might have noticed the examples above either use  `mut Context` or `Context`. This proposal can be implemented independent of having support for `mut` in Fe but it would naturally benefit from having `mut` support because all functionality that modifies the blockchain such as creating logs or contracts would require to obtain a `mut Context` reference whereas readonly access such as `ctx.blocknumber()` does not need require a mutable reference to the context.

#### ABI conformity

[...]

PR #705, opened 2022-05-09
Temporary default stateMutability to payable in ABI

cburgdorf:
### What was wrong?

The ABI metadata that we generate does currently not include the `stateMutability` field. This piece of information is important for tooling such as hardhat because it determines whether a function needs to be called with or without sending a transaction.

As soon as we have support for `mut self` and `mut ctx` we will be able to derive that information from the function signature. In the meantime we should default to `payable`. 

### How was it fixed?

Add `stateMutability` field and default it to `payable`.

### DIFF
--- a/crates/abi/src/builder.rs
+++ b/crates/abi/src/builder.rs
@@ -1,6 +1,6 @@
 use crate::elements::{
     Component, Contract, Event, EventField, FuncInput, FuncOutput, FuncType, Function, JsonAbi,
-    ModuleAbis,
+    ModuleAbis, StateMutability,
 };
 use crate::AbiError;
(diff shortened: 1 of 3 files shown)


Jzow:
Do you mean to add a parameter to the `&mut` variability of rust to control the variability of the **function**?

cburgdorf:
@Jzow Take a look at the "abi conformity" section of https://github.com/ethereum/fe/issues/558
In the future we will have support for the `mut` keyword enabling us to take `self` or `mut self` and `ctx` or `mut ctx`. We can then derive the `stateMutability` section of the ABI based on that. For now, we can not which gives tooling such as hardhat a hard time because it can not infer whether a function needs to be called with a transaction or if a readonly call would do it.

g-r-a-n-t (approved):

PR #707, opened 2022-05-12
braces

sbillig:
Stage 1 of the pythonish -> rusty syntax transition.
- braces instead of colons and indents
- `else if` instead of `elif`
- no more `pass` statement

```
struct Empty {}
contract Foo {
  pub fn bar() {
    if true {
      1 + 2
      3 + 4
    } else if false {
    } else {
  }
}
```

### To-Do

- [x] Clean up commit history


docs/src/spec/expressions/name.md
```diff
@@ -9,11 +9,12 @@ A *name expression* resolves to a local variable.
 Example:
 
 ```python
```
cburgdorf, line 11: Not needed for this PR but we should probably switch from `python` syntax highlighting to `rust` syntax highlighting or else the snippets in the docs will look less good. I think that will also require a tweak in some script that performs the tests on top of the docs snippets.

docs/src/spec/lexical_structure/tokens.md
```diff
@@ -8,14 +8,6 @@ In Fe whitespace is not just used to guide the human eye but it also carries mea
 
 A token that represents a new line.
 
-### `INDENT`
```
cburgdorf, line 11: The paragraph on top ("significant whitespace") should be removed as well.

docs/validate_doc_examples.py
```diff
@@ -77,15 +80,19 @@ def create_fe_file(snippet: CodeSnippet) -> pathlib.Path:
...
+    build = subprocess.run(["cargo", "build"])
+    if build.returncode == 1:
+        exit(1)
+
     for path in get_all_doc_files():
         for snippet in get_all_snippets_in_file(path):
             if not snippet.unique_name() in SKIP_LIST:
                 fe_snippet = create_fe_file(snippet)
-                compile_result = run_snippet(fe_snippet)
+                compile_result = subprocess.run(["./target/debug/fe", str(fe_snippet), '--overwrite', '-e', 'abi'])
```
cburgdorf, line 95: I guess we should either move this back in `run_snippet` or remove `run_snippet`

newsfragments/707.feature.md
```diff
@@ -0,0 +1,28 @@
+Braces! Fe has abandoned python-style significant whitespace in favor of the
+trusty curly brace.
```
cburgdorf, line 2: Just making sure `trusty` is an intended word play :D

cburgdorf (approved):
Looks great! Just left a few nitpicks.

docs/src/spec/expressions/name.md
sbillig: mdbook strips out lines that start with `#` in rust-formatted blocks, so I added a simple custom fe syntax highlighter instead.

sbillig:
I'll merge this after #676 lands.

docs/src/spec/expressions/name.md
cburgdorf: Sweet

mriise:
<3 I am a big fan of curly bois

PR #710, opened 2022-05-18
Implement traits, impls and generic functions with trait bounds

cburgdorf:
### Abstract

We want to have traits as well as generic functions with trait bounds. One of the main motivations is to properly implement `emit` on the `Context` struct as `pub fn emit<T: Event>(val: T) { ... }`

### How was it fixed?

Working example below:

https://github.com/ethereum/fe/blob/f97afea7937202e32b40de5f866b960b60d84c0a/crates/test-files/fixtures/features/generic_functions.fe#L1-L42

Some notes:

- Generic functions a currently monomorphized at MIR lowering
- Because trait functions are usually (and currently only) defined without body the signature part was ripped out of `FunctionId` and into `FunctionSigId` to allow traits to reuse a bunch of code that doesn't care about the body anyway.
- There are certain limitations that can potentially be improved in the future e.g.
  - support traits for non-struct types
  - allow trait methods to define a body as a default implementation
  - support generic functions in traits (e.g. `trait PartialEq { pub fun eq<T>(self, other: T); }`)
  - support multiple bounds
  - support unbounded generics
  - support trait associated functions (e.g. `SomeTrait::some_fun()`

### DIFF
--- a/crates/analyzer/src/namespace/items.rs
+++ b/crates/analyzer/src/namespace/items.rs
@@ -1,4 +1,5 @@
 use crate::context;
+use fe_common::diagnostics::Label;
 
 use crate::context::{Analysis, Constant};
 use crate::errors::{self, IncompleteItem, TypeError};
@@ -9,6 +10,7 @@ use crate::{builtins, errors::ConstEvalError};
 use fe_common::diagnostics::Diagnostic;
 use fe_common::files::{common_prefix, Utf8Path};
 use fe_common::{impl_intern_key, FileKind, SourceFileId};
+use fe_parser::ast::GenericParameter;
 use fe_parser::node::{Node, Span};
 use fe_parser::{ast, node::NodeId};
 use indexmap::{indexmap, IndexMap};
@@ -16,6 +18,8 @@ use smol_str::SmolStr;
 use std::ops::Deref;
 use std::rc::Rc;
 use strum::IntoEnumIterator;
+
+use super::types::Type;
 
 /// A named item. This does not include things inside of
 /// a function body.
@@ -469,9 +473,18 @@ impl ModuleId {
         db.module_is_incomplete(*self)
     }
 
+    pub fn is_in_std(&self, db: &dyn AnalyzerDb) -> bool {
+        self.ingot(db).name(db) == "std"
+    }
+
     /// Includes duplicate names
     pub fn all_items(&self, db: &dyn AnalyzerDb) -> Rc<[Item]> {
         db.module_all_items(*self)
+    }
+
+    /// Includes duplicate names
+    pub fn all_impls(&self, db: &dyn AnalyzerDb) -> Rc<[Impl]> {
+        db.module_all_impls(*self)
     }
 
     /// Returns a map of the named items in the module
@@ -669,6 +682,10 @@ impl ModuleId {
         self.all_items(db)
             .iter()
             .for_each(|id| id.sink_diagnostics(db, sink));
+
+        self.all_impls(db)
+            .iter()
+            .for_each(|val| val.sink_diagnostics(db, sink));
     }
 
     #[doc(hidden)]
@@ -747,6 +764,7 @@ impl ModuleConstantId {
 pub enum TypeDef {
     Alias(TypeAliasId),
     Struct(StructId),
+    Trait(TraitId),
     Contract(ContractId),
     Primitive(types::Base),
 }
@@ -778,6 +796,7 @@ impl TypeDef {
         match self {
             TypeDef::Alias(id) => id.name(db),
             TypeDef::Struct(id) => id.name(db),
+            TypeDef::Trait(id) => id.name(db),
             TypeDef::Contract(id) => id.name(db),
             TypeDef::Primitive(typ) => typ.name(),
         }
@@ -787,6 +806,7 @@ impl TypeDef {
         match self {
             TypeDef::Alias(id) => Some(id.name_span(db)),
             TypeDef::Struct(id) => Some(id.name_span(db)),
+            TypeDef::Trait(id) => Some(id.name_span(db)),
             TypeDef::Contract(id) => Some(id.name_span(db)),
             TypeDef::Primitive(_) => None,
         }
@@ -800,6 +820,10 @@ impl TypeDef {
                 name: id.name(db),
                 field_count: id.fields(db).len(), // for the EvmSized trait
             })),
+            TypeDef::Trait(id) => Ok(types::Type::Trait(types::Trait {
+                id: *id,
+                name: id.name(db),
+            })),
             TypeDef::Contract(id) => Ok(types::Type::Contract(types::Contract {
                 id: *id,
                 name: id.name(db),
@@ -812,6 +836,7 @@ impl TypeDef {
         match self {
             Self::Alias(id) => id.is_public(db),
             Self::Struct(id) => id.is_public(db),
+            Self::Trait(id) => id.is_public(db),
             Self::Contract(id) => id.is_public(db),
             Self::Primitive(_) => true,
         }
@@ -821,6 +846,7 @@ impl TypeDef {
         match self {
             TypeDef::Alias(id) => Some(id.parent(db)),
             TypeDef::Struct(id) => Some(id.parent(db)),
+            TypeDef::Trait(id) => Some(id.parent(db)),
             TypeDef::Contract(id) => Some(id.parent(db)),
             TypeDef::Primitive(_) => None,
         }
@@ -830,6 +856,7 @@ impl TypeDef {
         match self {
             TypeDef::Alias(id) => id.sink_diagnostics(db, sink),
             TypeDef::Struct(id) => id.sink_diagnostics(db, sink),
+            TypeDef::Trait(id) => id.sink_diagnostics(db, sink),
             TypeDef::Contract(id) => id.sink_diagnostics(db, sink),
             TypeDef::Primitive(_) => {}
         }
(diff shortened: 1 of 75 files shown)


cburgdorf:
I just changed this back to draft because I will update it soon™ with all the remaining parts to cover `impl` blocks. Just need a few more cleanups and bits here and there.

crates/analyzer/src/namespace/items.rs
```diff
@@ -1324,6 +1489,292 @@ impl StructFieldId {
...
+                    self.data(db).ast.span,
+                    format!(
+                        "Neither `{}` nor `{}` are in the ingot of the `impl` block",
+                        self.data(db).receiver,
+                        self.data(db).trait_id.name(db)
+                    ),
+                )],
+                vec![],
+            ))
+        }
```
Y-Nak, line 1566: Why is this restriction necessary?

crates/analyzer/src/namespace/items.rs
```diff
@@ -1324,6 +1489,292 @@ impl StructFieldId {
...
+    }
+
+    pub fn sink_diagnostics(&self, db: &dyn AnalyzerDb, sink: &mut impl DiagnosticSink) {
```
Y-Nak, line 1569: It seems trait visibility is not considered, should we validate it in this function?

crates/mir/src/lower/function.rs
```diff
@@ -28,20 +31,53 @@ use crate::{
...
+pub fn lower_monomorphized_func_signature(
+    db: &dyn MirDb,
+    func: analyzer_items::FunctionId,
+    concrete_args: &[Type],
+) -> FunctionId {
     // TODO: Remove this when an analyzer's function signature contains `self` type.
     let mut params = vec![];
+    let mut concrete_args: &[Type] = concrete_args;
     let has_self = func.takes_self(db.upcast());
+    //println!("{} has self {}", func.name(db.upcast()), has_self);
```
Y-Nak, line 45: leftover?

Y-Nak (commented):
I left some comments.

Also, the below code causes ICE.

```
trait Computable {
  fn compute(self) -> u256;
}

struct Mac {
  pub fn compute(self) -> u256 {
     return 0
  }
}

impl Computable for Mac {
  fn compute(self) -> u256 {
    return 1
  }
}

contract Bar {
  pub fn foo() -> u256 {
    let mac: Mac = Mac()
    return mac.compute()
  }
  
  pub fn foo2() -> u256 {
    let mac: Mac = Mac()
    return foo_computable(val: mac)
  }
}

fn foo_computable<T: Computable>(val: T) -> u256 {
  return val.compute()
}
```
```
thread 'main' panicked at 'should be lowered in `lower_types_in_functions`', crates/mir/src/lower/types.rs:25:13
```
I haven't looked into this yet, but it'd be better to address this in this PR.

crates/analyzer/src/namespace/items.rs
cburgdorf: I believe it is limiting potential conflicts. I was following the rules that rust imposes on traits where you either need to own the trait or the type for which you want to write an `impl`.

A bit of background that I found
- https://stackoverflow.com/questions/43989065/cant-implement-a-trait-i-dont-own-for-all-types-that-implement-a-trait-i-do-ow
- https://github.com/rust-lang/rust/issues/23086

crates/mir/src/lower/function.rs
cburgdorf: Oops :sweat_smile:

crates/analyzer/src/namespace/items.rs
```diff
@@ -1109,24 +1127,152 @@ impl FunctionId {
             None
         }
     }
+    pub fn signature(&self, db: &dyn AnalyzerDb) -> Rc<types::FunctionSignature> {
+        db.function_signature(*self).value
+    }
 
     pub fn is_public(&self, db: &dyn AnalyzerDb) -> bool {
-        self.pub_span(db).is_some()
+        self.is_trait_fn(db) || self.pub_span(db).is_some()
```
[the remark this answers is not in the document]
cburgdorf: Ah, that's right. I believe functions defined in `impl` blocks should be considered `pub` as well :+1:

crates/analyzer/src/namespace/items.rs
```diff
@@ -747,6 +772,7 @@ impl ModuleConstantId {
 pub enum TypeDef {
     Alias(TypeAliasId),
     Struct(StructId),
+    Trait(TraitId),
```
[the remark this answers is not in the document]
cburgdorf: Ah, yes. That's a good point. I initially made it a type because I started by ignoring generics to just support something like:

```rust
pub fn bar(val: SomeTrait) {
}
```

But then later I noticed that it would be better to add generics first and only support 

```rust
pub fn bar<T: SomeTrait(val: T) {
}
```

I don't know if we ever want to support the first example (or maybe `val: impl SomeTrait` as rust does) or just leave it to generics entirely. My current thinking is that we should not support the first example and leave it to generics to work with traits and hence could remove `Trait` as a type and add it as its own category of `Item`.

crates/analyzer/src/namespace/items.rs
Y-Nak: Oh, that makes sense👍

cburgdorf:
>Also, the below code causes ICE.

Thanks, this code should have been rejected because generics should not be supported yet outside of struct functions. Fixed it.

crates/analyzer/src/namespace/items.rs
```diff
@@ -1109,24 +1127,152 @@ impl FunctionId {
...
+    }
+
+    pub fn generic_params(&self, db: &dyn AnalyzerDb) -> Vec<GenericParameter> {
+        self.data(db).ast.kind.generic_params.kind.clone()
+    }
+
+    pub fn generic_param(&self, db: &dyn AnalyzerDb, param_name: &str) -> Option<GenericParameter> {
+        self.generic_params(db)
+            .iter()
+            .find(|param| match param {
```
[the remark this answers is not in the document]
cburgdorf: Fixed locally

Y-Nak:
Ok, I misunderstood the cause of ICE.  I wrote the test case to check what would happen when method name conflicts happen, and it caused ICE, so I wrongly suspected that function name conflict was the cause of the ICE.

After modifying the test case, it causes ICE because of name conflicts.
```
trait Computable1 {
    fn compute(self) -> u256;
}

trait Computable2 {
    fn compute(self) -> u256;
}


struct Mac {
}

impl Computable1 for Mac {
    fn compute(self) -> u256 {
        return 1
    }
}

impl Computable2 for Mac {
    fn compute(self) -> u256 {
        return 2
    }
}

contract Bar {
    pub fn foo() -> u256 {
        let mac: Mac = Mac()
        let mys: MyStruct = MyStruct()
        return mys.foo_computable1(val: mac)
    }
    
    pub fn foo2() -> u256 {
        let mac: Mac = Mac()
        let mys: MyStruct = MyStruct()
        return mys.foo_computable2(val: mac)
    }
}

struct MyStruct {
    pub fn foo_computable1<T: Computable1>(self, val: T) -> u256 {
        return val.compute()
    }

    pub fn foo_computable2<T: Computable2>(self, val: T) -> u256 {
        return val.compute()
    }
}
```
 ↓
```
Error: DeclarationError: Function name foo$Mac$compute already taken in this scope.
  --> input.yul:51:13:
   |
51 |             function foo$Mac$compute() -> $$ret {
   |             ^ (Relevant source part starts here and spans across multiple lines).


thread 'main' panicked at 'Yul compilation failed with the above errors', crates/driver/src/lib.rs:192:13

```

So we need to mangle the function name to fix the bug, I think it would be better to implement the mangling in `codegen`, not in `mir`.

crates/analyzer/src/namespace/items.rs
cburgdorf: Eagle eyes :eagle: Fixed it.

cburgdorf:
Oh, I see. I believe this is because here we are considering the `name` of the `class` which for the `impl` is the class of the receiver. If instead we would have the `impl` here instead then we could consider a combination of *trait name* and *receiver name* which should make it unique.

https://github.com/ethereum/fe/blob/9ff971ebd1a49ff927c50c6cbc829ba3b3862ca8/crates/mir/src/db/queries/function.rs#L84-L86

> think it would be better to implement the mangling in `codegen`, not in `mir`

Do you suggest to fix it here? 

https://github.com/ethereum/fe/blob/9ff971ebd1a49ff927c50c6cbc829ba3b3862ca8/crates/codegen/src/db/queries/function.rs#L19-L24

Wouldn't that mean that we let `name_with_class` return an faulty name? Why wouldn't we want to fix it earlier?

Follow up question, we could fix that by making `Impl` a variant of `Class` but you suggested to get rid of `Class` altogether.  If we do that should `Impl` also become an `Item` and should we consider *trait name* + *receiver name* as the name of the impl?

crates/analyzer/src/namespace/items.rs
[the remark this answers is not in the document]
cburgdorf: That is now fixed as well

cburgdorf:
Thanks again for the feedback @Y-Nak 
I renamed `name_with_class` to `debug_name` and made it format the impl function as `<MyStruct as MyTrait>::method_name` per your suggestion. I also tweaked `symbol_name` directly so that `MyTrait$MyStruct$method_name` is used there. 

I didn't feel like adding the removal of `Class` to this PR so instead I added `Class::Impl` for now as an interim solution.

I think the only thing left from your suggestions is the removal of `Trait` from `TypeDef` which I started with but then ran into issues. Since I'm not yet entirely sure if we eventually want to support using traits in places for types maybe we can leave that as is for now and revisit later as we extend and tune the trait support?

Y-Nak (approved):
Great work @cburgdorf!

> we can leave that as is for now and revisit later as we extend and tune the trait support?
 
👍

PR #713, opened 2022-05-22
Unary minus mismatch type

### LINKED ISSUE #651: Analyzer doesn't check unary minus operation on unsigned type
### What is wrong?
```python
contract Foo:
    pub fn bar(self) -> u32:
        let unsigned: u32 = 1
        return -unsigned
```

It's better to reject the code above for safety.

### How can it be fixed
Tweaking [this lines](https://github.com/ethereum/fe/blob/50443ebbf15ceb567d592fcab8f0ccee96d4355b/crates/analyzer/src/traversal/expressions.rs#L762-L777) would fix the issue.

vuvoth:
### What was wrong?
Fix #651 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -705,18 +705,33 @@ fn expr_unary_operation(
 
         return match op.kind {
             fe::UnaryOperator::USub => {
-                let int_type = expected_type.as_int().unwrap_or(Integer::I256);
+                let expected_int_type = expected_type.as_int().unwrap_or(Integer::I256);
                 match operand_attributes.typ {
                     Type::Base(Base::Numeric(_)) => {
                         if let fe::Expr::Num(num_str) = &operand.kind {
                             let num = -to_bigint(num_str);
-                            validate_numeric_literal_fits_type(context, num, exp.span, int_type);
+                            validate_numeric_literal_fits_type(
+                                context,
+                                num,
+                                exp.span,
+                                expected_int_type,
+                            );
+                        }
+                        if !expected_int_type.is_signed() {
+                            context.error(
+                                "Can not apply unary operator",
+                                op.span + operand.span,
+                                &format!(
+                                    "Expected unsigned type `{}`. Unary operator `-` can not be used here.",
+                                    expected_int_type,
+                                ),
+                            );
                         }
                     }
                     _ => emit_err(context, "a numeric type"),
                 }
                 Ok(ExpressionAttributes::new(
-                    Type::int(int_type),
+                    Type::int(expected_int_type),
                     Location::Value,
                 ))
             }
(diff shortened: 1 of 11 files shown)


crates/analyzer/src/traversal/expressions.rs
```diff
@@ -705,18 +705,34 @@ fn expr_unary_operation(
...
-                            validate_numeric_literal_fits_type(context, num, exp.span, int_type);
+                            validate_numeric_literal_fits_type(
+                                context,
+                                num,
+                                exp.span,
+                                expected_int_type,
+                            );
+                        }
+
+                        if !(operand_type.is_signed() || expected_int_type.is_signed()) {
```
cburgdorf, line 721: The following code still isn't rejected though even though it should.

```
contract Foo:
    pub fn bar(self) -> u32:
        let signed: i32 = 1
        return -signed
```
I haven't dug into it but I think it would be worth to address this as well.

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -705,18 +705,34 @@ fn expr_unary_operation(
...
+                            context.error(
+                                "types always mismatch",
+                                operand.span,
+                                &format!(
+                                    "this has type `{}`; always return signed integer after apply '-' operator",
```
cburgdorf, line 726: minor grammer nitpicks: "this has type `{}`; always return**s** signed integer after apply**ing** '-' operator"

cburgdorf (commented):
We will need a file `651.bugfix.md` in our `newsfragments` directory that briefly explains the fixed bug so that it gets included in the next release notes.

crates/analyzer/src/traversal/expressions.rs
cburgdorf: >signed is i32 => -signed could be unsigned number(u32 in this case). => We shouldn't throw error.

Yes, I thought so, too for a brief moment but I checked Rust and they would not let you do this and so I think we shouldn't either.


```
fn foo() -> u32 {
    let signed: i32 = -1;
    return -signed
}
```

Even though `-` applied to `-1` would make it `1` and would fit `u32` Rust will still reject this and I think its for a good reason.

vuvoth: > ```rust
> fn foo() -> u32 {
>     let signed: i32 = -1;
>     return -signed
> }
> ```

I worry how we can casting type of expression. Code below will throw error when compile. 

```python
contract Foo:
    pub fn bar() -> u32:
        let a: i32 = -1
        return u32(-a)
```

We can do somthing like below to casting type, but I don't like it.

```python
contract Foo:
    pub fn bar() -> u32:
        let a: i32 = -1
        let b: i32 = -a
        return u32(b)
```

cburgdorf:
Thanks again @vuvoth I did some minor tweaks to the message (hope you don't mind). I'll merge it when CI goes green.

vuvoth:
> Thanks again @vuvoth I did some minor tweaks to the message (hope you don't mind). I'll merge it when CI goes green.

No problem! Thank you a lot!

PR #722, opened 2022-05-26
Fix regression of stateMutability not being included in ABI

cburgdorf:
### What was wrong?

The new ABI crate that landed with #676 did not have the `stateMutability` field included. It should default to `payable` for now.

### How was it fixed?

This is what #705 did but for the new ABI crate

### DIFF
--- a/crates/abi/src/function.rs
+++ b/crates/abi/src/function.rs
@@ -4,6 +4,16 @@ use serde::Serialize;
 
 use super::types::AbiType;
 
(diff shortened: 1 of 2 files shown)


Y-Nak (approved):
I'm sorry that I forgot to reflect the change. Thanks!

cburgdorf:
> I'm sorry that I forgot to reflect the change. Thanks!

No worries!

PR #727, opened 2022-05-31
Remove lowering cruft from ingot/module items

### LINKED ISSUE #720: Cleanup leftover of old lowering crate
### What is wrong
We successfully removed the old lowering crate, but there are still leftovers that were necessary for `lowering`.

### How can it be fixed
Identify leftovers and remove them.

sbillig:
### What was wrong?

Closes #720 (I think; @Y-Nak, are you aware of any other leftorovers?)

The only functional change is that `db.set_ingot_modules` has been replaced with `db.set_ingot_files`. The former was a compromise to work with lowered modules. Now the latter will be used by the (imaginary) language server to update the list of files in a project.

### How was it fixed?
:brain:

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/tests/analysis.rs
+++ b/crates/analyzer/tests/analysis.rs
@@ -416,7 +416,7 @@ fn function_diagnostics(fun: items::FunctionId, db: &dyn AnalyzerDb) -> Vec<Diag
         // signature
         build_debug_diagnostics(&[(fun.data(db).ast.span, &fun.signature(db))]),
         // declarations
-        label_in_non_overlapping_groups(&lookup_spans(&body.var_decl_types, &body.spans)),
+        label_in_non_overlapping_groups(&lookup_spans(&body.var_types, &body.spans)),
         // expressions
         label_in_non_overlapping_groups(&lookup_spans(&body.expressions, &body.spans)),
         // emits
--- a/README.md
+++ b/README.md
@@ -90,8 +90,7 @@ The most advanced example that we can provide at this point is an implementation
 
 The Fe implementation is split into several crates. Crates that depend on the
 solidity compiler (directly or indirectly) are licensed GPL-3.0-or-later. This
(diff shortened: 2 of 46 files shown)


Y-Nak (approved):
Seems nice!

In addition to the changes that you've made, we could remove https://github.com/ethereum/fe/blob/43b58a8d28e07c1bdeab5a8f5e34df73fdf08464/crates/analyzer/src/context.rs#L380 and https://github.com/ethereum/fe/blob/43b58a8d28e07c1bdeab5a8f5e34df73fdf08464/crates/analyzer/src/namespace/scopes.rs#L196

`var_decl_types` are referred by `fe-lowering`, but it's no longer used. Also it wrongly stores type annotation spans as its key, instead of variables spans.
So I think it'd be better to remove them in favor of https://github.com/ethereum/fe/blob/43b58a8d28e07c1bdeab5a8f5e34df73fdf08464/crates/analyzer/src/context.rs#L376

Of course, it's completely fine if you merge this PR as is. 
I like mindless work and I'll do it after mering the PR 😄

crates/analyzer/tests/analysis.rs
```diff
@@ -415,8 +415,6 @@ fn function_diagnostics(fun: items::FunctionId, db: &dyn AnalyzerDb) -> Vec<Diag
     [
         // signature
         build_debug_diagnostics(&[(fun.data(db).ast.span, &fun.signature(db))]),
-        // declarations
-        label_in_non_overlapping_groups(&lookup_spans(&body.var_decl_types, &body.spans)),
```
Y-Nak, line 419: How about keeping this line? 
It may be helpful when a type inference is implemented.
```
let my_array: Array<u256, 10>
    ^^^^^^^^ Array<u256, 10>
```

Y-Nak (approved):
👍 
Left one nit.

PR #729, opened 2022-06-01
Support encoding/decoding nested struct types

### LINKED ISSUE #635: Support returning complex structs from public functions
### What is wrong?

Complex structs (structs where the type of at least one field is not a base type) can't yet be returned from public functions.

### How can it be fixed

We need to implement ABI encoding for these types of structs.

### LINKED ISSUE #733: Allow nested struct type arguments in public functions
### What is wrong?
The older yulgen doesn't support functions that have nested struct arguments.
Also, new codegen doesn't support them.

### How can it be fixed
Properly decode nested struct types.

Y-Nak:
### What was wrong?
close #635, close #733

### How was it fixed?
Implement encoding/decoding runtime functions for nested struct types.

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
Hot stuff :fire: 
LGTM

Issue #724, opened 2022-05-27
Reject uninitialized variable


Y-Nak:
### What is wrong?
Currently, we don't perform analysis for uninitialized variable use. This could potentially lead to unintentional error.
We 

### How can it be fixed
* [ ] Add an array repeat expression `[Value; N] to make an array creation easier.
* [ ] Perform the analysis on MIR.

PR #731, opened 2022-06-02
memory allocation bugs

g-r-a-n-t:
### What was wrong?

Similar to https://github.com/ethereum/fe/pull/684.

I should look over all uses of `avail` and `alloc`.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/codegen/src/yul/runtime/contract.rs
+++ b/crates/codegen/src/yul/runtime/contract.rs
@@ -26,7 +26,7 @@ pub(super) fn make_create(
     let func = function_definition! {
         function [func_name.ident()]([value.ident()]) -> addr {
             (let [size.ident()] := datasize([contract_symbol.clone()]))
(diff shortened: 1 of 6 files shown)


Y-Nak:
I've been considering this and now think that it'd be better to resolve this problem in #724.
When we switch to sonatina, this problem would arise again because it'll reuse allocated memory again and again.
We could resolve this by always zeroing memory when a variable is declared but this will cause performance degradation.

Y-Nak:
But we should fix dynamic type encoding functions, where I wrongly assumed that allocated space for encoding is always zeroed. e.g., padding bytes should be explicitly zeroed here. https://github.com/ethereum/fe/blob/7621ff9ab197cc35e5e363475de6c45f7ac1870a/crates/codegen/src/yul/runtime/abi.rs#L139-L147

g-r-a-n-t:
see #747

PR #734, opened 2022-06-04
Allow using a complex type as an array element type

### LINKED ISSUE #730: Allow using complex type as array element type
### What is wrong?
Currently, array element types are restricted to primitive types. 
Now, we can remove the restriction with the new codegen.

The test case below should be passed at least.
```
pub struct Pair {
    pub x: i32
    pub y: i32
}

contract Foo {
    pub fn static_sized_type_elem() -> Array<Pair, 2> {
        let pair1: Pair = Pair(x: 1, y: 2)
        let pair2: Pair = Pair(x: 10, y: 20)

        let arr: Array<Pair, 2> = [pair1, pair2]
        return arr
    }

    pub fn dynamic_sized_type_elem() -> Array<String<10>, u8>{
        let arr: Array<String<10>, 2> = ["Foo", "Bar"]
        return arr
    }
}
```

### How can it be fixed
This can be fixed just by removing the restriction in the `analyzer`.
NOTE: It'd be better to wait for #729 to be merged because it will allow returning nested complex types.

Y-Nak:
close #730 
1. Allow using a complex type as an array element type
2. Fix a bug that incorrect instruction is selected when the operands of a comp instruction are a signed type.

Example: 
```
contract Foo {
    pub fn bar() -> i256 {
        let my_array: Array<Pair, 3> = [Pair::new(-1, 1), Pair::new(10, -10), Pair::new(-100, 100)]

        let sum: i256 = 0
        for pair in my_array {
            sum += pair.l1_norm()
        }

        return sum
    }
}

struct Pair {
    pub x: i256
    pub y: i256

    pub fn new(_ x: i256, _ y: i256) -> Pair {
        return Pair(x, y)
    }

    pub fn l1_norm(self) -> i256 {
        return abs(self.x) + abs(self.y)
    }
}

fn abs(_ val: i256) -> i256 {
    if val < 0 {
        return -val
    } else {
        return val
    }
}
```

NOTE: 
Please start reviewing this after #729 is merged.

- [x] Clean up commit history

### DIFF
--- a/crates/abi/src/types.rs
+++ b/crates/abi/src/types.rs
@@ -59,16 +59,15 @@ impl AbiType {
     pub fn header_size(&self) -> usize {
         match self {
             Self::UInt(_) | Self::Int(_) | Self::Address | Self::Bool | Self::Function => 32,
-            Self::Array { elem_ty, len } => elem_ty.header_size() * len,
-            Self::Tuple(fields) => {
-                if self.is_static() {
-                    fields
-                        .iter()
-                        .fold(0, |acc, field| field.ty.header_size() + acc)
-                } else {
-                    32
-                }
-            }
+
+            Self::Array { elem_ty, len } if elem_ty.is_static() => elem_ty.header_size() * len,
+            Self::Array { .. } => 32,
+
+            Self::Tuple(fields) if self.is_static() => fields
+                .iter()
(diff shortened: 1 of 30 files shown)


sbillig (approved):
Looks good! Sorry for the delay.

PR #738, opened 2022-06-08
Intern analyzer Types (TypeId)

sbillig:
### What was wrong?

The analyzer `Type` is clone-only, because it contains a vec and a box. It should be copyable. Also `Struct` and `Contract` contain redundant copies of their names.

### How was it fixed?

Type::Struct and ::Contract now just contain ids. All types are interned to get a TypeId. `Type` fields containing things like `Box<Type>` are now contain `TypeId`s instead. Most functions that took `Type` now take `TypeId`. Some of this is messy. I might go through and tidy up a bit, but I'm also quite sick of this so maybe not.

This means that `Type` is no longer `Display`able, except via the new `.display(db)` method. So we traded off `.clone()` for `.display(db)`. Was it worth it? I dunno.

All the analysis test snapshots changed, because Debug printing of `Type` is now useless (ids), so now FunctionSignature printing is terse, and the `emit` event debug print is gone. It'll magically reappear (sort of) when `ctx.emit(my_event)` is implemented.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/analyzer/src/db/queries/functions.rs
+++ b/crates/analyzer/src/db/queries/functions.rs
@@ -1,11 +1,12 @@
 use crate::context::{AnalyzerContext, CallType, FunctionBody};
 use crate::db::{Analysis, AnalyzerDb};
+use crate::display::Displayable;
 use crate::errors::TypeError;
 use crate::namespace::items::{
     Class, DepGraph, DepGraphWrapper, DepLocality, FunctionId, FunctionSigId, Item, TypeDef,
 };
 use crate::namespace::scopes::{BlockScope, BlockScopeType, FunctionScope, ItemScope};
-use crate::namespace::types::{self, Contract, CtxDecl, Generic, SelfDecl, Struct, Type};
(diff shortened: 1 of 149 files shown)


Y-Nak (approved):
LGTM 👍

Y-Nak:
I agree with you; it'd be better to just intern every type for simplicity in the near future.

crates/analyzer/src/db/queries/functions.rs
```diff
@@ -213,15 +214,10 @@ pub fn function_signature(
...
                 }
-                Ok(Type::unit())
+                Ok(TypeId::unit(scope.db()))
             } else {
                 match type_desc(&mut scope, type_node)? {
-                    Type::Struct(val)
-                        if val.id.has_complex_fields(db) && function.is_public(db) =>
-                    {
-                        Ok(Type::Struct(val))
-                    }
```
sbillig, line 223: This seemed like pointless leftover code, but maybe I'm mistaken?

PR #744, opened 2022-06-17
Treat `trait` exclusively as `Item` not `Type`

cburgdorf:
### What was wrong?

This is a follow up to #710 which defined `Type::Trait`. Traits are currently only supported as bounds on generics. Since for the foreseeable future we do not plan to support traits in places where types are supported we should remove it from `Type` and instead create an `Item::Trait(TraitId)`

### How was it fixed?

Pretty much what the description already spoiled.


sbillig (approved):
:+1:

Issue #745, opened 2022-06-20
Usage of negative i8 in const leads to ICE


cburgdorf:
### What is wrong?

The following leads to an ICE

```
const INIT_VAL: i8 = -1

contract Foo {
  pub fn init_bar() {
    let x: i8 = INIT_VAL
  }
}
```

### How can it be fixed

The `-1` is taken as a literal `-1`. It should be replaced with the corresponding hex value.

PR #746, opened 2022-06-20
Fix ICE with negative const

cburgdorf:
### What was wrong?

The following leads to an ICE

```
const INIT_VAL: i8 = -1

contract Foo {
  pub fn init_bar() {
    let x: i8 = INIT_VAL
  }
}
```

### How was it fixed?

YUL doesn not allow literals with minus symbol (e.g. `-1`) so instead we now turn the constants into hexadecimal representation.


Y-Nak:
close #745

PR #747, opened 2022-06-22
require initialization of structs and arrays at decl site

g-r-a-n-t:
### What was wrong?

see #731

### How was it fixed?

Added array repeat expressions and checks for initialization of structs and arrays.

### To-Do

- [x] Clean up commit history


crates/analyzer/src/traversal/declarations.rs
```diff
@@ -31,6 +31,22 @@ pub fn var_decl(scope: &mut BlockScope, stmt: &Node<fe::FuncStmt>) -> Result<(),
...
+        } else {
+            if matches!(declared_type.typ(scope.db()), Type::Array(_)) {
+                scope.error(
+                    "uninitialized variable",
+                    target.span,
+                    "array types must be initialized at declaration site",
+                );
+            }
+
+            if matches!(declared_type.typ(scope.db()), Type::Struct(_)) {
```
cburgdorf, line 43: Doesn't the same apply to tuples?

crates/analyzer/tests/snapshots/analysis__aug_assign.snap
```diff
@@ -264,7 +264,7 @@ note:
    ┌─ aug_assign.fe:65:5
    │  
 65 │ ╭     pub fn add_from_mem(a: u256, b: u256) -> u256 {
-66 │ │         let my_array: Array<u256, 10>
+66 │ │         let my_array: Array<u256, 10> = [0; 10]
```
cburgdorf, line 267: Adding this repeat expression feels a bit like a crutch to me. It doesn't seem overly useful for the user since it only allows a single value to be repeated. Did you consider to just initialize the array with default values on behalf of the user. Maybe that gets complicated if the array is over a complex type such as `Array<Customer, 10>`. However, in the future we could have a `Default` trait and if a user wants to initialize an array of `Array<Customer, 10>` then it will automatically use the values provided by `Customer::default()` or else error if `Customer` doesn't implement the `Default` trait. I'm just thinking out loud here...I'm not really against introducing the repeat expression but if in the future we find a way to get rid of it again or make it more useful I wouldn't be sad to see it go :smile:

crates/mir/src/lower/function.rs
```diff
@@ -668,6 +668,25 @@ impl<'db, 'a> BodyLowerHelper<'db, 'a> {
...
+                {
+                    array_type
+                } else {
+                    panic!("not an array");
+                };
+
+                let args = (0..array_type.size)
+                    .into_iter()
+                    .map(|_| self.lower_expr_to_value(value))
+                    .collect();
```
Y-Nak, line 685: Should an expression for array elements value be evaluated N times? At least, Rust evaluates it just once. I think Rust can do this because a type in the array repetition initialization must implement `Copy`. Also, we could assume that all types in `Fe` implements `Copy`. So it would be fine to evaluate it just once, and it would greatly reduce gas usage when the expression is expensive.
But from a language-design perspective, I'm not sure which is easier for users to understand the semantics. It might be difficult for those who are addicted to list comprehension. On the other hand, I wonder a list comprehension might be confusing when the language allows side effects.

@sbillig, @cburgdorf  I pinged you two because this discussion would affect the language specification.

cburgdorf (approved):
Left some inline comments. Most important: Can you confirm we don't need that for tuple types? Wouldn't something like `let val: (u8, u8, address)` run into the same issue?

crates/analyzer/tests/snapshots/analysis__aug_assign.snap
cburgdorf: Ok, I just learned that one can use the same approach to initialize fixed size arrays in Rust. So, I guess it's fine. I still think that in the future it could be nice syntactical sugar if one could just write `let x: Array<Customer, 10>` and it would de-sugar into something like `let x: Array<Customer, 10> = [Customer::default(), 10]`. And if `T` does not implement the `Default` trait it would cause a compile error.

crates/mir/src/lower/function.rs
g-r-a-n-t: I'll change it to evaluate once

PR #750, opened 2022-06-23
Fix duplicated block in control flow uplifting

### LINKED ISSUE #749: Certain code with if statement leads to an ICE
### What is wrong?

The following leads to an ICE

```rust
contract Bar {

    something: bool

    pub fn move_field(self) {
        if true {
            if self.something { 
                return 
            }
        }
        if true {
            if self.something { 
                return 
            }
        }
    }
}

```

ICE:

```
Error: DeclarationError: Variable not found or variable not lvalue.
  --> input.yul:33:17:
   |
33 |                 $$tmp_5 := and($$sptr_load(add(0, 0), 8), 0xff)
   |                 ^^^^^^^


Error: DeclarationError: Identifier "$$tmp_5" not found.
  --> input.yul:34:24:
   |
34 |                 switch $$tmp_5
   |                        ^^^^^^^

```

Looking at the relevant YUL we can see that the second `$$tmp_5` is used but not declared. Or maybe the initial declaration of `$$tmp_5` should have been made one block higher :thinking: I haven't really analyzed it yet.

```
switch 1
case 1 {
    let $$tmp_3
    $$tmp_3: = and($$sptr_load(add(0, 0), 8), 0xff)
    switch $$tmp_3
    case 1 {
        leave
    }
    case 0 {
        switch 1
        case 1 {
            let $$tmp_5
            $$tmp_5: = and($$sptr_load(add(0, 0), 8), 0xff)
            switch $$tmp_5
            case 1 {
                leave
            }
            case 0 {
                leave
            }
        }
        case 0 {
            leave
        }
    }
}
case 0 {
    switch 1
    case 1 {
        $$tmp_5: = and($$sptr_load(add(0, 0), 8), 0xff)
        switch $$tmp_5
        case 1 {
            leave
        }
        case 0 {
            leave
        }
    }
    case 0 {
        leave
    }
}
}
```

It's as minimal as I could get it:

The following alternations would make the ICE go away:

- removing the `return`
- unnesting
- removing the second (duplicated) if block
- using a local condition instead of `self.something`

### How can it be fixed

:hocho:

Y-Nak:
### What was wrong?
close #749 
The problem is that the same basic block is included in serialized `StructuralInst` and then translated into yul more than once in the below CFG condition.
1. CFG doesn't have a canonical exit block
2. Two paths can be merged at a certain block

e.g.,
graphviz
In this CFG, `bb2` is a `merge` block of two paths, `bb0 -> bb1 -> bb4 -> bb2` and `bb0 -> bb2`.
But the compiler can't find merge block properly, and it lead to ICE as a result.
### How was it fixed?
I wrongly assumed that `if` merge block should be the post dominator of `if` header.
To find `if` merge block correctly, we should follow the below algorithm.

#### 1. Find`if` merge block candidate
When we see the `branch` instruction in a non-loop header, we can find a merge block candidate by applying the following criteria to two destination blocks of the `branch` instruction respectively.

```
block `cand` can be a candidate of a `merge` block iff
  1. `cand` is a dominance frontier of `block`.
  2. `cand` is NOT  a dominator of `block`.
  3. `cand` is NOT a "merge" block of a parent `if`.
  4. `cand` is NOT a "loop_exit" block of parent `loop`.
```

#### 2. Find `merge` block.
We can find the merge block according to the result of `1.`.

```
a. When both destinations have the same candidate, then the candidate is the merge block.

b. When only one of the destinations has a candidate, then the candidate is the merge block iff the candidate is identical to another destination that doesn't have a candidate.

c. When none of the destinations has a candidate, then a `block` is the merge block iff 
  1. the `block` is the post dominator of the `branch` instruction's block
  2. the `block` is inside the same loop if the `branch` instruction's block is in a loop.
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/codegen/src/yul/isel/inst_order.rs
+++ b/crates/codegen/src/yul/isel/inst_order.rs
@@ -27,6 +27,7 @@ pub(super) struct InstSerializer<'a> {
     cfg: ControlFlowGraph,
     loop_tree: LoopTree,
     df: DFSet,
+    domtree: DomTree,
     pd_tree: PostDomTree,
     scope: Option<Scope>,
 }
@@ -44,6 +45,7 @@ impl<'a> InstSerializer<'a> {
             cfg,
             loop_tree,
             df,
+            domtree,
             pd_tree,
             scope: None,
         }
@@ -125,7 +127,6 @@ impl<'a> InstSerializer<'a> {
         let mut else_body = vec![];
 
         self.enter_if_scope(merge_block);
-
         let mut serialize_dest =
             |dest_info, body: &mut Vec<StructuralInst>, merge_block| match dest_info {
                 TerminatorInfo::Break => body.push(StructuralInst::Break),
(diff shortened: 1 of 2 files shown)


cburgdorf (approved):
Looks good!

PR #755, opened 2022-06-28
Allow traits to be implemented for almost all types

cburgdorf:
### What was wrong?

Currently traits can only be implemented for structs. The reason for that is that prior to monomorphization we do not know the `Location` of a type generic type (e.g. `T`) and for that reason hard coded it to `Location::Memory` which would exclude primitive types such as `u8`.

### How was it fixed?

I made attempts to get rid of `Location` in the analyzer entirely but I kept failing. Then I realized that we actually don't yet allow to assign generic arguments anyway. Since the location is only relevant for when variables are assigned we could have used `Location::Value` and it would have worked exactly the same, it's superfluous information for now.
Hence, I thought we could introduce a `Location::Unresolved` for now and in return lift the restriction to allow traits to be implemented for all base types, arrays and tuples. 

In addition this fixes two ICEs related to calling `to_mem()` or `clone()` on a generic type.

E.g.

```
trait Double {
  fn double(self) -> u256;
}

impl Double for u8 {
  fn double(self) -> u256 {
    return u256(self + self)
  }
}

impl Double for u16 {
  fn double(self) -> u256 {
    return u256(self + self)
  }
}

impl Double for (u256, u256) {
  fn double(self) -> u256 {
    return u256(self.item0 + self.item0 + self.item1 + self.item1)
  }
}

struct Runner {
  pub fn run<T: Double>(self, _ val: T) -> u256 {
    return val.double()
  }
}


contract Example {
  pub fn generic_compute(self) {
    let runner: Runner = Runner();
    assert runner.run(u8(1)) == 2
    assert runner.run(u16(1000)) == 2000
    assert runner.run((1, 1)) == 4
  }
}
```

### DIFF
--- a/crates/analyzer/src/context.rs
+++ b/crates/analyzer/src/context.rs
@@ -350,6 +350,8 @@ pub enum Location {
     },
     Memory,
     Value,
+    // An unresolved location is used for generic types prior to monomorphization.
+    Unresolved,
 }
 
 impl Location {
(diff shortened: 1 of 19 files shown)


sbillig (approved):
:+1: :sunglasses: :+1:

PR #762, opened 2022-07-06
Allow trait methods to be called on non-generic types if trait is in scope

cburgdorf:
### What was wrong?

Traits can't currently be used without generics but they should be.

### How was it fixed?

Example of what the PR enables:

```rust
trait Double {
  fn double(self) -> u256;
}

impl Double for (u256, u256) {
  fn double(self) -> u256 {
    return (self.item0 + self.item1) * 2
  }
}

contract Example {

  pub fn run_test(self) {
    assert (0, 1).double() == 2
  }
}
```

Some notes:

- If a call turns out to be ambigious the compiler currently asks the user to disambiguate via renaming. In the future we will likely introduce a syntax to allow to disambiguate at the callsite.
- refactored to get rid of `Class` entirely (it's not too useful as a concept since via `impl` and traits any type can have member functions)
- Ensured we fetch all impls even from foreign ingots. E.g. ingot `Foo` may have a struct `Thingy` and a trait `SuperPower` with an `impl` for Thingy. If one imports and uses `Thingy` then we still have to check if `SuperPower` is in scope for `thingy.super_power_move()` to work.
- All the `impl` handling was removed from `Struct` since it is now generalized on `TypeId` directly

### DIFF
--- a/crates/analyzer/src/namespace/types.rs
+++ b/crates/analyzer/src/namespace/types.rs
@@ -114,11 +114,10 @@ impl TypeId {
         }
     }
 
+    /// Return the `impl` for the given trait. There can only ever be a single implementation
+    /// per concrete type and trait.
     pub fn get_impl_for(&self, db: &dyn AnalyzerDb, trait_: TraitId) -> Option<ImplId> {
-        match self.typ(db) {
-            Type::Struct(id) => id.get_impl_for(db, trait_),
-            _ => trait_.module(db).impls(db).get(&(trait_, *self)).cloned(),
-        }
+        db.impl_for(*self, trait_)
     }
 
     // Returns all `impl` for the type even from foreign ingots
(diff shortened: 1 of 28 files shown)


crates/analyzer/src/traversal/expressions.rs
```diff
@@ -1726,184 +1800,186 @@ fn expr_call_type_attribute(
...
 
-            // Check for visibility of callee.
-            // Emits an error if callee is not public and caller doesn't belong to the same
-            // class as callee.
-            if !function.is_public(context.db()) && is_same_class(class) {
-                context.fancy_error(
+        // Check for visibility of callee.
+        // Emits an error if callee is not public and caller doesn't belong to the same
+        // class as callee.
+        if !sig.is_public(context.db()) && is_same_class(target_type.typ(context.db())) {
```
sbillig, line 1937: Presumably this should be `!is_same_class(..)`? (I guess this is an uncaught mistake from a previous change)

sbillig (approved):
Looks good!

crates/analyzer/src/traversal/expressions.rs
cburgdorf: Nice catch. Unfortunately, just flipping the condition doesn't cut it here. I'm not sure if the code is exercised at all anymore. It inspired me to file https://github.com/ethereum/fe/issues/765

I will leave this as is to not bloat the PR but I'll investigate it on Monday (I'm a bit short on time today)

crates/analyzer/src/namespace/types.rs
```diff
@@ -95,15 +99,68 @@ impl TypeId {
...
+            .iter()
+            .filter_map(|impl_| impl_.function(db, name))
+            .map(|fun| fun.sig(db))
+            .collect();
+
+        if let Some(fun) = native {
+            fns.push(fun)
+        }
+        fns
+    }
```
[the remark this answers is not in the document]
cburgdorf: Makes sense. Fixed.

crates/analyzer/src/traversal/expressions.rs
cburgdorf: Fixed in https://github.com/ethereum/fe/pull/767

PR #767, opened 2022-07-11
Fix enforcement of visibility rules of function calls

cburgdorf:
### What was wrong?

The code that validates the visibility rules of function calls is duplicated, faulty and incomple.

### How was it fixed?

1. Created a single function `validate_visibility_of_called_fn` that can be called from any place that needs to enforce the rules
2. Ensured the rules are enforced for calls to associated functions
3. Added a bunch more tests
4. Bonus: Made it possible to use `MyContract::static_func()` syntax which wasn't yet possible for contracts (but worked for structs)
5. More test coverage

### DIFF
--- a/crates/analyzer/src/namespace/items.rs
+++ b/crates/analyzer/src/namespace/items.rs
@@ -814,26 +814,17 @@ pub enum TypeDef {
     Contract(ContractId),
     Primitive(types::Base),
 }
+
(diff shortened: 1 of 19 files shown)


sbillig (approved):
:+1: Thanks for fixing everything while the rest of us twiddle our thumbs.

PR #768, opened 2022-07-12
Remove Location in favor of Type::SPtr

sbillig:
### What was wrong?

With mir and the new codegen, the analyzer's `Location` tracking is mostly useless.

### How was it fixed?

Added `Type::SPtr(TypeId)`, which is the type of all contract `self` fields.


crates/mir/src/lower/types.rs
```diff
@@ -8,9 +8,13 @@ use crate::{
 
 use fe_analyzer::namespace::{items as analyzer_items, types as analyzer_types};
 
-pub fn lower_type(db: &dyn MirDb, analyzer_ty: &analyzer_types::TypeId) -> TypeId {
+pub fn lower_type(db: &dyn MirDb, analyzer_ty: analyzer_types::TypeId) -> TypeId {
     let ty_kind = match analyzer_ty.typ(db.upcast()) {
-        analyzer_types::Type::Base(base) => lower_base(&base),
+        analyzer_types::Type::SPtr(inner) => {
+            // TODO: lower to SPtr(lower_type(db, inner).
```
sbillig, line 14: I attempted this, but ended up with an SPtr(SPtr(Map)) type that I couldn't easily find the source of.

Y-Nak: what code ended up with the nested `SPtr`?
I tried the code below and it seems it returns collect MIR type `SPtr<Map<i32, i32>>`
```fe
contract Foo {
    m: Map<i32, i32>
    
    fn test(self) {
        self.m
    }
}
```
graphviz (2)

sbillig:
Superseded by #781

sbillig:
Jk

sbillig:
I added a Load instruction to MIR. A Load is inserted in `fn lower_expr_to_value` if the type is a ptr to a primitive. Might not be a perfect heuristic, but it works. I'm open to changing any of this, of course.

This accidentally fixes some gas inefficiency in the yul codegen. Eg in guest_book.fe:
```
    pub fn get_msg(self, addr: address) -> String<100> {
        return self.messages[addr].to_mem()
    }
```
before:
```
            function guest_book$GuestBook$get_msg($addr) -> $$ret {
                let $$tmp_3
                $$tmp_3 := add(0, 0)
                let $$tmp_4 := $$alloc(132)
                $scopym($$map_value_ptr_with_primitive_key($$tmp_3, $addr), $$tmp_4, 132)
                let $$tmp_5 := $$alloc(132)
                $mcopym($$tmp_4, $$tmp_5, 132)
                $$ret := $$tmp_5
                leave
            }
```
after:
```
            function guest_book$GuestBook$get_msg($addr) -> $$ret {
                let $$tmp_3
                $$tmp_3 := add(0, 0)
                let $$tmp_4
                $$tmp_4 := $$map_value_ptr_with_primitive_key($$tmp_3, $addr)
                let $$tmp_5 := $$alloc(132)
                $scopym($$tmp_4, $$tmp_5, 132)
                $$ret := $$tmp_5
                leave
            }
```

crates/mir/src/lower/function.rs
```diff
@@ -726,7 +729,15 @@ impl<'db, 'a> BodyLowerHelper<'db, 'a> {
     fn lower_expr_to_value(&mut self, expr: &Node<ast::Expr>) -> ValueId {
         let ty = self.expr_ty(expr);
         let inst = self.lower_expr(expr);
-        self.map_to_tmp(inst, ty)
+        let value = self.map_to_tmp(inst, ty);
+
+        // Load primitive types onto stack
+        if ty.is_ptr(self.db) && ty.deref(self.db).is_primitive(self.db) {
+            let inst = self.builder.load(value, expr.into());
+            self.map_to_tmp(inst, ty.deref(self.db))
```
Y-Nak, line 737: I think the `load` insertion should be performed in the `lower_expr` or another transformation pass because `lower_expr_to_value` is not always called and the mapping `value` to `inst` can be performed manually in some AST -> MIR lowering. e.g, `Assign` lowering.

So as a result, the below example is lowered into the inconsistent MIR regarding `SPtr` handling.
```
contract Foo {
    s: Array<i32, 20>

    pub fn test(self) -> i32 {
        let elem: i32 = self.s[0]
        return elem
    }
}
```
↓

graphviz

This inconsistency is not a problem for now because the ugly heauristic in `codegen::yul::isel::function::assign_inst_result` handles this.

But this inconsistency is error-prone, so I think it's better to address this in this PR.
Do you think it's worth adding another transformation pass to perform these heuristics after lowering?

sbillig:
Rolled into #777; maintaining separate PRs was a pain.

PR #769, opened 2022-07-14
zero padding on bytes and string encodings

g-r-a-n-t:
### To-Do

- [x] Clean up commit history


Y-Nak (approved):
LGTM!

Issue #717, opened 2022-05-23
Remove `event` type and keyword


Y-Nak:
Currently, we use `event` type defined with `event` keyword to emit an event.
But actually, we could handle `event` types in the same way as types defined with the `struct` keyword, i.e., normal user-defined struct. 
The only difference between `event` and `struct` would be `topic`, so it would be necessary to introduce some special syntax or language features that enable it to be specified.

Currently, I'd like to propose allowing `@[indexed]` or `@[topic]` attribute in the field like below.
```python
struct MyEvent:
    @[indexed]
    x: i32

    y: i32
```

In the short-term plan, we could treat `@[indexed]` attribute specially, then generalize the attribute feature in the long-term plan.

distractedm1nd:
What effect does this have (or would it have in the future) on indexed fields?

cburgdorf:
>In the short-term plan, we could treat @[indexed] attribute specially, then generalize the attribute feature in the long-term plan.

I agree we should do this. This is basically the only thing holding us back from adding an `emit<T: Event>(val: T) {..}` message directly on `context.fe` in the std lib.

cburgdorf:
> What effect does this have (or would it have in the future) on indexed fields?

I think the main advantages of treating events as structs are:

1. Less special syntax
2. Code re-use (e.g. being able to use the a `Foo` struct for foo things and just happen to also emit it as an event)
3. Flexibility. If events are just structs, chances are people can come up with nice little helper libraries down the road

Y-Nak:
#772

PR #772, opened 2022-07-27
Remove event/emit and introduce ctx.emit that works with the Emittable trait

cburgdorf:
### What was wrong?

As described in #717 we want to remove the `event` type and the `emit` keyword and make structs usable as events via `ctx.emit<T: Emittable>(val: T)`

### How was it fixed?

- Removed `event` type and `emit` statement
- Added support for `#indexed` attribute on struct fields
- Introduced `Emittable` trait which is automatically implemented for all structs
  - the trait is not implementable manually (it therefore depends on a private type)
  - the compiler still implements the logic behind the `emit` trait function internally (just like before)
 - updated all tests and code examples


cburgdorf:
I have no idea where this "CircleCI Pipeline" job comes from. We don't use CircleCI :shrug:

cburgdorf:
This is now rebased on top of latest master

Y-Nak (commented):
I read through the code and thought we should discuss the whole design before getting into the details.

Stopgap solutions should not affect a large part of a codebase, but this implementation seems to do so.
I think it would be enough to add a single small module that provides
1. semantic checks for the `event` features like the number of `indexed` fields
2. Automatic implementation of the `Emittable` trait to structs

What's in my mind is this module should essentially be a temporary alternative to the actual `event` procedural macro implementation until we provide a proc-macro-like feature, and the module will simply be removed when the day comes instead of re-modifying the various parts of the analyzer.

crates/analyzer/src/namespace/items.rs
```diff
@@ -1891,6 +1878,11 @@ impl TraitId {
     }
 
     pub fn is_implemented_for(&self, db: &dyn AnalyzerDb, ty: TypeId) -> bool {
+        // All structs automagically implement the Emittable trait
+        if self.is_std_trait(db, EMITTABLE_TRAIT_NAME) && ty.is_struct(db) {
+            return true;
+        }
```
Y-Nak, line 1884: This breaks the invariant of a group of methods related to trait implementation because this function returns `true` even if the actual implementation doesn't exist. So all other codes which depend on the method would need additional checks to ensure the actual implementation exists.

crates/codegen/src/db/queries/abi.rs
```diff
@@ -27,8 +27,9 @@ pub fn abi_contract(db: &dyn CodegenDb, contract: ContractId) -> AbiContract {
     }
 
     let mut events = vec![];
-    for &event in db.contract_all_events(contract).as_ref() {
-        let mir_event = db.mir_lowered_event_type(event);
+    // We consider all structs that are defined or imported in the scope of the contract as possible events
+    for &event in db.module_structs(contract.module(db.upcast())).as_ref() {
+        let mir_event = db.mir_lowered_type(event.as_type(db.upcast()));
```
Y-Nak, line 32: This is `#event/#indexed` specific.

crates/codegen/src/db/queries/abi.rs
```diff
@@ -207,21 +207,22 @@ pub fn abi_type(db: &dyn CodegenDb, ty: TypeId) -> AbiType {
...
 
     let fields = event_def
         .fields
         .iter()
-        .map(|(name, ty, indexed)| {
+        .map(|(name, attr, ty)| {
             let ty = db.codegen_abi_type(*ty);
-            AbiEventField::new(name.to_string(), ty, *indexed)
+            let indexed = attr.iter().any(|attr| attr == INDEXED);
+            AbiEventField::new(name.to_string(), ty, indexed)
```
Y-Nak, line 225: ditto.

crates/codegen/src/db/queries/types.rs
```diff
@@ -37,11 +37,11 @@ pub fn legalized_type(db: &dyn CodegenDb, ty: TypeId) -> TypeId {
                 .fields
                 .iter()
                 .cloned()
-                .filter_map(|(name, ty)| {
+                .filter_map(|(name, attrs, ty)| {
```
Y-Nak, line 40: ditto

crates/codegen/src/db/queries/types.rs
```diff
@@ -37,11 +37,11 @@ pub fn legalized_type(db: &dyn CodegenDb, ty: TypeId) -> TypeId {
...
                     if ty.is_zero_sized(db.upcast()) {
                         None
                     } else {
-                        Some((name, legalized_type(db, ty)))
+                        Some((name, attrs, legalized_type(db, ty)))
```
Y-Nak, line 44: ditto

crates/codegen/src/db/queries/types.rs
```diff
@@ -54,38 +54,16 @@ pub fn legalized_type(db: &dyn CodegenDb, ty: TypeId) -> TypeId {
...
-            TypeKind::Event(new_def)
-        }
-
         TypeKind::Contract(def) => {
             let fields = def
                 .fields
                 .iter()
                 .cloned()
-                .filter_map(|(name, ty)| {
+                .filter_map(|(name, attrs, ty)| {
```
Y-Nak, line 62: ditto

crates/codegen/src/db/queries/types.rs
```diff
@@ -54,38 +54,16 @@ pub fn legalized_type(db: &dyn CodegenDb, ty: TypeId) -> TypeId {
...
                     if ty.is_zero_sized(db.upcast()) {
                         None
                     } else {
-                        Some((name, legalized_type(db, ty)))
+                        Some((name, attrs, legalized_type(db, ty)))
```
Y-Nak, line 66: ditto

crates/mir/src/lower/function.rs
```diff
@@ -946,28 +944,29 @@ impl<'db, 'a> BodyLowerHelper<'db, 'a> {
...
                     .call(func_id, method_args, CallType::Internal, source)
             }
+            AnalyzerCallType::TraitValueMethod {
+                trait_id, method, ..
+            } if trait_id.is_std_trait(self.db.upcast(), EMITTABLE_TRAIT_NAME)
+                && method.name(self.db.upcast()) == EMIT_FN_NAME =>
+            {
+                let event = self.lower_method_receiver(func);
+                self.builder.emit(event, source)
+            }
```
Y-Nak, line 969: This is `#event/#indexed` specific.

crates/mir/src/ir/types.rs
```diff
@@ -67,7 +66,7 @@ pub struct TupleDef {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct StructDef {
     pub name: SmolStr,
-    pub fields: Vec<(SmolStr, TypeId)>,
+    pub fields: Vec<(SmolStr, Vec<SmolStr>, TypeId)>,
```
Y-Nak, line 69: I don't think we should add attributes to MIR types.

Y-Nak (commented):
I enumerated the `#event/#indexed` specific code in the change. It seems still scattered for me. Also, almost all changes happen in the public functions, which means they potentially affect the broader part of the codebase.

crates/mir/src/db/queries/structs.rs
```diff
@@ -11,7 +11,7 @@ pub fn mir_lower_struct_all_functions(
     struct_
         .all_functions(db.upcast())
         .iter()
-        .map(|func| db.mir_lowered_func_signature(*func))
+        .map(|func| db.mir_lowered_pseudo_monomorphized_func_signature(*func))
```
Y-Nak, line 14: It seems weird to use the function because it is annotated as `NOTE: THIS SHOULD ONLY BE USED IN TEST CODE`.

crates/library/std/src/context.fe
```diff
@@ -5,6 +5,15 @@ use ingot::error::{
...
 }
 
+struct OutOfReachMarker {}
+
+// ctx.emit(my_event) should be the only way to emit an event. We achieve this by defining the
+// private `OutOfReachMarker` here to which only the `Context` has access.
+// Now there is no way to call `emit` directly on an Emittable.
+pub trait Emittable {
+  fn emit(self, _ val: OutOfReachMarker);
+}
```
Y-Nak, line 15:
```suggestion
pub trait Emittable {
  fn emit(self, _ ctx: Context);
}
```
I think exposing private types from a public function is not a good idea especially when it's a trait function, how about receiving `Context` instead of the `OutOfReachMarker` type?

Y-Nak (commented):
I added comments, which are not directly related to the current discussion but feel worth noting.

crates/analyzer/src/namespace/items.rs
cburgdorf: >even if the actual implementation doesn't exist

Can you elaborate on that? The method is implemented by the compiler (compiler magic basically) for any struct (well, maybe we need to restrict it a bit more because as you hinted structs with enums should maybe not implement `Emittable`). For any `Emittable` the compiler ensures that `emit` can be called.

crates/mir/src/db/queries/structs.rs
cburgdorf: Ah, yes...I guess I forgot that this is also used by the graphviz stuff. I guess what I meant to say is that it should not be used in the actual compilation pipeline. The reason it is needed at all is because the std lib test start failing as soon as the std lib included generic code because the std lib tests iterate over these functions without them being actually called which means we can not monomorphize the generic parameters. So in order to be able to iterate over the functions I created a function that would populate them in a way that maps each generic param to `()`. There's probably a better way that I'm not seeing :)

crates/library/std/src/context.fe
cburgdorf: >how about receiving `Context` instead of the `OutOfReachMarker` type

I thought about that as well but that would mean that people can bypass the `ctx.emit(..)` API and instead call `MyEvent().emit(ctx)`. That may be fine (since they are still depending on the `Context` but I kinda like it more if we force them to go through `ctx.emit(..)`. But I don't have very strong feelings on this.

crates/mir/src/ir/types.rs
cburgdorf: Ah, ok. :+1:  Probably no big deal because we can lookup the analyzer type.

crates/mir/src/lower/function.rs
cburgdorf: It is, but so is the status quo with the `Event` type and the `lower_event_type` function. After all, this PR doesn't aim to change the internal handling of events it only changes the trigger from being `emit SomeEventType` to being trait based via  `Emittable.emit(..)`.

crates/analyzer/src/constants.rs
```diff
@@ -1 +1,4 @@
+pub const EMITTABLE_TRAIT_NAME: &str = "Emittable";
+pub const EMIT_FN_NAME: &str = "emit";
+pub const INDEXED: &str = "indexed";
```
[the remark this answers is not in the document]
cburgdorf: I could move these to a place closer to their usage. My reasoning for putting them here is that they are simple strings and I dislike using magic strings inline because it is easy to mistype a string without the compiler noticing (e.g. "inexed" but you can not mistype `INDEXED` without the compiler noticing it.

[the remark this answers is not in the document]
Y-Nak: Why should `INDEXED` or `EMITTABLE` be used from several modules? I think exposing constants in an arbitrary way is not good especially if you know these constants are used as a stopgap.  How long do you think these constants will remain?

crates/analyzer/src/namespace/items.rs
Y-Nak: If the method is properly implemented by the compiler,  then why are these awkward lines necessary?
Do these types have `ImplId` for the `Emittable` trait?  I'd like to expect that `ImplId` should exist when the `is_implemented_for` returns `true`.

crates/mir/src/lower/function.rs
Y-Nak: > this PR doesn't aim to change the internal handling of events

This is the point. Again, I'd like to ask how long this stopgap solution exist. And I think the core of the codebase is affected too much just to remove the superficial syntax.

cburgdorf: >Again, I'd like to ask how long this stopgap solution will remain

I don't know but I would assume it is still a while out until we can implement the actual event logic in Fe code natively.

>And I think the core of the codebase is affected too much

But wouldn't you agree that the PR actually lowered the complexity of the code base in general? I mean, look at all the event handling logic that was deleted. More code was deleted than added in this PR. It's actually very little code that was added to enable the struct + trait based event handling.

crates/mir/src/ir/types.rs
cburgdorf: Removed them.

crates/mir/src/lower/function.rs
Y-Nak: I don't think complexity could/should be evaluated by the lines of code added/deleted. And how extent the codebase is affected is almost nothing to do with the LoC added/deleted.

I think the way of abstraction introduced here makes the codebase more complicated.

> There is no ImplId because there is no impl block for it.

This inconsistency didn't exist before the PR, for example. As I mentioned, this would break the invariant of the trait-related methods should provide, and if the PR is merged as is, it will force us to care about the `Emittable` trait wherever we handle traits. You exposed the name of the `Emittable` trait as a public constant, and it proves that we should treat the trait specially even now, and in the future, we would start writing other stopgaps to adapt to the unexpected behavior of the core functions. So stopgaps will be chained and expanded.

Also, this PR "hides" the stopgap behind the pseudo abstraction. This makes it difficult to remove these stopgaps when we finally reach the "right" solution. I mean Rust compiler won't complain about the stopgaps in the form of compile errors. So, as a result, we'll have to find and remove them manually. I am not confident in remembering the exact functions and locations where stopgaps are embedded. And as I said, these stopgaps will be chained/expanded as time passes.

So, to be honest, I have to say this PR increases the complexity of the code, and that's why I strongly disagree with merging this PR as is.

cburgdorf: No worries, we can try to work out a solution when we co-work in Berlin next week. And if we can not find a satisfying solution than I can live with the current system based on `event` and `emit` until we have all the ingredients to fully implement something like the `Emittable` trait in Fe itself.

That said, I do disagree on the question of complexity. I think having the `event` type and the `emit` keyword with all of its extra checks is more complex than having a trait in the std lib that is automagically implemented by the compiler as a temporary solution. Especially since it is also ensured that no one else can try to implement it manually for the time being.

Y-Nak: I don't disagree with adding the `Emittable` trait and implementing it to some types, though I don't think it should be implemented to *all* types. What I disagree with is the way of realizing it. Again, I still think we shouldn't arbitrarily tweak the core functions on which many other parts depend.

But yeah, we will be able to discuss this face-to-face the next week, so it'd be fine to stop the discussion on GH.

Issue #712, opened 2022-05-20
Forbid inedexed zero sized type


Y-Nak:
### What is wrong?
Currently, it is allowed to mark a zero-sized type field as an indexed field, but this doesn't make sense because a zero-sized type will never be encoded.

e.g.
```
event MyEvent:
    idx x: ()
```
should be rejected.

kris524:
Hello, can I take this task? Is there any additional information I might need?

Y-Nak:
Hi @kris524! Sorry for the late response. 
Thanks for your contribution, please feel free to take this issue.

To resolve this issue, you would need to add `is_zero_sized` or a similar method to https://github.com/ethereum/fe/blob/f810961f7702c8a50300fd9dcf0acb50f098f774/crates/analyzer/src/namespace/types.rs#L38.  Then, check the indexed field types and reject them if they are zero-sized.

Please feel free to ask me any questions.

PR #774, opened 2022-07-29
Reject zero type when declare event fields

vuvoth:
### What was wrong?
https://github.com/ethereum/fe/issues/712

### How was it fixed?

Add salsa db query `is_zero_type` and check declare field data in event should not zero-sized

Special thank to @Y-Nak! 
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/analyzer/src/db/queries/events.rs
+++ b/crates/analyzer/src/db/queries/events.rs
@@ -42,9 +42,9 @@ pub fn event_type(db: &dyn AnalyzerDb, event: EventId) -> Analysis<Rc<types::Eve
             let typ = type_desc(&mut scope, typ_node).and_then(|typ| match typ {
                 typ if typ.has_fixed_size(scope.db()) => {
                     if !typ.is_zero_size(scope.db()) {
-                        return Ok(typ)
+                        Ok(typ)
                     } else {
-                        return Err(TypeError::new(scope.error(
+                        Err(TypeError::new(scope.error(
                             "event field type must have a non-zero types",
                             typ_node.span,
                             "this type can't be used as an event field",
--- a/crates/analyzer/src/db/queries/events.rs
+++ b/crates/analyzer/src/db/queries/events.rs
@@ -40,7 +40,17 @@ pub fn event_type(db: &dyn AnalyzerDb, event: EventId) -> Analysis<Rc<types::Eve
             } = &field.kind;
 
             let typ = type_desc(&mut scope, typ_node).and_then(|typ| match typ {
-                typ if typ.has_fixed_size(scope.db()) => Ok(typ),
+                typ if typ.has_fixed_size(scope.db()) => {
+                    if !typ.is_zero_size(scope.db()) {
+                        return Ok(typ)
+                    } else {
+                        return Err(TypeError::new(scope.error(
+                            "event field type must have a non-zero types",
+                            typ_node.span,
+                            "this type can't be used as an event field",
+                        )))
+                    }
+                }
                 _ => Err(TypeError::new(scope.error(
                     "event field type must have a fixed size",
                     typ_node.span,
--- a/crates/analyzer/src/db/queries/events.rs
+++ b/crates/analyzer/src/db/queries/events.rs
@@ -39,24 +39,34 @@ pub fn event_type(db: &dyn AnalyzerDb, event: EventId) -> Analysis<Rc<types::Eve
                 typ: typ_node,
             } = &field.kind;
 
-            let typ = type_desc(&mut scope, typ_node).and_then(|typ| match typ {
-                typ if typ.has_fixed_size(scope.db()) => {
-                    if *is_idx && typ.is_zero_sized(db) {
-                        Err(TypeError::new(scope.error(
-                            "event field type must have a non-zero types",
+            let typ =
+                type_desc(&mut scope, typ_node).and_then(|typ| match db.lookup_intern_type(typ) {
+                    types::Type::Generic(_)
+                    | types::Type::Map(_)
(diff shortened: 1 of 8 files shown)


crates/analyzer/src/db/queries/events.rs
```diff
@@ -40,7 +40,17 @@ pub fn event_type(db: &dyn AnalyzerDb, event: EventId) -> Analysis<Rc<types::Eve
             } = &field.kind;
 
             let typ = type_desc(&mut scope, typ_node).and_then(|typ| match typ {
-                typ if typ.has_fixed_size(scope.db()) => Ok(typ),
+                typ if typ.has_fixed_size(scope.db()) => {
```
Y-Nak, line 43: There is no need to reject all zero-sized fields. It'd be better to reject a field iff its type is zero-sized *and* it is indexed.

crates/analyzer/src/namespace/types.rs
```diff
@@ -494,6 +497,14 @@ impl Type {
         false
     }
 
+    pub fn is_zero_size(&self, db: &dyn AnalyzerDb) -> bool {
+        match &self {
+            Type::Base(Base::Unit) => true,
+            Type::Contract(inner) => inner.fields(db).is_empty(),
+            Type::Struct(inner) => inner.fields(db).is_empty(),
+            _ => false,
```
Y-Nak, line 505: `contract/struct` can be zero-sized even if their fields are not empty.

Example.
```
struct Zero {
    x: ()
}
```

Also, other types can be zero-sized.

* `Tuple` becomes a zero-sized type if it has no elements or all elements are zero-sized type.
* `Array` becomes a zero-sized type if its length is 0 or its element type is a zero-sized type.

vuvoth: I create salsa query `is_zero_sized` to handle it. I think about this should be `type_size` will more general.

crates/analyzer/src/db/queries/events.rs
vuvoth: Done

crates/analyzer/src/db/queries/types.rs
```diff
@@ -69,3 +69,45 @@ pub fn type_alias_type_cycle(
...
+                // assume all type is correct.
+                let field_type = db.struct_field_type(*field_type_id).value.unwrap();
+                if !(field_type).is_zero_sized(db) {
+                    return false;
+                }
+            }
+            true
+        }
+        Type::Contract(contract_id) => {
+            for field_type_id in db.contract_all_fields(contract_id).iter() {
```
vuvoth, line 98: Logic to handle struct and contract very similar. Have any better way to refactor it....

crates/analyzer/src/db/queries/types.rs
```diff
@@ -69,3 +69,45 @@ pub fn type_alias_type_cycle(
...
+                let field_type = db.contract_field_type(*field_type_id).value.unwrap();
+                if !(field_type).is_zero_sized(db) {
+                    return false;
+                }
+            }
+            true
+        }
+        Type::Generic(_) | Type::Map(_) | Type::SelfContract(_) => false,
```
vuvoth, line 106: Is Generic type always have size > 0?

crates/analyzer/src/db/queries/types.rs
sbillig: A contract type is really just an address (its size is that of the `address` type), so you don't need to look at the fields.

crates/analyzer/src/db/queries/types.rs
sbillig: It could be zero-sized or not, depending on how its used. For example:

```
event Foo<T> {
  idx val: T
}

pub fn emit_foo<T>(ctx: Context, val: T) {
  emit Foo(val)
}

emit_foo(ctx, 10) // ok
emit_foo(ctx, ())  // not ok, T is zero-sized
```

For now, it's safe to use `unreachable!()` for this case, because we don't support generic fields in structs or events yet. When we do, we'll have to determine the size of the concrete type.

(Map and SelfContract can also be `unreachable!()`, as they can't appear in a field.)

crates/analyzer/src/db/queries/types.rs
vuvoth: > A contract type is really just an address (its size is that of the `address` type), so you don't need to look at the fields.

Thanks! I fixed it

crates/analyzer/src/db/queries/types.rs
vuvoth: I think better we should throw `fe error` in this case. Use `unreachable!()` maybe non sense I think.

crates/analyzer/src/db/queries/types.rs
```diff
@@ -69,3 +69,36 @@ pub fn type_alias_type_cycle(
...
+            true
+        }
+        Type::Generic(_) | Type::Map(_) | Type::SelfContract(_) | Type::Contract(_) => false,
+    }
+}
+
+// because we already catch this error so this should be empty
+pub fn is_zero_sized_cycle(_db: &dyn AnalyzerDb, _cycle: &[String], _ty: &types::TypeId) -> bool {
+    true
+}
```
Y-Nak, line 104: This feels like a little bit dirty hack. I think it'd be better to detect a struct cycle in `type_desc`, but I'm not sure.

crates/analyzer/src/db/queries/types.rs
Y-Nak: I don't think returning `Error` here is a good option since using errors to report/hide internal bugs is not a good practice. Entering the `Type::Generic` arm means ICE in the current state, so `unreachable!()` would be suitable here.

sbillig:
Defunct

PR #775, opened 2022-07-29
Reject pure functions being called on instances

cburgdorf:
### What was wrong?

There were certain cases were the compiler did not reject code where a pure fn was called on an instance.

### How was it fixed?

Minor tweak

### DIFF
--- a/crates/test-files/fixtures/features/two_contracts.fe
+++ b/crates/test-files/fixtures/features/two_contracts.fe
@@ -12,7 +12,7 @@ contract Foo {
         return self.other.answer()
     }
 
-    pub fn add(_ x: u256, _ y: u256) -> u256 {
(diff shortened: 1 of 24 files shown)


crates/test-files/fixtures/features/two_contracts.fe
```diff
@@ -12,7 +12,7 @@ contract Foo {
         return self.other.answer()
     }
 
-    pub fn add(_ x: u256, _ y: u256) -> u256 {
```
cburgdorf, line 15: A lot of these tests in the diff would not need `self` but I don't want to change the meaning of the test. I did add a test that demonstrates that calling pure functions on an external contract works just fine via `ExternalContract::some_func()`

Y-Nak (approved):
Cool!

PR #776, opened 2022-08-02
Change comments from using `#` to using `//`

cburgdorf:
### What was wrong?

It seems like we have consensus that we don't want to continue to use `#` for comments but rather use `//` instead. This PR changes that.

### How was it fixed?

- Tiny change in the parser
- Updated code files
- Updated examples + docs
- Updated highlight.js on the website to use the custom `fe` highlighting that the docs already use

### DIFF
--- a/crates/analyzer/tests/snapshots/analysis__const_generics.snap
+++ b/crates/analyzer/tests/snapshots/analysis__const_generics.snap
@@ -7,7 +7,7 @@ note:
    ┌─ const_generics.fe:2:5
    │  
  2 │ ╭     pub fn bar() {
- 3 │ │         # const generics with literal.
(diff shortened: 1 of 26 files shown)


Y-Nak (approved):
Perfect!

cburgdorf:
This actually needed to be rebased. It's now causing a CI failure on master. I'll send a follow up.

cburgdorf:
> This actually needed to be rebased. It's now causing a CI failure on master. I'll send a follow up.

I pushed a fix directly to master :hand_over_mouth:  https://github.com/ethereum/fe/commit/9300ef4c9c9d4327a4551b036006cad2803e4abe

Y-Nak:
Yeah, I missed that. Thanks for following up.

PR #782, opened 2022-08-25
Fix platform detection in Firefox

cburgdorf:
### What was wrong?

On FireFox we currently fail to detect the platform (windows, linux, mac).

### How was it fixed?

User `navigator.userAgent` instead of `navigator.appVersion`
´


cburgdorf:
I've already deployed that on the live site.

cburgdorf:
I'm going to merge this as the lint fixes are important to get CI green on master and the website fix is deployed already anyway :D

PR #783, opened 2022-09-02
Set function access in Json ABI based on presence of ctx and self parameter

### LINKED ISSUE #678: Set function access in Json ABI based on presence of `ctx` and `self` parameter
### What is wrong?

As noted in #638, [`FunctionSignature`](https://github.com/ethereum/fe/blob/766d44c9fd1e88509b7539f5b69be766bb7043f4/crates/analyzer/src/namespace/types.rs#L156-L161)s now include a `ctx_decl` attribute. If both the `ctx_decl` and `self_decl` attributes are `None`, then we can mark it as a pure function in the [Json ABI](https://github.com/ethereum/fe/tree/766d44c9fd1e88509b7539f5b69be766bb7043f4/crates/abi/src).

example:

```
contract Foo:
  # this should be marked as pure
  pub fn add(a: u256, b: u256):
    pass
```

### How can it be fixed

:keyboard:

vuvoth:
### What was wrong?
Resolve #678  

### To-Do

- [ ] Clean up commit history


crates/abi/src/function.rs
```diff
@@ -31,6 +31,8 @@ impl AbiFunction {
         name: String,
         args: Vec<(String, AbiType)>,
         ret_ty: Option<AbiType>,
+        has_self: bool,
+        has_ctx: bool,
```
Y-Nak, line 35: It's better to take simply `state_mutability` instead of introducing Fe-specific concepts like `self` and `ctx` to avoid unnecessary tight coupling with Fe language specification.

Y-Nak (commented):
`mut` would be launched soon. So I think it'd be better to finish the PR after its deployment.

crates/abi/src/function.rs
vuvoth: > `mut` would be launched soon. So I think it'd be better to finish the PR after its deployment.

Okay. I got it!

sbillig:
> `mut` would be launched soon. So I think it'd be better to finish the PR after its deployment.

"soon" has finally come, a couple months late :sweat_smile:. I didn't update the json abi in the `mut` PR, @vuvoth.

sbillig:
Sorry @vuvoth, I just removed CtxDecl because it seemed pointless. Didn't mean to make extra work for you :grimacing:

vuvoth:
> Sorry @vuvoth, I just removed CtxDecl because it seemed pointless. Didn't mean to make extra work for you grimacing

No problem!

crates/abi/src/function.rs
```diff
@@ -14,6 +14,36 @@ pub enum StateMutability {
...
 
+impl StateMutability {
+    pub fn decision(mut_self: Option<bool>, mut_ctx: Option<bool>) -> Self {
+        // Check ABI conformity
+        // https://github.com/ethereum/fe/issues/558
+        //
+        //             |none self             |    self  |  mut self             |
+        // | none ctx  |    pure              |    view  |  payable/nonpayable   |
+        // | ctx       |    view              |    view  |  payable/nonpayable   |
+        // | mut ctx   |   payable/nopayable  |    view  |  payable/nonpayable   |
```
g-r-a-n-t, line 25: @cburgdorf `mut ctx` + `self` should be `payable/non payable`, right?

cburgdorf: Yes, as soon as we have `mut` in a public contract function, then it needs to be `payable`

vuvoth: I think more sense if we only allow `mut ctx` follow after `mut self` because we want to `modify` something.

cburgdorf: These are two different things. You can take `self` non mutable but take `mut ctx` because e.g. you want to emit an event. So you aren't changing any storage of the contract but you are still making changes to the blockchain. In terms of the ABI both of these should be marked as `payable`. There's a longer form write-up of the different access classes under that scheme: https://github.com/ethereum/fe/issues/558

sbillig:
@vuvoth I pushed a commit to your branch, changing the `Option<bool>` args to enums, and using a match statement instead of an if/else tree. I also removed some changes that weren't necessary (unused functions, etc), and added a newsfragment.

g-r-a-n-t (approved):

Y-Nak (approved):

Y-Nak:
CI book fails because the request to `alchemy` returns 403. I'll fix it in another PR.

PR #785, opened 2022-09-04
Tuple pattern

Y-Nak:
### What was wrong?
Add tuple pattern matching.
This PR depends on #770 

### To-Do

- [ ] Clean up commit history


Y-Nak:
I decided to include this in #770

PR #790, opened 2022-09-09
Properly parse struct definition by rejecting invalid public identifiers, Fixes: #787

### LINKED ISSUE #787: Reject invalid `pub` in a `struct` definition
### What is wrong?
Currently, an invalid `pub` appearance is not rejected, e.g.,
```rust
pub struct Bar {
    pub x: i32
pub}
```

### How can it be fixed

Tweaking the below code will resolve the issue.
https://github.com/ethereum/fe/blob/1190d93ef644d4e993497bf27ce5a48068c872ba/crates/parser/src/grammar/types.rs#L28-L55

Narasimha1997:
Fix #787 
Before, the struct definitions like:
```
pub struct Bar {
    pub x: i32
pub}
```
were not rejected by the parser. (as mentioned in #787 ). This pull request fixes this issue.
Also tests are written to validate this.

@Y-Nak

### DIFF
--- a/crates/parser/src/grammar/types.rs
+++ b/crates/parser/src/grammar/types.rs
@@ -25,6 +25,7 @@ pub fn parse_struct_def(
     let mut fields = vec![];
     let mut functions = vec![];
     par.enter_block(span, "struct body must start with `{`")?;
+
     loop {
         par.eat_newlines();
         let pub_qual = par.optional(TokenKind::Pub).map(|tok| tok.span);
@@ -42,7 +43,7 @@ pub fn parse_struct_def(
             TokenKind::Fn | TokenKind::Unsafe => {
                 functions.push(parse_fn_def(par, pub_qual)?);
             }
-            TokenKind::BraceClose => {
+            TokenKind::BraceClose if pub_qual.is_none() => {
                 span += par.next()?.span;
                 break;
             }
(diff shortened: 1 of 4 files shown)


crates/parser/src/grammar/types.rs
```diff
@@ -43,6 +44,12 @@ pub fn parse_struct_def(
                 functions.push(parse_fn_def(par, pub_qual)?);
             }
             TokenKind::BraceClose => {
+                if pub_qual.is_some() {
+                    let tok: Token = par.next()?;
+                    par.unexpected_token_error(&tok, "failed to parse struct definition", vec![]);
+                    return Err(ParseFailed);
+                }
```
Y-Nak, line 51:
```suggestion
            TokenKind::BraceClose if pub_qual.is_none() => {

```

It seems the if block is the same as the block of the `_` arm, so I think it would be better to use the match guard or return a more detailed error message if the `TokenKind::BraceClose` has its own check for `pub_qual`.

Y-Nak (commented):
Thanks for submitting PR! 
Basically, you're on the right track. I left a nit and will merge this PR after it is addressed.

Y-Nak:
Also, could you add `fix #787` to the PR comments so that the issue is automatically closed when the PR is merged?

Narasimha1997:
@Y-Nak , I have updated the changes you suggested

crates/parser/src/grammar/types.rs
Narasimha1997: Done! decided to go with the former one.

Y-Nak:
Thanks!

PR #797, opened 2022-09-27
Struct pattern

### LINKED ISSUE #794: Panics when a struct contains a enum type
### What is wrong?
The code below causes panic 
```
enum Accumulator {
    Zero
    Int(u32)
    Clo(i32, i32)
}

struct MyStruct {
    pub acc: Accumulator
    pub y: i32
}

contract Bar {
}
```
Ref: https://github.com/ethereum/fe/pull/772#discussion_r970844998

@cburgdorf
Suggestion: 
1. I think we still need `#event` attributes in addition to `#index` attributes so a struct can contain enum types.
2. It'd be better to forbid implementing `Emittable` to private structs.

Y-Nak:
## Add struct pattern
Add a struct pattern that can be used in `match` statements.
e.g.,
```rust
struct MyS {
    pub x: i32
    pub b: bool

    fn eval(self) -> i32 {
        match self {
            MyS {x: x, b: true} => {
                return x
            }
            MyS {x: x, b: false} => {
                return -x
            }
        }
    }
}
```

#### NOTE 
1. Currently, field labels are required even if a binding name is the same as the field name.
2. To bind a field, the fields must be `pub`. Otherwise, the compiler complains about it.

## FIx #794 
```rust
enum Accumulator {
    Zero
    Int(u32)
    Clo(i32, i32)
}

struct MyStruct {
    pub acc: Accumulator
    pub y: i32
}

contract Bar {
}
```

I noticed the above example still causes an ICE even in the current master.
In #798, I overlooked that we should modify `codegen/abi.rs` so that non-emittable structs are excluded from abi events.
I fixed the problem in this PR for testing a nested struct pattern. 

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/src/namespace/items.rs
+++ b/crates/analyzer/src/namespace/items.rs
@@ -1880,7 +1880,7 @@ impl TraitId {
     pub fn is_implemented_for(&self, db: &dyn AnalyzerDb, ty: TypeId) -> bool {
         // All encodable structs automagically implement the Emittable trait
         // TODO: Remove this when we have the `Encode / Decode` trait.
-        if self.is_std_trait(db, EMITTABLE_TRAIT_NAME) && ty.is_emitable(db) {
+        if self.is_std_trait(db, EMITTABLE_TRAIT_NAME) && ty.is_emittable(db) {
             return true;
         }
 
--- a/crates/analyzer/src/traversal/functions.rs
+++ b/crates/analyzer/src/traversal/functions.rs
@@ -483,7 +483,7 @@ fn struct_pattern(
                 let err = scope.fancy_error(
                     &format!("field `{}` is not public field", f_name),
                     vec![
-                        Label::primary(span, "missing field"),
+                        Label::primary(span, format!("`{}` is not public", f_name)),
                         Label::secondary(field.span(scope.db()), "field is defined here"),
                     ],
                     vec![],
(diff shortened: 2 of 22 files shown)


crates/analyzer/src/namespace/items.rs
```diff
@@ -1880,7 +1880,7 @@ impl TraitId {
     pub fn is_implemented_for(&self, db: &dyn AnalyzerDb, ty: TypeId) -> bool {
         // All encodable structs automagically implement the Emittable trait
         // TODO: Remove this when we have the `Encode / Decode` trait.
-        if self.is_std_trait(db, EMITTABLE_TRAIT_NAME) && ty.is_emitable(db) {
+        if self.is_std_trait(db, EMITTABLE_TRAIT_NAME) && ty.is_emittable(db) {
```
cburgdorf, line 1883: Oops :sweat_smile:

crates/analyzer/src/traversal/functions.rs
```diff
@@ -446,26 +559,84 @@ fn match_tuple_pattern(
...
+    if rest_pat_pos.is_some() && tuple_elts.len() - 1 > expected_tys.len() {
+        return emit_len_error(tuple_elts.len() - 1, expected_tys.len());
+    } else if rest_pat_pos.is_none() && tuple_elts.len() != expected_tys.len() {
+        return emit_len_error(tuple_elts.len(), expected_tys.len());
     };
 
+    collect_binds_from_pat_vec(scope, tuple_elts, expected_tys)
+}
+
+// fn match_tuple_pattern(
```
cburgdorf, line 571: Was this supposed to stay here?

cburgdorf (approved):
Great stuff!

crates/analyzer/src/traversal/functions.rs
Y-Nak: Good catch! Fixed.

PR #801, opened 2022-10-25
Reject to_mem on primitive types in storage

cburgdorf:
### What was wrong?

The following code is currently not rejected by the compiler even though it should.

```
contract Test {
   foo: u256

    pub fn bar(self)-> u256{
        return self.foo.to_mem()
    }
}
```

### How was it fixed?

If a value is in `storage`, check if it is a primitive and if so present the user error


crates/analyzer/src/traversal/expressions.rs
```diff
@@ -1767,7 +1767,21 @@ fn expr_call_builtin_value_method(
                         vec![],
                     );
                 }
-                Location::Storage { .. } => {}
+                Location::Storage { .. } => {
+                    if value_attrs.typ.is_base(context.db()) {
```
Y-Nak, line 1771: I think we also have to handle the case where `typ == Type::Contract(_)` because it's a `newtype`'d address and is treated as a base type internally.

cburgdorf: Good catch! Fixed it.

PR #802, opened 2022-10-25
docs: example codes url correction

ray-themedium:
### What was wrong?
URL of Uniswap example codes reference other branch not master

### How was it fixed?
Fix URL


cburgdorf:
Thanks!

PR #805, opened 2022-11-04
Implement trait associated functions

cburgdorf:
### What was wrong?

We don't have trait associated functions.

### How was it fixed?

Basically just enhanced `expr_call_path` to keep looking for candidates that are implemented via traits if we can not resolve the path directly on the type.

This also includes #815 which can be reviewed seperately.

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -5,7 +5,7 @@ use crate::errors::{self, FatalError, IndexingError, TypeCoercionError};
 use crate::namespace::items::{
     EnumVariantId, EnumVariantKind, FunctionId, FunctionSigId, ImplId, Item, StructId, TypeDef,
 };
-use crate::namespace::scopes::BlockScopeType;
+use crate::namespace::scopes::{check_visibility, BlockScopeType};
 use crate::namespace::types::{Array, Base, FeString, Integer, Tuple, Type, TypeDowncast, TypeId};
 use crate::operations;
 use crate::traversal::call_args::{validate_arg_count, validate_named_args};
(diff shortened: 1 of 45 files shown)


crates/analyzer/src/traversal/expressions.rs
```diff
@@ -1,7 +1,7 @@
 use crate::display::Displayable;
 use crate::errors::{FatalError, IndexingError};
 use crate::namespace::items::{FunctionId, FunctionSigId, ImplId, Item, StructId, TypeDef};
-use crate::namespace::scopes::BlockScopeType;
+use crate::namespace::scopes::{check_visibility, BlockScopeType};
```
cburgdorf, line 4: Should this be moved on the `AnaylzerContext` instead of importing it here?

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -995,8 +995,139 @@ fn expr_call_path<T: std::fmt::Display>(
     generic_args: &Option<Node<Vec<fe::GenericArg>>>,
     args: &Node<Vec<Node<fe::CallArg>>>,
 ) -> Result<(ExpressionAttributes, CallType), FatalError> {
-    let named_thing = context.resolve_path(path, func.span)?;
-    expr_call_named_thing(context, named_thing, func, generic_args, args)
+    match context.resolve_any_path(path) {
```
Y-Nak, line 998: Why is `resolve_any_path` called instead of `resolve_visible_path`? If a "normal" associated function is not visible, it should not be counted as a callable candidate.

For example,  there is no ambiguity in the following example, but the compiler rejects it.
```rust
// a.fe
pub struct MyS {
    fn x() -> u256 {
        return 10
    }
}

// main.fe
use a::MyS

trait Trait {
    fn x() -> u256;
}

impl Trait for MyS {
    fn x() -> u256 {
        return 10
    }
}

fn main() -> u256 {
    return MyS::x()
}
```

cburgdorf: Good catch! Eagle eyes! :eagle: I pushed a solution in another commit. I'm not super happy with it but I'm not very creative this morning.

Y-Nak (commented):
Seems great! I just left one nit.

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -988,7 +988,11 @@ fn expr_call_path<T: std::fmt::Display>(
         // If we we can't resolve a call to a path e.g. `foo::Bar::do_thing()` there is a chance that `do_thing`
         // still exists as as a trait associated function for `foo::Bar`.
         None => expr_call_trait_associated_function(context, path, func, generic_args, args),
-    }
+    };
+
+    if x.is_err() {}
```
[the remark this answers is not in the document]
cburgdorf: Ooops...indeed.

PR #808, opened 2022-11-11
Move feature test_method_return tests from rstest to a macro

sbillig:
### What was wrong?

Most of our "feature" tests are (rstest cases) of the `test_method_return` variety. These tests don't have descriptive names ("case_136", etc), which makes it hard to debug a broken test. They're also numbered automatically, which means that if a new test is added to the middle of the list (next to similar cases, say), all of the following tests will be renumbered, and the snapshot file names will all change.

rstest allows you to specify a case name (eg `case::while_loop(...)`) which solves the first issue, but there doesn't seem to be a way to disable the case numbering.

(Note: this is a change I made as part of #777; I'd like this to land on master first so we can easily compare the gas usage of these tests before and after the mut changes) 

### How was it fixed?

Replaced this use of rstest with a macro.


cburgdorf (approved):
LGTM

PR #813, opened 2022-11-15
Extern and intrinsics

Y-Nak:
### To-Do

- [ ] Clean up commit history


cburgdorf:
I understood [your comment](https://discord.com/channels/834132302031487017/842843251756630038/1002306004869980200) here that you were ok with attributes that don't need brackets. Did I misinterpret or have you changed your mind about that?

Y-Nak:
I just forgot [sean's comment]( https://discord.com/channels/834132302031487017/842843251756630038/1002257110379274290) and thought brackets are necessary to allow `NAME = EXPR` in attributes. But as Sean pointed out in the comment, it should be much better for attributes to have the same syntax as a function call like `#target(evm)` or `#cfg(target: evm, abi: nu)`. I'll remove the commit and make it back to the original. Thanks for pointing that out!

PR #815, opened 2022-11-16
Disallow importing private type via `use`

cburgdorf:
### What was wrong?

We currently do not error when trying to import types via `use` whith are not `pub`. E.g. `use foo::PrivateStruct` does not error.

### How was it fixed?

Trivial

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/analyzer/tests/snapshots/errors__bad_visibility.snap
+++ b/crates/analyzer/tests/snapshots/errors__bad_visibility.snap
@@ -3,11 +3,47 @@ source: crates/analyzer/tests/errors.rs
 expression: error_string_ingot(&path)
 
 ---
-error: unresolved path item
+error: type MyInt is private
+  ┌─ compile_errors/bad_visibility/src/main.fe:1:11
+  │
+1 │ use foo::{MyInt, MY_CONST, MyStruct, MyTrait, my_func, MyContract, MyEnum }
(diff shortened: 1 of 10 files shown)


crates/analyzer/tests/snapshots/errors__bad_visibility.snap
```diff
@@ -3,11 +3,47 @@ source: crates/analyzer/tests/errors.rs
...
+  ┌─ compile_errors/bad_visibility/src/main.fe:1:56
+  │
+1 │ use foo::{MyInt, MY_CONST, MyStruct, MyTrait, my_func, MyContract, MyEnum }
+  │                                                        ^^^^^^^^^^ MyContract
+
+error: type MyEnum is private
   ┌─ compile_errors/bad_visibility/src/main.fe:1:68
   │
 1 │ use foo::{MyInt, MY_CONST, MyStruct, MyTrait, my_func, MyContract, MyEnum }
-  │                                                                    ^^^^^^ not found
```
cburgdorf, line 10: Not strictly related to the fix but the test was wrong in that it imported a non existing Enum rather than an enum that isn't `pub`.

Y-Nak (approved):
LGTM

sbillig:
Closing; already merged as part of #805

PR #818, opened 2022-11-23
Canonicalize function dependency resolusion

Y-Nak:
### What was wrong?
The output yul wasn't canonicalized in terms of the output function order due to the usage of `FxHashMap` and `BTreeSet`. This caused non-deterministic byte code output.

### How was it fixed?
Rewrite them to `IndexMap` and `IndexSet` respectively.

### To-Do

- [x] Clean up commit history


sbillig (approved):
:+1:

PR #822, opened 2022-11-28
Fix nested array encoding

Y-Nak:
### What was wrong?
fix #821

NOTE:
I modified the makefile to remove the `redundant_closure_for_method_calls` lint.

### How was it fixed?
Change to find a tuple type in an array type recursively.

### To-Do

- [x] Clean up commit history


cburgdorf:
Great!

PR #823, opened 2022-11-28
Reflect `mut` feature to differential tests

Y-Nak:
### What was wrong?
`differential tests` failed because of missing `mut` feature modification.

### How was it fixed?
Add the `mut` keyword to some variables where it's required.

### To-Do

- [x] Clean up commit history


sbillig (approved):
:+1:

PR #824, opened 2022-11-28
Copy complex return values; fixes "bountiful" struct aliasing bug

sbillig:
### What was wrong?

Values returned from functions weren't being copied. If a struct function returns a (non-primitive) field, it would be a reference to that field, which can lead to unexpected behavior.

See `bountiful_struct_copy_bug.fe`

### How was it fixed?

Changed a boolean. Caused a gas penalty, because we don't have a good way to elide unnecessary copies.

### To-Do

- [ ] Clean up commit history


sbillig:
Note that we still don't have even a rudimentary borrow checker, so the original `b.swap_y(p1: b.first, p2: b.second)` line is still broken.

cburgdorf (approved):
Sweet!

PR #825, opened 2022-12-01
Add dumb borrow check of fn call args

### LINKED ISSUE #810: Add simple "borrow checker" to restrict fn call arg borrows
### What is wrong?

(Follow-up issue for #777)

It shouldn't be possible to pass more than one reference to an object into a function call when at least one of those references is `mut`.

For example, in the `fn foo` below, `q` is non-mut, and thus shouldn't change during the duration of the function call. However, we don't currently prevent a single `Point` being passed in as both `mut p` and non-mut `q`.

```rust
let mut p: Point = Point(x: 10, y: 20)
modify_points(p, p)

fn foo(mut _ p: Point, q: Point) {
  assert q.x == 10
  p.x = 100
  assert q.x == 100 // XXX wtf, shouldn't happen
}
```

### How can it be fixed

:brain:

sbillig:
### What was wrong?

fixes #810

### How was it fixed?

This is a temporary solution to the only borrowing bug that's possible in the current semantics of Fe, namely passing a mutable reference to a non-primitive object into a fn call, and another reference to that same object (mutable or otherwise).

This is an ugly brute force solution that will definitely not scale beyond this simple case, and doesn't do anything smart like allow disjoint partial borrows.

This should be replaced with a proper borrow checker (presumably operating on MIR) when Fe gains some kind of reference/projection type.

### DIFF
--- a/crates/analyzer/src/context.rs
+++ b/crates/analyzer/src/context.rs
@@ -240,7 +240,7 @@ pub enum NamedThing {
     },
     // SelfType // when/if we add a `Self` type keyword
     Variable {
-        name: String,
+        name: SmolStr,
         typ: Result<TypeId, TypeError>,
         is_const: bool,
         span: Span,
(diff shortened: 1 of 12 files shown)


cburgdorf (approved):
LGTM

Y-Nak (approved):
LGTM!

PR #827, opened 2022-12-05
Christoph/fix/ctx abi

cburgdorf:
### What was wrong?

1. The `__init__` function of a contract should always be categorized as `payable` even if it looks pure from its signature. 

2. the `CtxDecl` wasn't correctly inferred in case the context was taken with `mut`

### How was it fixed?

Trivial changes


sbillig (approved):
👍

PR #828, opened 2022-12-06
Update website example to match latest Fe

cburgdorf:
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -196,26 +196,25 @@
 
                       </div>
                       <code class="language-fe flex-auto relative block text-white pt-4 pb-4 px-4 overflow-auto">
+// Structs can be emitted as events
+struct Signed {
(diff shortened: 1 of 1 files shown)


docs/src/spec/items/events.md
```diff
@@ -0,0 +1 @@
+# Events
```
Y-Nak, line 1: Is this a leftover?

docs/src/spec/type_system/types/event.md
```diff
@@ -0,0 +1 @@
+# Event Types
```
Y-Nak, line 1: ditto

cburgdorf: Ah, those are auto generated

Y-Nak:
👍

PR #832, opened 2022-12-31
Cst

Y-Nak:
Implement a new parser that builds CST(Concrete Syntax Tree); please see [rust-analyzer's doc](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/syntax.md) for more details about CST.

This PR is intended to replace the old parser and add higher-level syntax to the language.
In this implementation, I also added an error recovery mechanism to the parser; I'll describe it later.

### `impl` block
Add an `impl` block to implement methods and forbid defining methods inside `struct` and `enum` bodies.
#### Example
```rust
impl MyStruct {
    pub fn foo() {...}
}
```

### Logical Operators
Remove python style logical operators `or` and `and.` Now they are replaced with `||` and `&&`.
#### Example
```rust  
b1 && b2 || b3
```

### Change type annotation to optional
In the `let` statement, type annotations are optional because we'll add type inference in this refactoring process.
#### Example
```rust
let x = 1
```

### Add pointer type
Allow using pointer types.
#### Examples
```rust
let x: *u8
unsafe fn mem_copy(dst: *u8, src: *u8, len: usize) {...}
```

### Allow patterns in binding
Now, all binding LHSs are changed to `pattern`. This affects `let` LHS and `for` LHS.
This will reduce the ad-hoc analysis in binding check because we can use the pattern match anomaly check, which we already have.
#### Example
 ```
let MyEnum::A(x) | MyEnum::B(x) = my_enum
for MyEnum::A(x) in enum_arr {...}
```
 
### Change `if` and `match` to expressions.
Now `if` and `match` become expressions. Accordingly, the python-style ternary operator is removed.
#### Example
```rust
let x = if b { 1 } else { 2 }
let y = match x { 
    MyEnum::A(a) => a
    MyEnum::B(b) => b
}
```

### Struct Constructor
Add syntax to initialize a record type.
#### Example
```rust
Struct { x: 1, y }
```

### Generic Parameters
Allow ADT to have generics in its fields.
#### Example
```rust
struct Foo<T> {
     t: T
}

enum Option<T> {
     Some(T)
     None
}
```

### Multiple Type Bounds
Allow describing multiple type bounds using the `+` operator.
#### Example
```rust
struct S<T: Trait1 + Trait2> {
    t: T
}
``` 

### `where` Clause
Allow describing type bounds in the `where` clause.
#### Example
```rust
impl Foo<T> for Bar<T> 
    where 
           T: Clone
           Option<T>: Clone
{ ... }
```

### `use` Changes
1. use tree can have nested groups as its children
2. `as` keywords are introduced for renaming
3. `self` keywords can be used to import its parent module.
4. wildcard `*` can be used to import all items from the module
#### Example
```rust
use foo::{self as self_, Bar::{Bar as _, Baz}, *}
```

### Add `extern` block
Allow extern block to declare functions
#### Example
```rust
#[intrinsic]
extern {
    pub unsafe mem_copy(dst: *u8, src: *u8, len: usize) -> bool
    ...
} 
```

## Error Recovery
The new parser has the ability to recover from invalid syntax and continue parsing to collect valid pieces from the source text.
This feature is important, especially when it's used by the LSP so it can emit a semantic error even if some parts are syntactically invalid.

For example, the parser will build the CST from the below invalid call expression.
```rust
foo(x, y a, z ;)
```
↓
```
  CallExpr@0..16
    Path@0..3
      PathSegment@0..3
        Ident@0..3 "foo"
    CallArgList@3..16
      LParen@3..4 "("
      CallArg@4..5
        Path@4..5
          PathSegment@4..5
            Ident@4..5 "x"
      Comma@5..6 ","
      WhiteSpace@6..7 " "
      CallArg@7..10
        Path@7..8
          PathSegment@7..8
            Ident@7..8 "y"
        WhiteSpace@8..9 " "
        Error@9..10
          Ident@9..10 "a"
      Comma@10..11 ","
      WhiteSpace@11..12 " "
      CallArg@12..15
        Path@12..13
          PathSegment@12..13
            Ident@12..13 "z"
        WhiteSpace@13..14 " "
        Error@14..15
          SemiColon@14..15 ";"
      RParen@15..16 ")"
```

The call expression is obviously invalid, but the analyzer can use some information from this invalid CST.
e.g., the analyzer can know that the caller function name is `foo`, and it will use valid argument information, `x`, `y`, and `z` in type checking.

The below code is a more complex example of error recovery.
```rust
impl X<T, u for Y<T where T: X {}
```
In the current implementation, the analyzer can use the below information
1. this is an impl trait item
2. `X<T, u>` is implemented to `Y<T>`
3. `where T: X`
4. The impl block is empty (This is valid syntax)

More examples can be found in `parser2/test_files/error_recovery`.

### Architecture
The main players in the new parser are the `Parser` struct and scopes defined by the `define_scope` macro

#### `Parser`
Against the expectation from its name, `Parser` doesn't parse the source file.
Instead of doing parsing, it manages the parsing environment and provides a bunch of methods that can be used for each scope to parse the source.
Its main function is `Parser::parse(scope: impl Parse, ..)`, which sets up the environment for `scope` and invokes `scope`'s `parse` method.
Please see the `parser2/src/parser/mod.rs` for more details. 

#### Scope
Scope plays the main role in parsing. It should be defined by `define_scope` macro like below.
```rust
define_scope! {
    pub(crate) TupleTypeScope{ allow_bounds: bool },
    TupleType,
    Override(
        RParen,
        Comma
    )
}
```
The first argument is the scope's type definition; it can be possible to add arbitrary fields to the type as long as the field implements `Default`.

The second argument is the syntax node kind that marks the current syntax scope(yes, the scope is almost corresponding to the branch in CST).

The third argument is used to define recovery tokens. There are two options, 1. `Inheritance`, 2. `Override`.
 If you choose `Inheritance`, the parent scope's recovery tokens are used in addition to its own recovery tokens to recover from the error(it is automatically used in `Parser::recover(..)` method.).  On the other hand, if you choose `Override`, the parent's recovery tokens are ignored. 
 
After defining the scope, you have to implement the `Parse` trait manually to implement the way of parsing itself.

#### Emit contextual errors.
In this PR, I don't make it possible to emit/build a contextual error. Contextual error is an error that describes information about the context that encloses the error.
For example, when the `path` scope parses `path`, the scope itself doesn't know its enclosing context, e.g., is it called in the `type` parsing phase or the `expression` parsing phase? 
To address this, we have to pass the context information when initializing a scope. This could be implemented relatively easily by modifying the `define_scope` macro and scope initialization (But it'd be quite boring work though..).

#### Test Helper
I wrote a small procedural macro to enumerate test files from a specified directory and dump `snap` files to a specified directory.
please see `parser2/tests` for how to use it.

#### Modify `assert_herness_gas_report` macro
See [this](https://insta.rs/docs/patterns/#rstest). The changes in `crates/tests/snapshot` are because of the change.


sbillig:
Why are the fields of the scope structs `Rc<Cell<_>>`? It's not needed in the code that exists here, so I guess it must be for future use, but I can't think of why the fields would need to be mutated.

sbillig:
Minor nitpick: I think `with_next_expected_tokens` and `with_recovery_tokens` are the rare case where it would read nicer if the closure arg came first. Eg

```rust
    parser.with_next_expected_tokens(|parser| {
        if !parser.bump_if(SyntaxKind::Ident) {
            parser.error_and_recover("expected ident for the function name", None)
        }
    }, &[SyntaxKind::Lt, SyntaxKind::LParen]);

    parser.with_next_expected_tokens(|parser| {
        if parser.current_kind() == Some(SyntaxKind::Lt) {
            parser.parse(GenericParamListScope::default(), None);
        }
    }, &[SyntaxKind::LParen]);
```
as this matches the order of the tokens in the fe source text being parsed.

crates/parser2/src/parser/expr_atom.rs
```diff
@@ -0,0 +1,301 @@
...
+    }
+}
+
+define_scope! { MatchExprScope, MatchExpr, Inheritance }
+impl super::Parse for MatchExprScope {
+    fn parse<S: TokenStream>(&mut self, parser: &mut Parser<S>) {
+        parser.bump_expected(SyntaxKind::MatchKw);
+
+        parser.with_next_expected_tokens(&[SyntaxKind::LBrace], |parser| {
+            parser.with_recovery_tokens(&[SyntaxKind::LBrace], parse_expr_no_struct)
```
sbillig, line 125: Is the `with_recovery_tokens` here redundant?

crates/parser2/src/parser/expr_atom.rs
```diff
@@ -0,0 +1,301 @@
...
+        parser.parse(RecordFieldListScope::default(), None);
+    }
+}
+
+define_scope! { RecordFieldListScope, RecordFieldList, Override(RBrace, Comma) }
+impl super::Parse for RecordFieldListScope {
+    fn parse<S: TokenStream>(&mut self, parser: &mut Parser<S>) {
+        parser.bump_expected(SyntaxKind::LBrace);
+
+        if parser.bump_if(SyntaxKind::LBrace) {
```
sbillig, line 204: RBrace?

Y-Nak:
> Why are the fields of the scope structs Rc<Cell<_>>? It's not needed in the code that exists here, so I guess it must be for future use, but I can't think of why the fields would need to be mutated.

This is used to change the syntax kind inside the scope. e.g.,
```rust
define_scope! { ParenScope, ParenExpr, Override(RParen, Comma) }
impl super::Parse for ParenScope {
    fn parse<S: TokenStream>(&mut self, parser: &mut Parser<S>) {
        parser.bump_expected(SyntaxKind::LParen);
        parse_expr(parser);
      
        while parser.bump_if(SyntaxKind::Comma) {
            // Here, we change the syntax kind to `TupleExpr` from `ParenExpr`
            self.set_kind(SyntaxKind::TupleExpr);
            ...
        }
        ...
    }
```

But I'd like to know if changing the syntax inside a scope is confusing.

Instead of changing the syntax kind, we could implement it like
```rust
fn parse_paren_expr(parser: &mut Parser) {
       let is_tuple_expr =  parser.dry_run(|parser| {
           parser.bump_expexted(SyntaxKind::Lparen);
           parse_expr(parser);
           parser.current_kind() == Some(SytnaxKind::Comma)
       });
       if is_tuple_expr {
           parser.parse(SyntaxKind::TupleExpr::default(), None);
       } else {
           parser.parse(SytnaxKind::ParenExpr::default(), None);
      }
}
```

Which do you think is better?

Y-Nak:
> Minor nitpick: I think with_next_expected_tokens and with_recovery_tokens are the rare case where it would read nicer if the closure arg came first. 

Makes sense! I'll change it.

sbillig:
> > Why are the fields of the scope structs Rc<Cell<_>>? It's not needed in the code that exists here, so I guess it must be for future use, but I can't think of why the fields would need to be mutated.
> 
> 
> 
> This is used to change the syntax kind inside the scope. e.g.,
> 
> ```rust
> 
> define_scope! { ParenScope, ParenExpr, Override(RParen, Comma) }
> 
> impl super::Parse for ParenScope {
> 
>     fn parse<S: TokenStream>(&mut self, parser: &mut Parser<S>) {
> 
>         parser.bump_expected(SyntaxKind::LParen);
> 
>         parse_expr(parser);
> 
>       
> 
>         while parser.bump_if(SyntaxKind::Comma) {
> 
>             // Here, we change the syntax kind to `TupleExpr` from `ParenExpr`
> 
>             self.set_kind(SyntaxKind::TupleExpr);
> 
>             ...
> 
>         }
> 
>         ...
> 
>     }
> 
> ```
> 
> 
> 
> But I'd like to know if changing the syntax inside a scope is confusing.

I think this is fine. The Cell isn't needed for this case though, as we have `&mut self`.

Y-Nak:
Ok, I see. You mean that field should be `Rc<SyntaxKind>`, right?

sbillig:
Or just `SyntaxKind`; why Rc a u16?

Y-Nak:
The reason for wrapping it by `Rc` is that we have to `clone` the scope in the `Parser::parse` method not to violate the borrowing rule. 
When the `Parser::parse` method is called, `Parser` clones a passed scope and stores it into its stack, then the `Parser` invokes the scope's `parse` method.  When the scope's `parse` method returns, the parser pops the cloned scope and looks up the syntax kind of the scope. So we need to wrap the type by `Rc` to reflect the change of scope's syntax kind on the cloned scope.

sbillig:
Ah, I see. Ok, then `Rc<Cell` is required, as mutating the Rc'd value won't work when there's a clone of it.

crates/parser2/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    #[cfg(test)]
+    println!("cargo:rerun-if-changed=./std");
+}
```
sbillig, line 4: Remove this file? Changing a file in test_files seems to rebuild automatically, so I guess it's not needed.

sbillig (approved):
This is a very elegant implementation, probably the cleanest (complex) parser I've seen.

There are some common patterns that could maybe be wrapped into utility functions, but I don't have strong feelings about this.

Eq
```rust
        if !parser.bump_if(SyntaxKind::Ident) {
            parser.error_and_recover("expected `key: value`", None);
        }

        if !parser.bump_if(SyntaxKind::Colon) {
            parser.error_and_recover("expected `key: value`", None);
        }
```
could be
```rust
parser.expect_or_recover(SyntaxKind::Ident, "expected `key: value`", None);
parser.expect_or_recover(SyntaxKind::Colon, "expected `key: value`", None);
```
Maybe this is premature, as refining the errors to include more context information might change how some of this is written.

Anyway, fantastic work!

crates/parser2/build.rs
Y-Nak: Oops, removed

Y-Nak:
I addressed all reviews and add `bump_or_recover` utility method. The implementation looks better now. Thanks for the review!

Y-Nak:
Maybe, there would be more common patterns. But I'll address them later after solidifying the error context implementation.

Y-Nak:
Update: I encountered a situation where changing test files doesn't rebuild. So I added 
```rust
fn main() {
    println!("cargo:rerun-if-changed=./test_files");
}
```

PR #841, opened 2023-02-03
Add wasm test for new parser

Y-Nak:
* Add wasm test
* Use `dir-test` crate and remove the internal proc macro.


sbillig (approved):
:+1: That's a mighty fine macro you got there.

PR #842, opened 2023-02-05
Bumpup dependencies

Y-Nak:
Bump up dependencies.
The important changes are,
* `ethabi` -> `18.0`
* `evm` -> `0.37`
* `evm-runtime` -> `0.37`
* `primitive-types` -> `0.12`
* `insta` -> `1.26`

The bug that @g-r-a-n-t encountered has been fixed in this PR; Please see https://github.com/paritytech/parity-common/pull/611 what has happened.

I updated `insta` as well so that `fe-v2` doesn't fall into conflict again and again.
According to this change, `test_utils::assert_snapshot` has been modified; Please see https://insta.rs/docs/patterns/#rstest.
I don't like this change, but it'd be fine, this macro will be permanently removed within two months at the same time when `test-utils`,`test-files`, and `tests` crates are eliminated from this world.


sbillig (approved):
:+1:

PR #843, opened 2023-02-07
AST

Y-Nak:
Implemented a thin AST layer on CST.
This is a "minimal" implementation, so we need to add more useful methods while implementing HIR analysis.
Also, I fixed minor bugs in the parser and refactored them.


sbillig (approved):
Looks great! If we care about tests running on wasm (:shrug:), these tests should too. Unfortunately, `wasm_bindgen_test` doesn't like test names like `r#struct` ("`__wbgt_r#struct_1` is not a valid identifier")

Y-Nak:
Ok, I added the `wasm_bindgen_tests`.

PR #849, opened 2023-02-19
Fix broken Discord link in README

cairoeth:
### What was wrong?
Discord link was not working. Same issue on website.

### How was it fixed?
Got a working link from [here ](https://ethereum.org/en/developers/docs/smart-contracts/languages/#important-links-3)

### To-Do
//


cburgdorf (commented):
I don't really get why that link is invalid already because I just generated it very recently. But seems to be the case. I altered the commit to include a few more places that needed to be changed.

PR #852, opened 2023-02-28
Change file name pattern for snap files

cburgdorf:
### What was wrong?

The snap file name pattern was recently tweaked but it resulted in file names with double quotes such as `fe_compiler_tests__features__aug_assign@"bit_xor".snap` those file names are illegal on windows and hence also incompatible with `cargo package`

### How was it fixed?

Tweaked it further so that `fe_compiler_tests__features__aug_assign@"bit_xor".snap` becomes `fe_compiler_tests__features__aug_assign@bit_xor.snap`


crates/tests/src/snapshots/fe_compiler_tests__features__aug_assign@add.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: crates/tests/src/features.rs
+assertion_line: 1779
```
cburgdorf, line 3: Not exactly sure why `assertion_line` wasn't part of the file before but it came up as I regenerated the files.

Y-Nak (approved):
LGTM!
BTW, I think it's about time to consider another way to record gas usage. Using insta for this purpose is not so good. @g-r-a-n-t

PR #853, opened 2023-03-06
Use foundry and sepolia network in deployment guide

cburgdorf:
### What was wrong?

The deployment guide was outdated in two ways:

1. The Görli network is deprecated and scheduled to sunset soon
2. Dapp.tools was a solid choice at the initial time of creating the tutorial but it was later rewritten in Rust and turned it its successor foundry

### How was it fixed?

- Rewrote tutorial to use the Sepolia testnet which is now the recommended testnet for application development
- Rewrote tutorial to use foundry instead of dapp.tools
- fixed wording and typos


cburgdorf:
I went rogue and already deployed that to the actual website :hand_over_mouth: 
https://fe-lang.org/docs/quickstart/deploy_contract.html

g-r-a-n-t (approved):

PR #856, opened 2023-03-07
std::abi

g-r-a-n-t:
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/library/std/src/abi.fe
+++ b/crates/library/std/src/abi.fe
@@ -11,6 +11,9 @@ pub trait AbiEncode {
     fn encode(self, mut writer: MemoryBufferWriter);
 }
 
+
+// unsigned integers
+
 impl AbiDecode for u256 {
     fn decode(mut reader: CalldataReader) -> Self {
         return reader.read_u256()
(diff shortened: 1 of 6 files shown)


Y-Nak:
Just a question: Is this PR targeting the current fe implementation? I thought you would work on the std for fe-v2 after testing implementation.

g-r-a-n-t:
> Just a question: Is this PR targeting the current fe implementation? I thought you would work on the std for fe-v2 after testing implementation.

this is some non-committal sketch code, ultimately Im targeting v2. is the v2 branch stable enough to work on?

Y-Nak:
I don't know how stable it should be for you to start working on it. At least the new parser will work. For semantics, I wouldn't add projection stuff, so basically, it's the same except for trait resolving, type inference, and pointer types.

As I said, some std implementations are necessary to implement v2, like encode/decode stuff. So, It'd be better for me if you start targeting v2 implementation. But it's up to you.

crates/library/std/src/abi.fe
```diff
@@ -0,0 +1,85 @@
...
+            let value: u256 = self.inner.decode(location, offset: offset + curr_offset).unwrap()
+            unsafe { evm::mstore(offset: mem_start + curr_offset, value) }
+            i += 1
+        }
+        return DecodeOutput::MPtr(mem_start)
+    }
+}
+
+pub enum AbiType {
+    Primative(AbiPrimative)
```
cburgdorf, line 70: typo?  Primative -> Primitive

g-r-a-n-t: Im just dumb

crates/library/std/src/buf.fe
```diff
@@ -297,3 +297,48 @@ pub struct RawCallBuffer {
         return self.buf.writer()
     }
 }
+
```
g-r-a-n-t, line 300: delete this

PR #859, opened 2023-03-15
Modernize uniswap demo code

cburgdorf:
Just a few tiny modernizations in the uniswap demo that I spotted while working on something else. In the future the whole demo should probably be converted to be split over multiple files to not repeat certain structs etc.


crates/test-files/fixtures/demos/uniswap.fe
```diff
@@ -1,10 +1,11 @@
+// Poor man's interface because current Fe has no interfaces yet
 contract ERC20 {
     pub fn balanceOf(self, _ account: address) -> u256 {
-        return 0
+        revert
```
cburgdorf, line 4: Made it slightly clearer that this is meant to be an interface

sbillig (approved):
:+1:

Issue #809, opened 2022-11-15
Allow use of primitive type constructors in const context


sbillig:
### What is wrong?

It's not possible to define a `const` address:
```
error: calling function outside function
  ┌─ /home/sean/code/eth/const.fe:1:23
  │
1 │ const ADDR: address = address(0x0)
  │                       ^^^^^^^ function can only be called inside function
```

### How can it be fixed

Loosen the "function can only be called inside function" error check to allow for primitive type constructor calls.
Contract type constructors should also be allowed, eg `const TRUSTED_CONTRACT: Foo = Foo(address(0x1234))`

PR #862, opened 2023-03-28
Allow calls to type constructors on primitives to appear in constant expressions

cburgdorf:
### What was wrong?

As stated in #809 one very annoying thing that arises in pretty much every attempt to write real world Fe code is that it is currently not possible to use addresses as constants.

### How was it fixed?

Hacked until it worked.

### DIFF
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -1381,7 +1381,7 @@ fn expr_call_pure(
 
     if function.is_test(context.db()) {
         context.fancy_error(
-            &format!("`{}` is a test function", fn_name),
+            &format!("`{fn_name}` is a test function"),
             vec![Label::primary(call_span, "test functions are not callable")],
             vec![],
         );
--- a/crates/analyzer/src/traversal/expressions.rs
+++ b/crates/analyzer/src/traversal/expressions.rs
@@ -905,13 +905,13 @@ fn expr_call(
         );
     }
 
-    if context.is_in_function() {
+    if context.is_in_function() || call_type.is_primitive_type_constructor(context.db()) {
         context.add_call(func, call_type);
     } else {
         context.error(
             "calling function outside function",
             func.span,
-            "function can only be called inside function",
+            "ffunction can only be called inside function",
         );
     }
 
--- a/crates/analyzer/src/context.rs
+++ b/crates/analyzer/src/context.rs
@@ -577,6 +577,14 @@ impl CallType {
             self.function().map(|id| id.is_unsafe(db)).unwrap_or(false)
         }
     }
+
+    pub fn is_primitive_type_constructor(&self, db: &dyn AnalyzerDb) -> bool {
+        if let CallType::TypeConstructor(type_id) = self {
+            type_id.is_primitive(db)
+        } else {
+            false
+        }
+    }
 }
 
 impl fmt::Display for CallType {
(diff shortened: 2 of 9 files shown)


Y-Nak:
Please give me a day to estimate how much this change affects the fe-v2 implementation. I guess this case would be fine.
But I want to share my feeling that I'm afraid fe-v2 implementation will never end if I have to implement all arbitrary changes in a "proper" way.

cburgdorf:
@Y-Nak Sure, no worries. I don't mean to burden your work but this issue is a real pain when writing real world Fe code so I hope we can resolve it before the next release.

Y-Nak (approved):
I considered the impact and concluded it'd be fine as long as `const` is limited to top-level declarations. We need to consider it more when it is allowed to be defined inside other items like `Function` or so. So I'm happy for the PR to be merged.

sbillig:
I think that instead of allowing primitive type constructors/casts, we should just interpret the value using the declared type, and error on values that don't fit that type.

Eg `const FOO: address = 0xfefefefe` would be allowed, but `const BAR: u8 = 1000` would be rejected.

cburgdorf:
@sbillig That's what I initially tried to do but it wasn't obvious to me how that would work.

 `const BAR: u8 = 1000` is rejected because `1000` is parsed as a `Expr::Num(num_str)` and so we check the `num_strut in` against the permitted size directly. But in the case of `const FOO: address = 0xfefefef` the `0xfefefef` is again parsed as an `Expr::Num(num_str)` and since `address` isn't a numeric type I'm not sure how exactly we would use it to constrain the size of the numeric expression. We could say that we can *coerce* a `u256` into an address but since `address` is a smaller type than `u256` it doesn't seem right to just truncate left or right.
 
A cleaner way imho would be to add `u160` as a numeric type and make `address` an alias to it.

crates/analyzer/src/traversal/expressions.rs
```diff
@@ -905,13 +905,13 @@ fn expr_call(
...
 
-    if context.is_in_function() {
+    if context.is_in_function() || call_type.is_primitive_type_constructor(context.db()) {
         context.add_call(func, call_type);
     } else {
         context.error(
             "calling function outside function",
             func.span,
-            "function can only be called inside function",
+            "ffunction can only be called inside function",
```
sbillig, line 914: rogue f

sbillig:
> `const BAR: u8 = 1000` is rejected because `1000` is parsed as a `Expr::Num(num_str)` and so we check the `num_str` against the permitted size directly. But in the case of `const FOO: address = 0xfefefef` the `0xfefefef` is again parsed as an `Expr::Num(num_str)` and since `address` isn't a numeric type I'm not sure how exactly we would use it to constrain the size of the numeric expression. We could say that we can _coerce_ a `u256` into an address but since `address` is a smaller type than `u256` it doesn't seem right to just truncate left or right.
> 
> A cleaner way imho would be to add `u160` as a numeric type and make `address` an alias to it.

A think Address should be a `newtype` rather than an alias, but it could certainly be defined in std in terms of u160.

For now, we can just constrain it to a u160 internally.  I think this is all doable without much trouble, by checking for `Type::Address` in `traversal::expressions::expr_num` and `Constant::from_num_str`, and checking if the string can be interpreted as a u160. This of course, would allow the use of a number literal anywhere an `address` is expected. I started along this path, but was using H160 which won't work because it expects a full 40-char hex string. I'm happy to continue, but also happy to have you do it instead :)

(Aside: perhaps non-hex address literals should be disallowed?)

cburgdorf:
@sbillig sounds good to me!

>I'm happy to continue

Yes, please take it, I'm happy to review it then.

crates/analyzer/src/traversal/expressions.rs
cburgdorf: oops

PR #864, opened 2023-04-03
address literals

### LINKED ISSUE #809: Allow use of primitive type constructors in const context

sbillig:
### What was wrong?

'twas impossible to define a `const` address, due to address values needing an explicit `address(...)` cast, which is treated as a function call internally, and thus not allowed in a `const` context yet.

Closes #809 (not exactly as stated, but with this change it's possible to define all primitive types in a const context so we don't need the type constructors/casts). This is an alternative to #862.

### How was it fixed?

Allow number literals to be interpreted as an addresses. Anywhere an `address` type is expected, a number literal is allowed (as long as it fits in a u160).
```rust
const FOO: address = 0xfe // OK!
fn foo(a: address) { .. }
foo(100) // OK! 
// ^ note that this is decimal 100! Perhaps we should only allow hex literals?
```

This might feel a bit loosey-goosey, but I think it'll be fine in real code, which likely won't often use address literals for anything other than 0 and `const` definitions.

### DIFF
--- a/crates/analyzer/src/context.rs
+++ b/crates/analyzer/src/context.rs
@@ -589,6 +589,7 @@ impl fmt::Display for CallType {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Constant {
     Int(BigInt),
+    Address(BigInt),
     Bool(bool),
     Str(SmolStr),
 }
--- a/crates/analyzer/src/namespace/types.rs
+++ b/crates/analyzer/src/namespace/types.rs
@@ -33,6 +33,10 @@ pub fn i256_min() -> BigInt {
     BigInt::from(-2).pow(255)
 }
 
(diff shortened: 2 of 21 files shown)


cburgdorf (approved):
Fine by me :+1:

Y-Nak (approved):
LGTM!

PR #865, opened 2023-04-07
Fix issue where generic params that are `mut` can never be satisfied by the caller

cburgdorf:
### What was wrong?

If generic parameters are marked as `mut` there is currently no way to call them as the analyzer doesn't recognize them as the same type.

### How was it fixed?

Threw a bunch more `deref()` around


sbillig (approved):
👍

PR #868, opened 2023-04-11
Fix ICE

cburgdorf:
### What was wrong?

This code currently crashes the compiler

```fe
struct Tx {
  pub data: Array<u8, 320>
}

contract Foo {
  pub fn bar(mut tx: Tx) {}
}
```

### How was it fixed?

Deref mptrs in `abi_type_minimum_size` and `abi_type_maximum_size`


sbillig (approved):
👍

PR #874, opened 2023-04-19
poc: remove Rc Cell from parser2 parse scope structs

sbillig:
I've been wasting my time profiling the new parser. I noticed that it does a decent amount of allocating, and figured there may be easy speed wins. I'm fully aware that no one cares about this and I have better things to do with my life.

Here's an ugly poc of a minor improvement from a simple change; putting it here so that it doesn't fade away in my pile of git stashes.

Parsing a slightly modified uniswap contract:
`739.23 µs` -> `647.37 µs`

(This pr is based on #860)


Y-Nak:
Nice! This is Interesting.
I tried to run rust-analyzer's benchmark for its parser.
It takes around 20-40ms for 8000 lines file(you may need to modify the test code slightly, it might have a bug).
The command was 
```bash
// in rust-analyzer/crates/syntax
RUN_SLOW_BENCHES=1 cargo test --release -- --nocapture --test-threads=1
```

Also, I tried running a benchmark on a file with about 9,000 lines file made by repeating `uniswap-paraser2.fe` about 20 times. It seemed to take approximately 20 ms to execute.

So, it depends on the later passes, but I guess it should be fine for now.

sbillig:
I agree, but if we remove the Rc Cell, it could take about *17ms!* ;)

PR #879, opened 2023-04-28
Expose bug

cburgdorf:
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/crates/tests/fixtures/files/arrays.fe
+++ b/crates/tests/fixtures/files/arrays.fe
@@ -1,10 +1,47 @@
 use std::evm
 
+const ZERO_ADDRESS: address = 0x0000000000000000000000000000000000000000
+
+struct AddressArray {
+    array: Array<address, 2>
+    length: u256
+
+    pub fn new() -> AddressArray {
+        return AddressArray (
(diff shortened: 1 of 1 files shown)


crates/tests/fixtures/files/arrays.fe
```diff
@@ -1,10 +1,47 @@
...
     pub fn __init__(mut self) {
         self.my_array = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        let mut addr_array: AddressArray = AddressArray::new()
+        addr_array.push(value: 0x627306090abaB3A6e1400e9345bC60c78a8BEf57)
+        addr_array.push(value: 0xf17f52151EbEF6C7334FAD080c5704D77216b732)
+        self.addr_array = addr_array
+    }
+
+    pub fn get_address_from_storage(self, index: u256) -> address {
+        return self.addr_array.get(index)
```
cburgdorf, line 44: The test works if we change that to be `self.addr_array.to_mem().get(index)`. I believe the compiler rejected that in the past but doesn't anymore.

cburgdorf:
This is fixed via https://github.com/ethereum/fe/pull/881

PR #880, opened 2023-05-01
`ctx` param in tests

g-r-a-n-t:
### What was wrong?

`ctx`  couldn't be used in tests functions without a bit of hacking.

### How was it fixed?

modified checks around the `ctx` param and updated tests.

### To-Do

- [x] Clean up commit history


sbillig (approved):
👍

PR #882, opened 2023-05-02
chore: update website copyright notice date

Lemma199:
### What was wrong?
I was browsing the Fe language [website](https://fe-lang.org/) and noticed the copyright notice was out of date. This is not a huge concern, but considering the website has been updated in 2022 and 2023 updating the date to a range to reflect that seems like a good idea.

### How was it fixed?
Updated the copyright notice in the footer of the website to be a range from 2021 (website first published) to 2023 (current year)


cburgdorf:
Thanks for that!

PR #885, opened 2023-05-11
MemoryBuffer, precompiles, raw_call

g-r-a-n-t:
added the following to the std lib:
- memory buffer abstraction
- support for all precompile calls
- implemented `ctx.raw_call(...)`

### To-Do

- [x] Clean up commit history


crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,43 @@
+use ingot::evm
+
+unsafe fn avail() -> u256 {
```
cburgdorf, line 3: We should probably expose that as an unsafe function somewhere outside of the memory buffer.

cburgdorf: Yep, I was expecting it to probably go away in the future with the rewrite. I was just thinking about the *current* Fe because this function may end up being duplicated across a bunch of different places in the std library.

g-r-a-n-t: I was thinking we should only use functions in a single lower level type and then swap out the internals when v2 lands

cburgdorf: I guess I'm just generally more ok to break user code while we are in `0.x` land. As I see it, users shouldn't expect API stability in the current phase of the project. 😉

Y-Nak: From my perspective, this is not a simple API stability problem.

crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+unsafe fn alloc(len: u256) -> u256 {
+    let ptr: u256 = avail()
+    evm::mstore(offset: 64, value: ptr + len)
+    return ptr
+}
+
+struct Cursor {
```
g-r-a-n-t, line 19: add docs everywhere

crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,281 @@
...
+                break 
+            } else {
+                writer.write_u8(value: self.read_u8())
+            }
+        }
+
+        return buf
+    }
+
+    // todo: fix compiler panic produced by this
```
g-r-a-n-t, line 236: report this if not a duplicate issue

crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,281 @@
...
+    // pub fn write<T: MemoryBufferWrite>(mut self, value: T) {
+    //     self.cur += value.write_buf(writer: self)
+    // }
+}
+
+
+// pub struct MemoryBufferReader {
```
g-r-a-n-t, line 243: delete

crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,281 @@
...
+//     pub fn read(mut self) -> u256 {
+//         unsafe {
+//             let value: u256 = evm::mload(offset: self.buf().offset() + self.cur)
+//             self.cur += 32 
+//             return value
+//         }
+//     }
+// }
+
+pub trait MemoryBufferWrite {
```
g-r-a-n-t, line 261: comment out

crates/library/std/src/buf.fe
```diff
@@ -0,0 +1,281 @@
...
+    fn write_buf(self, mut writer: MemoryBufferWriter) -> u256;
+}
+
+// todo: fix `Self` is not defined according to compiler
+// trait MemoryBufferRead {
+//     fn read_buf(mut buf: MemoryBufferReader) -> (Self, u256);
+// }
+
+// impl MemoryBufferWrite for u256 {
```
g-r-a-n-t, line 270: update code

crates/tests/fixtures/files/precompiles.fe
```diff
@@ -0,0 +1,148 @@
...
+fn test_ripemd_160() {
+    let buf: MemoryBuffer = MemoryBuffer::from_u8(value: 0xff) 
+    // todo: report "not yet implemented" panic with extra ')'
+    let result: u256 = precompiles::ripemd_160(buf)
+    assert result == 0x2c0c45d3ecab80fe060e5f1d7057cd2f8de5e557
+}
+
+// todo: require no return value
+#test
+// todo: yulgen panic on `identity`
```
g-r-a-n-t, line 33: report

crates/library/std/src/context.fe
```diff
@@ -86,7 +87,21 @@ pub struct Context {
         }
     }
 
+    pub fn raw_call(self, addr: address, value: u256, mut buf: MemoryBuffer) {
+        unsafe {
+            assert evm::call(
```
cburgdorf, line 92: Ignore this, if you already thought about this but is `assert` the right thing to do here? I believe we should just return a `bool` here and let the caller handle that. For instance, in the multisig wallet that I'm working on I'll have to inspect the return and need to emit `Success` or `Failure` events based on the outcome. But if `raw_call` does already internally fail with an assertion error, then I believe there would be no way to handle this gracefully, right?

g-r-a-n-t: returning bool here makes sense

crates/library/std/src/math.fe
```diff
@@ -0,0 +1,15 @@
+pub fn min(_ x: u256, _ y: u256) -> u256 {
```
cburgdorf, line 1: I just wanted to suggest that we could define an `Ord` trait and then have the `min` / `max` functions be generic and work with any `T` that is `Ord` but free floating functions such as these can not use generics yet as we only allow struct methods to be generic for now....so, nothing to see here.

cburgdorf (approved):
Looks nice. Just make sure to file the issues if you haven't already.

PR #886, opened 2023-05-17
solc v0.8.22; shanghai, push0 support

sbillig:
### What was wrong?

No support for push0, causing the wastage of *several* gas.

This updates to solc 0.8.22. I think it reduces gas for all simple tests, but unfortunately increases gas usage for the uniswap createPair test. I plan to look into that case and see what's up.

### How was it fixed?

solc ~~0.8.20~~ 0.8.22
evm crate had to be upgraded for tests; to keep the tests working I had to hack around a bunch of changes.
solc fails to generate json for the "irAst" output in cases with a lot of nesting in our generated Yul code, which happens for tests with a lot of `assert`s (binary_ops.fe and comp_ops.fe, in particular). I had to disable this output. (I'll create an issue for our perhaps less-than-ideal codegen)


crates/test-utils/src/lib.rs
```diff
@@ -161,18 +163,40 @@ impl ContractHarness {
...
+                let calldata = self.build_calldata(name, input);
+                let transact_cost = transaction_cost(
+                    gasometer::call_transaction_cost(&calldata, &[]),
+                    executor.config(),
+                );
+
+                // The simple_open_auction `withdraw` call somehow uses less gas
+                // than the reported transaction cost.
+                // We're changing all this testing code anyway so I'm just gonna
+                // hack around this. Don't know what's going on.
```
cburgdorf, line 183: Who am I to judge 😅🤭

cburgdorf (approved):

cburgdorf:
If it turns out that this will change the runtime code of the recently deployed contracts, then I will cut two releases just to make sure people have a way to verify the already deployed contracts.

cburgdorf:
Ok, I verified that this does indeed result in different bytecode. I'm going to cut a release now without this PR and merge it after that. We can then cut another release shortly before devconnect that includes this and potentially other stuff.

cburgdorf (approved):

sbillig:
> Ok, I verified that this does indeed result in different bytecode. I'm going to cut a release now without this PR and merge it after that. We can then cut another release shortly before devconnect that includes this and potentially other stuff.

This doesn't have to be merged or released before devconnect; it can wait for some future release. I'm just trying to tie up some loose ends.

PR #892, opened 2023-05-26
Get MemoryBufferWrite trait working

cburgdorf:
### What was wrong?

The `MemoryBufferWrite` trait wasn't working.

### How was it fixed?

There wasn't an issue with the trait itself but our `mir_lower_std_lib` tests call all generic functions with an `()` type and hence they expect an `impl` for the `()` type.


crates/library/std/src/buf.fe
```diff
@@ -97,71 +97,82 @@ pub struct MemoryBufferWriter {
...
+
+impl MemoryBufferWrite for u8 {
+    fn write_buf(self, mut writer: MemoryBufferWriter) {
+        let offset: u256 = writer.write_offset(len: 1)
+        unsafe { evm::mstore8(offset, value: self) }
+    }
+}
+
+// This is needed to prevent the `mir_lower_std_lib` to crash the compiler
+impl MemoryBufferWrite for () {
```
cburgdorf, line 173: @g-r-a-n-t The code itself was fine but we need this to prevent the `mir_lower_std_lib` crashing the compiler. This would note have happened in user code. It's a special annoyance when writing std lib code that the `mir_lower_std_lib` tests walk all functions and invoke generic functions with the `()` type.

cburgdorf:
@g-r-a-n-t Got it working. We can not yet do the same for the `read_x` APIs because we do not yet support functions to return generic values.

g-r-a-n-t (approved):
looks good

Issue #891, opened 2023-05-25
`std::evm` functions should take buffer types


g-r-a-n-t:
### What is wrong?

A number of functions defined in [std::evm](https://github.com/ethereum/fe/blob/master/crates/library/std/src/evm.fe) take offsets and lengths for regions of memory and calldata. These should take [buffer](https://github.com/ethereum/fe/blob/master/crates/library/std/src/buf.fe) types instead.

### How can it be fixed

- implement `CalldataBuffer`, `CalldataBufferReader`
- update `revert_mem`, `return_mem`, `call_data_copy`, `code_copy`, `ext_code_copy`, `keccak256_mem`, `create`, `create2`, `call`, and `log` 

run `$ cargo test std_lib` to check changes made to `crates/library`

Hrushi20:
Hey, can I work on this issue? Can you give me a brief idea about `CalldataBuffer` and `CalldataBufferReader`, so that I can start working on this issue.

g-r-a-n-t:
hey sorry for the late reply @Hrushi20

this work has been completed, but I could help you find another task if you're interested

Hrushi20:
Hey! I'm interested. Any beginner friendly issues I can get started with?

g-r-a-n-t:
would you be interested in adding support for the transient storage opcodes?

g-r-a-n-t:
here, just wrote an issue https://github.com/ethereum/fe/issues/978

PR #893, opened 2023-06-04
 std::evm functions should take buffer types #891 (Part 1)

saifalkatout:
### What was wrong?
A number of functions defined in [std::evm](https://github.com/ethereum/fe/blob/master/crates/library/std/src/evm.fe) take offsets and lengths for regions of memory and calldata. These should take [buffer](https://github.com/ethereum/fe/blob/master/crates/library/std/src/buf.fe) types instead.

### How was it fixed?

updated revert_mem, return_mem,  ext_code_copy, keccak256_mem, create, create2, call, and log


crates/library/std/src/evm.fe
```diff
@@ -268,24 +270,24 @@ pub unsafe fn ext_code_hash(_ addr: address) -> u256 {
...
-    return address(__create2(v, p, n, s))
+pub fn create2(value v: u256, buffer: buf::MemoryBuffer, salt s: u256) -> address {
+    unsafe { return address(__create2(v, buffer.offset(), buffer.len(), s)) }
 }
 
 // TODO: return bool (success)
-pub unsafe fn call(gas: u256, addr: address, value: u256, input_offset: u256, input_len: u256, output_offset: u256, output_len: u256) -> u256 {
-    return __call(gas, u256(addr), value, input_offset, input_len, output_offset, output_len)
+pub fn call(gas: u256, addr: address, value: u256, buffer: buf::RawCallBuffer) -> bool {
+  unsafe{ return __call(gas, u256(addr), value, buffer.offset(), buffer.input_len(), buffer.offset(), buffer.output_len()) == 1 }
```
g-r-a-n-t, line 290: the `buffer` parameter doesnt actually need to be mutable here for this to work, but `__call` will in fact mutate the buffer, which is why the `buf` param on `raw_call` was `mut` and should be here as well

g-r-a-n-t: since we are using `unsafe`, we need to make sure the semantics of the language still hold

crates/library/std/src/context.fe
```diff
@@ -96,18 +95,15 @@ pub struct Context {
         self, 
         addr: address, 
         value: u256, 
-        mut buf: RawCallBuffer 
+        buffer: RawCallBuffer
```
g-r-a-n-t, line 98: I would personally like to continue using the `buf` identifier

g-r-a-n-t:
it looks like the tests are failing because new snaposhots have been generated. try `cargo insta review`

https://docs.rs/insta/latest/insta/

overall this looks great

crates/library/std/src/evm.fe
saifalkatout: Is it enough to check changes in the compiled JSON files, or do we need to deploy to a testnet ?

g-r-a-n-t: in reference to the snapshot changes?

saifalkatout: in reference to the semantic changes.

g-r-a-n-t: ok, Im not quite sure I follow.. could you please explain further?

saifalkatout: So one of the comments entailed that we need to check the language semantics # # - because we are using unsafe, how would we go about testing/checking this ??

g-r-a-n-t: oh, what I mean is that the implementation of `evm::call` will mutate the `buf` parameter, even though `buf` does not need to be mutable for it's use in the `__call` intrinsic

it's important that `mut` semantics hold when using `evm::call`, so we need to set the `buf` parameter as mutable, that way someone using the `evm::call` function will have to set the `buf` to mutable.

so..

```
pub fn call(gas: u256, addr: address, value: u256, buf: buf::RawCallBuffer) -> bool {
  unsafe{ return __call(gas, u256(addr), value, buffer.offset(), buf.input_len(), buf.offset(), buffer.output_len()) == 1 }
```

becomes

```
pub fn call(gas: u256, addr: address, value: u256, mut buf: buf::RawCallBuffer) -> bool {
  unsafe{ return __call(gas, u256(addr), value, buffer.offset(), buf.input_len(), buf.offset(), buffer.output_len()) == 1 }
}
```

that way someone writing the following

```
let buf: RawCallBuffer = ...
evm::call(..., buf)
```

will get a compile error asking them to make `buf` mutable, like so

```
let mut buf: RawCallBuffer = ...
evm::call(..., buf)
```

---

`unsafe` code allows you to do things that are normally disallowed in safe code, such as direct memory mutation, which is what we are doing in `evm::call`. this is fine as long as we correctly set the parameters that end up getting mutated as `mut`

PR #898, opened 2023-06-17
Buf write simplification

g-r-a-n-t:
### What was wrong?

The buf write implementation seems unnecessarily complicated.

Currently we allocate the exact `len` in memory for the buffer and ensure via `rewrite_slot` that we do not modify data outside of the buf's allocated region.

This PR changes the buf implementation so that an extra 31 bytes is allocated for bufs. This eliminates the need for a function like `rewrite_slot`.

### How was it fixed?

:keyboard: 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/library/std/src/buf.fe
+++ b/crates/library/std/src/buf.fe
@@ -47,7 +47,7 @@ pub struct MemoryBuffer {
 
     pub fn new(len: u256) -> Self {
         unsafe {
-            return MemoryBuffer(offset: alloc(len), len)
+            return MemoryBuffer(offset: alloc(len: len + 30), len)
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/library/std/src/buf.fe
```diff
@@ -47,7 +47,7 @@ pub struct MemoryBuffer {
 
     pub fn new(len: u256) -> Self {
         unsafe {
-            return MemoryBuffer(offset: alloc(len), len)
+            return MemoryBuffer(offset: alloc(len: len + 31), len)
```
cburgdorf, line 50: Couldn't we check if `len` is a multiple of `32` and then not allocate 31 superfluous bytes?

cburgdorf (approved):

crates/library/std/src/buf.fe
g-r-a-n-t: it's actually not a multiple of 32 thing, as the buffer can be written to with anywhere between 1-32 bytes. problem is that we can only write 1 or 32 bytes at the evm level, so it needs to be padded `32 - <smallest write size>` where the "smallest write size" is the the smallest number of bytes that we write at the buffer level using the 32 byte write operation

Im realizing now that since this "smallest write size" is actually 2 bytes for `u16`, the number should isntead be 30

cburgdorf: Uuuh..I think I finally get it!

PR #901, opened 2023-07-05
Fix issue with imported constants

cburgdorf:
### What was wrong?

I ran into an issue were constants were not treated as such if they were imported from another file.

### How was it fixed?

changed `all_constants` to include those that are imported via `use`  statements so that subsequent calls to `resolve_constant` work as expected.


crates/test-files/fixtures/ingots/imported_constant/src/main.fe
```diff
@@ -0,0 +1,8 @@
+use constants::MAX_ITEMS
+
+contract Foo {
```
cburgdorf, line 3: This would be nice to use the new `#test` support but I noticed ingot tests aren't yet supported and I didn't feel like diving into it.

cburgdorf:
I'm closing this for now because there's something not quite right.

crates/test-files/fixtures/ingots/imported_constant/src/main.fe
g-r-a-n-t: hmm. I should implement this

cburgdorf: I have that working locally but it needs further cleanup.

cburgdorf: Opened a PR with what I have https://github.com/ethereum/fe/pull/904

PR #903, opened 2023-07-07
Calldata reader 

saifalkatout:
Implementation of CalldataReader. An instance of CalldataReader can be created using ctx.calldata_reader().

### DIFF
--- a/crates/library/std/src/context.fe
+++ b/crates/library/std/src/context.fe
@@ -32,7 +32,7 @@ pub struct CalldataReader {
     }
 
     pub fn advance(mut self, len: u256) -> u256 {
-        self.cur_offset += 32
+        self.cur_offset += len
         assert self.cur_offset <= self.len
         return self.cur_offset
     }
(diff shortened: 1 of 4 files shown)


crates/library/std/src/context.fe
```diff
@@ -19,6 +19,60 @@ pub trait Emittable {
...
+        self.cur_offset += 32
+        assert self.cur_offset <= self.len
+        return self.cur_offset
+    }
+
+    fn read_n(mut self, len: u256) -> u256 {
+        unsafe {
+            let value: u256 = evm::call_data_load(offset: self.cur_offset)
+            self.advance(len)
+            if len < 32 {
```
saifalkatout, line 44: The if statement here is somewhat costly, is the increased readability/abstration of making all the read_ functions use read_n worth it ??

g-r-a-n-t: the `read_u256` function could just use `call_data_load` itself and then this if statement can be removed

saifalkatout: Wouldn't that be redundant and reduce readability ??

g-r-a-n-t (approved):
nice work

PR #909, opened 2023-08-02
Improve error recovery in parsing fields definition

Y-Nak:
NOTE: Please start reviewing this PR after the name resolution PR is merged.

This PR adds the ad-hoc check to the parser to improve error messages when a function is defined inside a `struct/contract` definition.

The reason for this ad-hoc check is
1. This error is difficult to recover properly with the current parser design, and the emitted error message is confusing.
2. We anticipate that this error would often happen in the transition period to Fe-V2.


Y-Nak:
@sbillig would you review this ugly PR?

sbillig (approved):
👍

PR #911, opened 2023-08-04
Improve marker traits usability

Y-Nak:
NOTE: Please start reviewing after https://github.com/ethereum/fe/pull/909 is merged

The `SpannedHirDb` and `LowerHirDb` marker traits required `Self: Sized` in `as_lower_hir_db()` and `as_lower_hir_db()` methods.
This broke consistency with other db traits that implement `DbWithJar`. So I decided to add two jars for these two traits respectively.

As a result, users for these traits no longer need to implement `SpannedHirDb` and `LowerHirDb` manually; instead, those traits are automatically implemented when `#[salsa::db(hir::SpannedJar, hir::LowerJar)]` is specified.

E.g., the driver database would be like the below.
```rust
#[salsa::db(
    common::Jar,
    hir::Jar,
    hir::LowerJar,
    hir::SpannedJar,
    hir_analysis::Jar,
    Jar
)]
pub struct DriverDataBase {
    storage: salsa::Storage<Self>,
}
```

### DIFF
--- a/crates/driver2/src/diagnostics.rs
+++ b/crates/driver2/src/diagnostics.rs
@@ -6,22 +6,22 @@ use cs::{diagnostic as cs_diag, files as cs_files};
 
 use common::{
     diagnostics::{LabelStyle, Severity},
-    InputFile,
+    InputDb, InputFile,
 };
 use hir::diagnostics::DiagnosticVoucher;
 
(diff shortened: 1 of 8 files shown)


sbillig (approved):
👍

PR #912, opened 2023-08-06
Implement `Eq` in Fe 

saifalkatout:
### What was wrong?
Equality was using the yul backend

### How was it fixed?
Change equality to be fully implemented in Fe

### DIFF
--- a/crates/driver/src/lib.rs
+++ b/crates/driver/src/lib.rs
@@ -28,6 +28,7 @@ pub struct CompiledContract {
 }
 
 #[cfg(feature = "solc-backend")]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CompiledTest {
     pub name: SmolStr,
     bytecode: String,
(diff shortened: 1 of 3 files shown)


crates/library/std/src/buf.fe
```diff
@@ -1,6 +1,6 @@
 use ingot::evm
 use ingot::math
-
+use ingot::cmp::Eq
 unsafe fn avail() -> u256 {
```
g-r-a-n-t, line 4: put a line break here plz

crates/library/std/src/cmp.fe
```diff
@@ -0,0 +1,92 @@
...
+
+impl Eq for u8 {
+    fn compare(self, other: Self) -> bool {
+        return evm::eq(self, other)
+    }
+}
+
+impl Eq for i256 { 
+   fn compare(self, other: Self) -> bool {
+        return evm::eq(u256(self), u256(other))
```
g-r-a-n-t, line 44: try testing this with negative numbers

crates/library/std/src/cmp.fe
```diff
@@ -0,0 +1,92 @@
+use ingot::evm
+pub trait Eq {
+    fn compare(self, other: Self) -> bool;
+}
+
+impl Eq for u256 {
+    fn compare(self, other: Self) -> bool {
+        return evm::eq(self, other)
```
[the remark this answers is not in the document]
saifalkatout: When trying to send u256 into a u16 impl:

image

When trying to send u16 into a y256 impl, type coercion occurs

Issue #913, opened 2023-08-11
The fe test command should default to run from the project directory


cburgdorf:
### What is wrong?

Currently we need to run tests with `fe test .` (notice the dot for the current directory) even if we are running it from the base directory of the fe project.

### How can it be fixed

It should automatically provide the project base directory even if `fe test` is used from a directory somewhere deeper within the fe project  structure (e.g. `src/some/where`)

PR #915, opened 2023-08-12
Use project root as default path for `fe test`

cburgdorf:
### What was wrong?

As described in #913, `fe test` should always fallback to assume the root project directory if no path is specified.

### How was it fixed?

JUST DID IT! LIKE A PSYCHOPATH.


sbillig (approved):
👍

PR #916, opened 2023-08-22
v2: enum record variant support; one-liner record def

sbillig:
v2 parser and HIR now allow record-type enum variants.

Struct (and record enum variant) fields must now be separated by a comma (newlines are allowed but not required). This allows a struct with multiple fields to be defined on a single line, eg `struct S { x: u8, y: u8 }`.

Nit-picky feedback welcome.

### DIFF
--- a/crates/hir/src/visitor.rs
+++ b/crates/hir/src/visitor.rs
@@ -6,8 +6,9 @@ use crate::{
         FieldDef, FieldDefListId, FieldIndex, Func, FuncParam, FuncParamLabel, FuncParamListId,
         FuncParamName, GenericArg, GenericArgListId, GenericParam, GenericParamListId, IdentId,
         Impl, ImplTrait, ItemKind, LitKind, MatchArm, Mod, Partial, Pat, PatId, PathId, Stmt,
-        StmtId, Struct, TopLevelMod, Trait, TypeAlias, TypeBound, TypeId, TypeKind, Use, UseAlias,
-        UsePathId, UsePathSegment, VariantDef, VariantDefListId, WhereClauseId, WherePredicate,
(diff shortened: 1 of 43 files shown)


sbillig:
@Y-Nak I added the enum record variant fields to the scope graph, in the way that we decided wasn't very elegant 🙃. This seemed nicer than adding a lookup from `ScopeId(u32)` to `ScopeKind`

crates/hir/src/visitor.rs
```diff
@@ -1559,13 +1569,15 @@ pub fn walk_variant_def<V>(
...
-        )
+    match variant.kind {
+        VariantKind::Unit => {}
+        VariantKind::Tuple(t) => ctxt.with_new_ctxt(
+            |span| span.tuple_type_moved(),
+            |ctxt| visitor.visit_tuple_type(ctxt, t),
+        ),
+        VariantKind::Record(_) => {
+            todo!()
+        }
```
Y-Nak, line 1580: It seems you forgot here.

crates/parser2/src/parser/mod.rs
```diff
@@ -463,6 +463,16 @@ impl<S: TokenStream> Parser<S> {
         ErrorScope::default()
     }
 
+    fn error_at_current_pos(&mut self, msg: &str) -> ErrorScope {
```
Y-Nak, line 466: 👍

Y-Nak (approved):
Great!

I feel `comma` should be required for an enum definition as well if it is required for a struct definition. I want to hear your opinion.

sbillig:
> Great!
> 
> I feel `comma` should be required for an enum definition as well if it is required for a struct definition. I want to hear your opinion.

I agree; I'll make this change separately.

Y-Nak (approved):
LGTM!

PR #917, opened 2023-08-23
Finish buf refactor

g-r-a-n-t:
### What was wrong?

There were still some `std::evm` functions that used memory offset and length parameters instead of `MemoryBuffer`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/library/std/src/buf.fe
+++ b/crates/library/std/src/buf.fe
@@ -112,7 +112,7 @@ pub struct MemoryBuffer {
     }
 
     /// The start of the buffer in EVM memory.
-    pub fn offset(self) -> u256 {
+    pub unsafe fn offset(self) -> u256 {
         return self.offset
     }
 
@@ -146,14 +146,16 @@ pub struct MemoryBufferWriter {
         return self.cur.remainder()
     }
 
-    pub fn write_offset(mut self, len: u256) -> u256 {
+    pub unsafe fn write_offset(mut self, len: u256) -> u256 {
         return self.buf.offset() + self.cur.advance(len)
     }
 
     pub fn write_n(mut self, value: u256, len: u256) {
-        let offset: u256 = self.write_offset(len)
         let shifted_value: u256 = evm::shl(bits: 256 - len * 8, value)
-        unsafe { evm::mstore(offset, value: shifted_value) }
+        unsafe { 
+            let offset: u256 = self.write_offset(len)
+            evm::mstore(offset, value: shifted_value) 
+        }
     }
 
     pub fn write_buf(mut self, buf: MemoryBuffer) {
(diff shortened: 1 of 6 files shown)


crates/library/std/src/evm.fe
```diff
@@ -2,181 +2,181 @@ use ingot::buf::{MemoryBuffer, RawCallBuffer}
...
 }
 
 pub fn sgt(_ x: u256, _ y: u256) -> u256 {
-  unsafe { return __sgt(x, y) }
+    unsafe { return __sgt(x, y) }
 }
 
 pub fn eq(_ x: u256, _ y: u256) -> u256 {
-  unsafe { return __eq(x, y) }
+unsafe { return __eq(x, y) }
```
saifalkatout, line 135: inconsistency ?

crates/library/std/src/evm.fe
```diff
@@ -285,21 +285,36 @@ pub fn create2(value v: u256, buf: MemoryBuffer, salt s: u256) -> address {
...
+    unsafe { return  __delegatecall(gas, u256(addr), buf.offset(), buf.input_len(), buf.offset(), buf.output_len()) == 1 }
 }
 
-pub unsafe fn static_call(gas: u256, addr: address, input_offset: u256, input_len: u256, output_offset: u256, output_len: u256) -> u256 {
-    return __staticcall(gas, u256(addr), input_offset, input_len, output_offset, output_len)
+pub fn static_call(gas: u256, addr: address, mut buf: RawCallBuffer) -> bool {
+    unsafe { return  __staticcall(gas, u256(addr), buf.offset(), buf.input_len(), buf.offset(), buf.output_len()) == 1 }
+}
+
+pub unsafe fn call_2(gas: u256, addr: address, value: u256, input_offset: u256, input_len: u256, output_offset: u256, output_len: u256) -> bool {
```
saifalkatout, line 304: I guess these functions are made for scenarios where manual memory allocation is required (for example in precompiles.fe)

crates/library/std/src/evm.fe
```diff
@@ -238,40 +238,40 @@ pub unsafe fn call_data_size() -> u256 {
     return __calldatasize()
 }
 
-pub fn call_data_copy(buf: MemoryBuffer, from_offset f: u256) {
-   unsafe { __calldatacopy(buf.offset(), f, buf.len()) }
+pub fn call_data_copy(mut buf: MemoryBuffer, from_offset f: u256) {
```
saifalkatout, line 241: There are more functions that accept immutable `buf`s (e.g `call_data_copy`, `log` , `return_mem`), do we need to change those also  ?

saifalkatout (approved):

crates/library/std/src/evm.fe
g-r-a-n-t: `call_data_copy`?? typo?

the other ones are not writing to the buf, so they can stay as immutable

crates/library/std/src/buf.fe
```diff
@@ -112,7 +112,7 @@ pub struct MemoryBuffer {
     }
 
     /// The start of the buffer in EVM memory.
-    pub fn offset(self) -> u256 {
+    pub unsafe fn offset(self) -> u256 {
```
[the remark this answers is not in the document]
g-r-a-n-t: Im not sure if it needs to be and we can always change it back. it's kind of like getting a raw pointer tho, so I figure it should be unsafe

PR #920, opened 2023-09-01
Eq trait

saifalkatout:
### What was wrong?
`==` is being converted into its yul equivalent.

### How was it fixed?
 Implemented `Eq` trait in Fe

### To-Do
- implement `Eq` in standard library structs


g-r-a-n-t:
is there a reason why `Eq` is not implemented on `MemoryBuffer` in this PR? we could add that and then have test cases for it

PR #929, opened 2023-09-15
[website/docs]: add note on homebrew to install page

jmcook1186:
Adds note to `src/user-guide/installation.md` about installing using homebrew

### DIFF
--- a/docs/src/user-guide/installation.md
+++ b/docs/src/user-guide/installation.md
@@ -40,7 +40,7 @@ SUBCOMMANDS:
 
 ## Package managers
 
-Mac users can also get Fe from [Homebrew](https://brew.sh/). The following command installs Fe and exposes it as `fe` without any further configuration necessary:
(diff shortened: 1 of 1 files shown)


docs/src/user-guide/installation.md
```diff
@@ -38,7 +38,16 @@ SUBCOMMANDS:
     new      Create new fe project
 ```
 
-# Building from source
+## Package managers
+
+Mac users can also get Fe from [Homebrew](https://brew.sh/). The following command installs Fe and exposes it as `fe` without any further configuration necessary:
```
cburgdorf, line 43: Homebrew is available for Linux as well and can even be installed on >= Windows 10 via the available Linux sub system.

jmcook1186:
```suggestion
Fe can also be installed from [Homebrew](https://brew.sh/). Homebrew is available for Mac, Linux and Windows (via [WSL](https://learn.microsoft.com/en-us/windows/wsl/about)). The following command installs Fe and exposes it as `fe` without any further configuration necessary:
```

cburgdorf (approved):

PR #930, opened 2023-09-15
[website/docs]: add open auction tutorial

jmcook1186:
Adds new tutorial, borrowing from the example contracts in the Fe Github.

Implements an open auction contract and uses it as a vehicle to explain some foundational Fe concepts. 
Also includes brief build & deploy instructions.


docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,258 @@
...
+Deploy it to a local blockchain using:
+
+```sh
+anvil
+```
+
+and
+
+```sh
+cast send --rpc-url localhost:8545 --private-key <your-private-key> --create $(cat output/Auction/Auction.bin) --constructor-args <bidding_time> <beneficiary_addr>
```
jmcook1186, line 232: Now realize this deploy command will not work but am afk - will update shortly.

jmcook1186: fixed

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+
+Here, you are not only using `self` but you are prepending it with `mut`. `mut` is a keyword inherited from Rust that indicates that the value can be overwritten - i.e. it is "mutable". Variables are not mutable by default - this is a safety feature that helps protect developers from unintended changes during runtime. If you do not make `self` mutable, then you will not be able to update the values it contains.
+
+
+#### Context
+
+Context is used to track deadlines for requests. It gives the contract the ability to cancel functions that have taken too long to run, and provides a structure for storing values that are scoped to a particular request so that they can be accessed anywhere in the call chain. This will be familiar to Go and Rust developers, as Context is often used in the same way in those languages and is frequently employed when working with external API calls. It is conventional to name the context object `ctx`.
```
cburgdorf, line 94: >Context is used to track deadlines for requests. It gives the contract the ability to cancel functions that have taken too long to run, and provides a structure for storing values that are scoped to a particular request so that they can be accessed anywhere in the call chain.

That part is incorrect. We don't track deadlines for requests or have the ability to cancel functions that take too long (in fact, that doesn't make sense in the context of the EVM as functions can not reach out for things external to the blockchain and everything is ultimately bounded by gas usage).

Context is simply a way for us to make certain access explicit and hence not need extra keywords such as `view` or `pure` like solidity does. Here's a good read on the history of context. I recommend to read it and then re-write this part: https://github.com/ethereum/fe/issues/558

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+
+Here, you are not only using `self` but you are prepending it with `mut`. `mut` is a keyword inherited from Rust that indicates that the value can be overwritten - i.e. it is "mutable". Variables are not mutable by default - this is a safety feature that helps protect developers from unintended changes during runtime. If you do not make `self` mutable, then you will not be able to update the values it contains.
+
+
+#### Context
+
+Context is used to track deadlines for requests. It gives the contract the ability to cancel functions that have taken too long to run, and provides a structure for storing values that are scoped to a particular request so that they can be accessed anywhere in the call chain. This will be familiar to Go and Rust developers, as Context is often used in the same way in those languages and is frequently employed when working with external API calls. It is conventional to name the context object `ctx`. 
+
+Read more on [Context in Go](https://www.makeuseof.com/go-contexts/)
+Read more on [Context in Rust](https://docs.rs/ctx/latest/ctx/)
```
cburgdorf, line 97: Not at all releated, sorry :wink:

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+        self.pending_returns[self.highest_bidder] += self.highest_bid
+    }
+    self.highest_bidder = ctx.msg_sender()
+    self.highest_bid = ctx.msg_value()
+
+    ctx.emit(HighestBidIncreased(bidder: ctx.msg_sender(), amount: ctx.msg_value()))
+}
+```
+
+The method first checks that the current block timestamp is not later than the contract's `aution_end_time` variable. If it *is* later, then the contract reverts. This is triggered using the []`revert`](../spec/statements/revert.md) keyword. The `revert` can accept a struct that becomes encoded as [revert data](https://github.com/ethereum/EIPs/issues/838). Here you can just revert without any arguments. Add the following definition somewhere in `Auction.fe` outside the main contract definition:
```
cburgdorf, line 126: Looks like the `revert` link not formed correctly

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+You have now implemented all the logic to handle a bid!
+
+### Withdrawing
+
+A previous high-bidder will want to retrieve their ETH from the contract so they can either walk away or bid again. You therefore need to create a `withdraw` method that the user can call. The function will lookup the user address in `pending_returns`. If there is a non-zero value associated with the user's address, the contract should send that amount back to the sender's address. It is important to first update the value in `pending_returns` and *then* send the ETH to the user, otherwise you are exposing a [re-entrancy](https://www.certik.com/resources/blog/3K7ZUAKpOr1GW75J2i0VHh-what-is-a-reentracy-attack) vulnerability (where a user can repeatedly call the contract and receive the ETH multiple times).
+
+Add the following to the contract to implement the `withdraw` method:
+
+```rust
+pub fn withdraw(mut self, mut ctx: Context) -> bool {
```
cburgdorf, line 164: Just because I haven't seen it mentioned here but the reason why we need to use `mut` with the Context here is because `send_value` is causing changes to the blockchain. If you were just to use something like `ctx.msg_sender()` then you would not need `mut` because we are just accessing information but not causing any changes to the blockchain. Looking at it this way, we have 9 different categories that functions can be categorized in from the perspective of their power (See "ABI conformity" here: https://github.com/ethereum/fe/issues/558)

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+```rust
+pub fn check_highest_bidder(mut self, ctx: Context) -> address {
+    return self.highest_bidder;
+}
+
+pub fn check_highest_bid(mut self, ctx: Context) -> u256 {
+    return self.highest_bid;
+}
+
+pub fn check_ended(mut self, ctx: Context) -> bool {
```
cburgdorf, line 228: This doesn't need to use `ctx`

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+To help test the contract without having to decode transaction logs, you can add some simple functions to the contract that simply report the current values for some key state variables (specifically, `highest_bidder`, `highest_bid` and `ended`). This will allow a user to use `eth_call` to query these values in the contract. `eth_call` is used for functions that do not update the state of the blockchain and costs no gas because the queries can be performed on local data.
+
+You can add the following functions to the contract:
+
+```rust
+pub fn check_highest_bidder(mut self, ctx: Context) -> address {
+    return self.highest_bidder;
+}
+
+pub fn check_highest_bid(mut self, ctx: Context) -> u256 {
```
cburgdorf, line 224: This doesn't need to use `ctx`

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+
+
+## View functions
+
+To help test the contract without having to decode transaction logs, you can add some simple functions to the contract that simply report the current values for some key state variables (specifically, `highest_bidder`, `highest_bid` and `ended`). This will allow a user to use `eth_call` to query these values in the contract. `eth_call` is used for functions that do not update the state of the blockchain and costs no gas because the queries can be performed on local data.
+
+You can add the following functions to the contract:
+
+```rust
+pub fn check_highest_bidder(mut self, ctx: Context) -> address {
```
cburgdorf, line 220: This doesn't need to use `ctx`

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,349 @@
...
+Ethereum addresses are already hex, so there is no further encoding required. The following command will take the constructor function and the hex-encoded arguments and concatenate them into a contiguous hex string, and save it to `constructor_args.txt`. 
+
+```sh
+cast abi-encode "__init__(uint256,address)" "0xa" "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720" > constructor_args.txt
+```
+
+Now append these constructor argument bytes to the bytecode generated by `fe build` (not that the `cut -c3` command removes the leading `0x` from the bytes being appended)
+
+```sh
+cat constructor_args.txt | cut -c3- >> output/Auction/Auction.bin
```
cburgdorf, line 268: I wonder how we can make this part more ergonomic. Vyper should have the same problem with foundry and they also have a foundry-vyper helper which our foundry-fe helper is based on. But I see that mostly used for testing. I'm not sure yet if/how one could use that for simple deployments. I will look into that.

cburgdorf: I *think* if you use `cast send` with the `--create` flag then you'll be able pass the raw bytecode and will also be able to append constructor args and have foundry do the heavy lifting.

cburgdorf:
Just wanted to add that I love to see this guide coming together! :heart: Just needs a bit of tuning but it's very nice!

cburgdorf:
Ah, one more thing: For substantial changes to the docs like this one you should also add a news fragment so that it gets mentioned in the next release notes (which are automated). Basically create a new file here: https://github.com/ethereum/fe/tree/master/newsfragments

It should be named `930.doc.md` with the content being roughly something like this:

```
Added a new tutorial: Open Auction
```

docs/src/user-guide/tutorials/auction.md
jmcook1186: ok, thanks, fixing in next commit

docs/src/user-guide/tutorials/auction.md
jmcook1186: ok, got it.

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+```sh
+auction_end_time: u256
+beneficiary: address
+highest_bidder: address
+highest_bid: u256
+pending_returns: Map<address, u256>
+ended: bool
+```
+
+Notice that variables are named using camel case (lower case, underscore separated, `like_this`).
```
cburgdorf, line 48: I think the correct term is *snake case* because camel case would be `likeThis`

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+Notice that variables are named using camel case (lower case, underscore separated, `like_this`). 
+[Addresses](../../spec/type_system/types/address.md) have their own type in Fe - it represents 20 hex-encoded bytes as per the Ethereum specification.
+
+The variables that expect numbers are given the `u256` type. This is an unsigned integer of length 256 bits. There are other [choices for integers](../../spec/type_system/types/numeric.md) too, with both signed and unsigned integers between 8 and 256 bits in length.
+
+The `ended` variable will be used to check whether the auction is live or not. If it has finished `ended` will be set to `true`. There are only two possible states for this, so it makes sense to declare it as a `bool` - i.e. true/false.
+
+The `pending returns` variable is a mapping between N keys and N values, with user addresses as the keys and their bids as values. For this, a `Map` type is used. In Fe, you define the types for the key and value in the Map definition - in this case, it is `Map<address, u256>`. Keys can be any `numeric` type, `address`, `boolean` or `unit`.
```
cburgdorf, line 55: Missing underscore `pending_returns`

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+
+Now you should decide which of these variables will have values that are known at the time the contract is instantiated. It makes sense to set the `beneficiary` right away, so you can add that to the constructor arguments.
+
+The other thing to consider here is *how* the contract will keep track of time. On its own, the contract has no concept of time. However, the contract does have access to the current block timestamp which is measured in seconds since the Unix epoch (Jan 1st 1970). This can be used to measure the time elapsed in a smart contract. In this contract, you can use this concept to set a deadline on the auction. By passing a length of time in seconds to the constructor, you can then add that value to the current block timestamp and create a deadline for bidding to end. Therefore, you should add a `bidding_time` argument to the constructor. Its type can be `u256`.
+
+When you have implemented all this, your contract should look like this:
+
+```rust
```
cburgdorf, line 63: Is there a reason that this can not use the `fe`  formatting and hence benefit from CI checks? Looks pretty self contained to me.

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+
+
+## View functions
+
+To help test the contract without having to decode transaction logs, you can add some simple functions to the contract that simply report the current values for some key state variables (specifically, `highest_bidder`, `highest_bid` and `ended`). This will allow a user to use `eth_call` to query these values in the contract. `eth_call` is used for functions that do not update the state of the blockchain and costs no gas because the queries can be performed on local data.
+
+You can add the following functions to the contract:
+
+```rust
+pub fn check_highest_bidder(mut self) -> address {
```
cburgdorf, line 222: I didn't spot this in my first review but these do not need to take `mut` as they are readonly

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+    return self.highest_bidder;
+}
+
+pub fn check_highest_bid(mut self) -> u256 {
```
cburgdorf, line 226: Same here, no `mut` needed. In the future, a linter could spot these and yell at the user.

docs/src/user-guide/tutorials/auction.md
```diff
@@ -0,0 +1,351 @@
...
+    return self.highest_bid;
+}
+
+pub fn check_ended(mut self) -> bool {
```
cburgdorf, line 230: Same here, no `mut` needed

docs/src/user-guide/tutorials/auction.md
cburgdorf: Ok, good to know. I could also take a look

cburgdorf (commented):
I like this a lot! I just spotted a few more nitpicks. It can go live when those are fixed.

docs/src/user-guide/tutorials/auction.md
cburgdorf: I didn't get it to work exactly the way I expected it, but this worked for me:

```
cast send --from <whatever-anvil-says> --create $(cat output/Auction/Auction.bin) $(cast abi-encode "__init_(uint256, address)", 0xa 0xa0Ee7A142d267C1f36714E4a8F75612F20a79720)

```

docs/src/user-guide/tutorials/auction.md
jmcook1186: doh - yes of course. Sorry!

docs/src/user-guide/tutorials/auction.md
jmcook1186: no - tbh i nuked them all in favour of ``rust because some were upsetting the linter, but now I see I can run `validate_doc_examples.py` locally to audit the snippets.

PR #933, opened 2023-09-22
Test logging

g-r-a-n-t:
Test logging. Events are decoded and logs that do not match an event signature or cannot be decoded are printed in their raw form.

### To-Do

- [x] Clean up commit history


cburgdorf (approved):
Cool! Looks good to me! :shipit:

PR #934, opened 2023-09-25
v2: modest parser improvements

sbillig:
- [x] Require comma between enum variants (and allow single-line enum def)
- [ ] Add SyntaxKind for function signature
- [ ] Add SyntaxKind for trait "signature" (the part of a trait definition before `{`)
- [ ] Add SyntaxKind for trait impl "signature"
... ?


sbillig:
@Y-Nak I'm inclined to merge these few simple changes as-is.

I'm currently experimenting with the `trait TryParse` idea and converting parse errors to enums, but will do that in a separate PR.

sbillig:
By "as-is" I don't mean that I won't make requested changes, of course, just that I'll do further parser restructuring separately.

Y-Nak:
I left two nits. I feel it'd be better to continue the discussion on #972.  Please feel free to merge the PR after the nits are resolved.

sbillig:
@Y-Nak I decided to apply one more change to this PR. Previously, `x >   > 1` and `x +  = 1` would be accepted as `>>` and `+=` operations; I think it's better to require that there be no whitespace in the operator, and I think my implementation is the least-bad option to solve this. (The other obvious option is to add `+=`, `>>`, etc as lexer tokens and allow splitting `>>` tokens into two)

PR #944, opened 2023-10-18
[website/docs]: add page on EVM precompiles

jmcook1186:
This PR adds a new page documenting the EVM precompiles included in Fe

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -14,6 +14,8 @@
 * [Development](development/index.md)
     * [Build & Test](development/build.md)
     * [Release](development/release.md)
+* [Standard Library](std/index.md)
+  * [Precompiles](std/precompiles.md)
 * [Specification (WIP)](spec/index.md)
     * [Notation](spec/notation.md)
     * [Lexical Structure](spec/lexical_structure/index.md)
--- a/docs/src/std/index.md
+++ b/docs/src/std/index.md
@@ -1,0 +1,5 @@
+# Fe Standard Library
+
+The standard library includes commonly used algorithms and data structures that come bundled as part of the language.
+
+- [Precompiles](./precompiles.md)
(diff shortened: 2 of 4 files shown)


docs/src/spec/precompiles.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+## `SHA2_256`
+
+`SHA2_256` is a hash function. a hash function generates a unique string of characters of fixed length from arbitrary input data.
+
+
+### Parameters
+
+- `input_buf`: a sequence of bytes to hash, `MemoryBuffer`
```
sbillig, line 63: The input argument at the callsite is `buf`; `input_buf` is the name of the value used internally in the function (argument labelling scheme was borrowed from swift, btw). The signature should probably be written `sha2_256(buf: MemoryBuffer) -> u256`.

(I'd argue that these single-argument functions should be defined as e.g. `sha2_256(_ input_buf: MemoryBuffer) -> u256`, so they can be called without a label on the arg.)

docs/src/spec/precompiles.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Returns
+
+`SHA2_256` returns a hash as a `u256`
+
+### Function signature
+
+```fe,ignore,ignore
+pub fn sha2_256(buf input_buf: MemoryBuffer) -> u256 {}
```
sbillig, line 72: Since I'm nitpicking, the `{}` should be left off of the signature.

sbillig (approved):
Looks great, thanks for adding this.

We could borrow the descriptions of the functions and parameters from evm.codes (MIT licensed, should be fair game, right?) for a bit more detail, but I'd be happy to see this merged any time as it's obviously much better than what we have now.

docs/src/SUMMARY.md
```diff
@@ -81,6 +81,7 @@
         * [Memory](spec/data_layout/memory/index.md)
             * [Sequence types in memory](spec/data_layout/memory/sequence_types_in_memory.md)
     * [Function calls](spec/function_calls.md)
+    * [Precompiles](spec/precompiles.md)
```
cburgdorf, line 84: Since precompiles are part of the std library, I believe this should sit outside of `spec` as spec is supposed to as close as possible a language specification (inspired by https://doc.rust-lang.org/stable/reference/introduction.html) and hence not cover std library stuff. I think that this should either become a sub of "Using Fe" or become a sub of a completely new item but I'm not sure what that would be (maybe "Standard Library"). The more I think about it, I think I like the idea of organizing this under a "Standard Library" section.

docs/src/spec/precompiles.md
sbillig: I mean "left off of" as in "not included". So, like this:
`pub fn sha2_256(buf input_buf: MemoryBuffer) -> u256`

PR #947, opened 2023-10-31
Include runtime_bytecode in artifacts

cburgdorf:
### What was wrong?

Currently, the compiler would only outputs the bytecode that is used for contract deployment but not the runtime bytecode that is later saved to storage and can be read out from the blockchain.

Being able to obtain the runtime bytecode is useful for contract verification.

### How was it fixed?

Basically just expose the `deployedBytecode` that `solc` already gives us.
Note:

- I named the file `mycontract.runtime.bin` because I find the term *runtime bytecode* less ambigious than *deployed bytecode*
- Producing runtime bytecode is another variant of the `--emit` options. Note that the default did **NOT** change from `abi, bytecode` because there are certain contracts where we currently can not produce runtime bytecode for. These seem to be contracts that use `create`/`create2` in `__init__`.
- Related to the former point, getting the runtime bytecode is on demand e.g. `--emit abi,bytecode,runtime-bytecode`

### DIFF
--- a/crates/driver/src/lib.rs
+++ b/crates/driver/src/lib.rs
@@ -28,6 +28,8 @@ pub struct CompiledContract {
     pub yul: String,
     #[cfg(feature = "solc-backend")]
     pub bytecode: String,
+    #[cfg(feature = "solc-backend")]
+    pub runtime_bytecode: String,
 }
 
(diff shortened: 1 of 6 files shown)


sbillig (approved):
👍

PR #949, opened 2023-11-06
Minor improvements for installation guide

cburgdorf:
### What was wrong?

Installation guide should focus on brew as it is the easiest way of installing Fe at this point. Also as people pointed out Fe works fine on Windows via WSL.

### How was it fixed?

:writing_hand:


jmcook1186 (approved):
lgtm!

PR #950, opened 2023-11-08
Correct install instructions for compiling from source

keyneom:
### What was wrong?
Currently a debug build is produced and the following command attempts to run a release version of the code that has never been built.

### How was it fixed?
Adds the `-r` flag to the cargo build command to ensure the following step that checks for a successful build runs correctly.

Also adds instructions to change directories into the Fe repository directory before attempting to build (following current instructions step-by-step would lead to failures).

### To-Do

- [ ] Clean up commit history


sbillig (approved):
👍  Thanks!

Issue #951, opened 2023-11-09
Failed to Verify Contract


Philogy:
### What is wrong?

Following the [simple dao blog post](https://blog.fe-lang.org/posts/simple-dao-written-in-fe-go-break-it/) I installed fe via brew (`fe -V` => `fe 0.26.0`) and tried to run the command to verify the contract:

```
fe verify 0xf0adbb9ed4135d1509ad039505bada942d18755f $ETH_RPC_URL
```

Whereby I ran this in the `fe_contracts` directory of the https://github.com/cburgdorf/simple_dao/ repo on tag `SimpleDAOBountyDeployment` (also doesn't work on master). My RPC seems perfect functional with me able to retrieve both contract code (tested via `cast code <addr>`) and the current block number. I was using the public `https://eth.llamarpc.com` endpoint for reference.

Full error from running the command:

```
Failed to verify contract.

error decoding response body: missing field `result` at line 1 column 143
```

cburgdorf:
Hi Philippe, thanks for jumping on this and taking the time to report the issue. The reponse of the node that you are using seems to not have a `result` field. I'll take a closer look but in the meantime, try out this public node instead: https://eth.public-rpc.com

cburgdorf:
Ok, I see what the problem is, the node seems to be picky about the request not specifying the `Content-Type` header. It should be easy to fix.

cburgdorf:
I created a fix that will be in the next release: https://github.com/ethereum/fe/pull/952
Thanks again for reporting, please use another JSON-RPC provider such as https://eth.public-rpc.com/ in the meantime.

PR #952, opened 2023-11-09
Supply Content-Type header to JSON-RPC call

cburgdorf:
### What was wrong?

Some JSON-RPC APIs are picky about the `Content-Type` header for requests. We should provide it. See #951 

### How was it fixed?

Provided


sbillig (approved):
👍

PR #954, opened 2023-11-11
Fix rpc-url in Quickstart::Deploy your contract

KumaCrypto:
In the example where the contract is deployed on the anvil the sepolia rpc-url is used. Corrected to anvil rpc-url.


sbillig (approved):
Thank you!

PR #884, opened 2023-05-09
Language server first pass

micahscopes:
This is an initial implementation of the Fe language server.  See [the readme](https://github.com/ethereum/fe/blob/96e95279646a64ac11822be0fc96c6622bd2b7d2/crates/language-server/README.md) for further details.

#### May 2023
- [x] Orientation, research and study of rust-analyzer, Fe compiler
- [x] LSP server crate scaffolding
- [x] LSP-based VSCode extension
- [x] Hover prototype
  - printing debug information
- [x] Basic development setup and documentation
#### June 2023
- [x] Diagnostics publishing prototype
  - using current deprecated diagnostics functionality
#### July 2023
- [x] Syntax highlighting in the VSCode extension via Textmate grammar file
- Initial work on transitioning to the new compiler architecture
  - [x] Salsa2022 database setup for language server
  - [x] Overhaul diagnostics types to support types in new compiler crates
  - [x] Working prototype of new diagnostics functionality based on @Y-Nak's `name-resolution` branch
#### August 2023
- [x] Improvements to LSP server + VSCode extension development setup
- [x] Improved `document/didChange` handling
- [x] Flexible LSP server global logging support using the `log` crate
- [x] Further work on transitioning to the new compiler overhaul
  - [x] Go-to definition prototype working for standalone .fe files; based on @Y-Nak's compiler rewrite
    - [x] Name resolution from cursor position
    - [x] Snap tests and unit tests for go-to definition functionality
  - [x] Error handling improvements
    - At the time, the rewritten name resolution only supported type resolution in the early resolution stage, but the data types and LSP server implementation should seamlessly support progress on the new compiler's name resolution functionality without many changes
#### September 2023
- [x] Go-to definition cleanup
- [x] Initial "workspace" functionality
  - a data structure and API for generating, caching and synchronizing necessary salsa inputs to support language server functionality
  - should support file synchronization via filesystem watching and LSP file change events alike
  - [x] Sketch of initial workspace cache data structures
    - Featuring a prefix tree based lookup for salsa inputs representing files, ingots and top-modules
    - Supporting local, external and standalone ingot caches.
  - [x] Upgraded go-to definition functionality to work with "local" ingots containing multiple modules
  - [x] Integration of workspace data structures with existing language server data structures; slight refactor of existing architecture
  - [x] Initial workspace sync prototype and tests
- [x] Initial support for subdiagnostics
- [x] Research and study of rust-analyzer's file synchronization functionality
- [x] Workspace sync API brainstorming and design

[...]

PR #979, opened 2024-01-29
Language server concurrency (tower-lsp rewrite)

micahscopes:
(please merge #884 before considering this PR)

This is a work in progress on a `tower-lsp` based rewrite of the language server.  In order to support concurrent execution of tasks, `tower-lsp` seemed like a solid choice.  It has a clean API that makes it much easier to implement LSP features.

So far the prototype has gone pretty smoothly, but there are some areas of uncertainty:

### Managing request handler execution order
`tower-lsp` doesn't really provide a way of managing the order of handler execution.  Neither does `lsp-server` for that matter.  What exactly should this execution dependency graph look like in various cases?  It's hard to foresee what this will look like as more and more LSP functionality gets implemented, but it's clear that we need a way to control the order of task execution somehow or another.

As a simple example, if I were to rename a file in vscode it could trigger multiple handlers (`did_open`, `did_close`, `watched_files_did_change`) concurrently.

How can we ensure that this pair of LSP events gets handled appropriately?  We're not just opening any file, we're opening a renamed file and we need to ensure that the workspace cache is updated to reflect this before executing diagnostics.  The `watched_files_did_change` handler ends up being redundant in this case if it gets executed after the `did_open` handler, but in the case where a file gets directly renamed outside of the LSP client, it's still important.

In this example, it's not a big deal to check for deleted (renamed) files in both handlers, since those checks are relatively short lived given how infrequently they'd be executed.  But it'll be important to ensure that e.g. diagnostics are run carefully, that the salsa inputs are set up correctly before running diagnostics or other complex tasks.  It's also important to ensure that expensive tasks aren't triggered redundantly in parallel.

### Shared state and deadlocks
Related to the issue of concurrency... How do we manage sharing of the salsa database and inputs cache in a concurrent environment?

In this prototype the salsa db and workspace cache are currently shared via `std::sync::Mutex` locks and the language server client is shared via `tokio::sync::Mutex` locks.  This works just fine but it requires care not to cause deadlocks.  The [tokio shared state docs](https://tokio.rs/tokio/tutorial/shared-state) were very helpful for me in understanding this better.

### Useful info
- [tokio shared state docs](https://tokio.rs/tokio/tutorial/shared-state)
- [A comprehensive high level comparison of `tower-lsp` and `lsp-server`](https://github.com/astral-sh/ruff/pull/7262)  
- [lamentations on concurrency issues in rust-analyzer](https://www.reddit.com/r/rust/comments/oo917a/comment/h5x2xkj/)
- [A discussion of concurrency issues with `tower-lsp`](https://github.com/ebkalderon/tower-lsp/issues/284)

### Conclusions so far
I like tower-lsp's clean API for dealing with LSP events.  It doesn't provide any guidance or tools for controlling the concurrency the order or flow of LSP event handlers.  Are there any simple ways of layering this on top of the existing tower-lsp implementation?

One possibility could be to use streams to manage combinations of LSP events, e.g. to debounce related, potentially interfering or redundant events.


micahscopes:
Notes from review with @sbillig and @Y-Nak:

- avoid mutable references when we aren't exclusively modifying salsa inputs
- salsa inputs should be modified explicitly in a single step that doesn't do other stuff, e.g.
  - don't modify salsa inputs in the hover handler, they should already be updated from the change handler
  - shouldn't need to store diagnostics on the language server db, diagnostics shouldn't need a mutable reference
  - generally we shouldn't see mutable references to language server db outside of document change handler
  - this will avoid excessive cache invalidation and free up multithreaded computation either using a rwlock or salsa's snapshot mechanism
- salsa's snapshot mechanism is similar to a rwlock but avoids deadlocks in case of cycles
- avoid using broadcast channels by default, only use them if broadcasting/forking is strictly necessary
  - maybe we can think in this order of preference: using mpsc channels should by default; then some kind of stream splitting mechanism (potentially using broadcast channels but guarding access to create extra receivers); then broadcast channels
  - this will allow just sending the raw oneshot channel instead of wrapping it
- try to avoid forking streams unnecessarily; be intentional about forking them 
- tokio multithreaded executor is currently unnecessary for the current stream handler configuration (one loop with a big `select!` statement containing all the stream handlers)
  - stream handler will be more i/o bound once we get the multithreaded salsa queries are sorted out
  - need to think more carefully about how to break stuff up
  - maybe we can have one select loop for mutating stream handlers and spawn separate select loops (on a separate executor?) for read only stream handlers, that way we can get all the read only stuff done in parallel and free up the db lock as soon as possible
- use `tracing` for logging, it will help guarantee that logs are correctly ordered

Y-Nak:
> salsa's snapshot mechanism is similar to a rwlock but avoids deadlocks in case of cycles

This is not true. What `Snapshot` does is almost the same as what `RwLock` does, so it's possible to introduce a deadlock either way. But it's rather difficult to cause a deadlock as long as we don't try to mutate the db from the salsa tracked function (e.g.,) by sending an event to the main thread to let the thread mutate the db.
E.g., 
```rust
#[salsa::tracked]
// Sender needs to implement `Clone` and `Hash` to be an argument of
// a salsa-tracked function, but I ignore the fact for simplicity.
fn rename(db: &dyn Db, rename_event: Event, tx: Sender ) {
      // Perform Reneming.
      // ...

      // Send an event to the main thread, and the main thread will try to mutate the database. 
      // This might cause a deadlock.
      tx.send(Event::SourceTextChanged) 
     
     // ...
}
```

Another possibility for the deadlock is not related to mutability thing, i.e., the deadlock situation might happen even if we only use `&Db` in muti-thread settings, but salsa detects this deadlock situation and raises a cycle error, which is, of course, nice.

Please refer to the below links for more information.
* https://salsa-rs.netlify.app/salsa2022/plumbing/cycles
* [Salsa Architecture Walkthrough(on youtube)](https://www.youtube.com/watch?v=vrnNvAAoQFk)
* [Salsa cycle-handling walkthrough(on youtube)](https://youtu.be/ES4H818jPus?si=IpwukW53as1yzSNG)
* [`jars_mut` implementation in salsa related to snapshot locking mechanism](https://github.com/salsa-rs/salsa/blob/f1d318a2795e87081e316a5f31ef02f0cd6a1c83/components/salsa-2022/src/storage.rs#L105-L125)

crates/language-server/src/functionality/handlers.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    pub(super) async fn handle_hover(
+        &mut self,
+        params: lsp_types::HoverParams,
+        responder: tokio::sync::oneshot::Sender<
+            Result<Option<lsp_types::Hover>, tower_lsp::jsonrpc::Error>,
+        >,
+    ) {
+        let db = self.db.snapshot();
```
sbillig, line 119: There's no reason to use a snapshot here, right? `hover_handler(&self.db, ...` should suffice (unless this is changed so that the task is spawned off to run separately)

crates/language-server/src/functionality/item_info.rs
```diff
@@ -0,0 +1,77 @@
...
+        ItemKind::Enum(enum_) => enum_.attributes(hir_db).data(hir_db).clone(),
+        ItemKind::TypeAlias(type_alias) => type_alias.attributes(hir_db).data(hir_db).clone(),
+        ItemKind::Trait(trait_) => trait_.attributes(hir_db).data(hir_db).clone(),
+        ItemKind::Impl(impl_) => impl_.attributes(hir_db).data(hir_db).clone(),
+        // ItemKind::Body(body) => body.attributes(hir_db).data(hir_db).clone(),
+        // ItemKind::Const(const_) => const_.attributes(hir_db).data(hir_db).clone(),
+        // ItemKind::Use(use_) => use_.attributes(hir_db).data(hir_db).clone(),
+        ItemKind::Contract(contract) => contract.attributes(hir_db).data(hir_db).clone(),
+        _ => vec![],
+    }
```
Y-Nak, line 21:
```suggestion
    let docstring = match item {
        ItemKind::Func(func) => func.attributes(hir_db).data(hir_db).,
        ItemKind::Mod(mod_) => mod_.attributes(hir_db).data(hir_db),
        ItemKind::Struct(struct_) => struct_.attributes(hir_db).data(hir_db),
        ItemKind::Enum(enum_) => enum_.attributes(hir_db).data(hir_db),
        ItemKind::TypeAlias(type_alias) => type_alias.attributes(hir_db).data(hir_db),
        ItemKind::Trait(trait_) => trait_.attributes(hir_db).data(hir_db),
        ItemKind::Impl(impl_) => impl_.attributes(hir_db).data(hir_db),
        // ItemKind::Body(body) => body.attributes(hir_db).data(hir_db),
        // ItemKind::Const(const_) => const_.attributes(hir_db).data(hir_db),
        // ItemKind::Use(use_) => use_.attributes(hir_db).data(hir_db),
        ItemKind::Contract(contract) => contract.attributes(hir_db).data(hir_db),
        _ => return None,
    }
```

No need to clone attributes.

crates/language-server/src/server.rs
```diff
@@ -1,157 +1,97 @@
...
+    }
+
+    pub(crate) fn new(client: Client, messaging: MessageSenders) -> Self {
+        Self { messaging, client }
     }
 }
 
-pub fn run_server() -> Result<()> {
-    let (connection, io_threads) = Connection::stdio();
+#[language_server_macros::message_channels]
```
Y-Nak, line 38: I don't quite understand why we need as many channels as LSP request kinds.
It feels that this macro is an abuse of proc macro, decreases readability and maintainability, and also the generated structs already contain unused channels.

Is there any specific reason that a message enum doesn't work in this case?

crates/language-server/src/main.rs
```diff
@@ -1,19 +1,48 @@
...
-    let _ = run_server();
+    tokio::select! {
+        // setup logging
+        _ = handle_log_messages(rx, server.client.clone()) => {},
+        // start the server
+        _ = tower_lsp::Server::new(stdin, stdout, socket)
+            .serve(service) => {}
+        // backend
+        _ = functionality::streams::setup_streams(&mut backend, message_receivers) => {}
+    }
```
Y-Nak, line 47: I don't understand why you need `tokio::select` here.

Is there any reason that you can't do the one below?
```rust
// Prepare channels for `Server` <-> `Backend` communication.
let (tx, rx) = make_channels(...);

let backend = Backend::new(rx, ...);
// Run backend in a child thread.
tokio::spawn(backend.run());

// Make `Server`, `Backend` and the thread for logging is managed by 
// this server via channels for e.g., graceful shutdown.
let server = ...

// Run server.
server.serve(service).await;
```

crates/language-server/src/functionality/item_info.rs
```diff
@@ -0,0 +1,77 @@
+use hir::{
+    hir_def::{Attr, ItemKind},
+    span::LazySpan,
+    HirDb, SpannedHirDb,
+};
+
+pub fn get_item_docstring(item: &ItemKind, hir_db: &dyn HirDb) -> Option<String> {
```
Y-Nak, line 7:
```suggestion
pub fn get_item_docstring(item: ItemKind, hir_db: &dyn HirDb) -> Option<String> {
```

ItemKind implements `Copy` and its size is 8 bytes. No need to pass a reference.

crates/language-server/src/functionality/item_info.rs
```diff
@@ -0,0 +1,77 @@
...
+    .collect::<Vec<_>>();
+
+    if docstring.is_empty() {
+        None
+    } else {
+        Some(docstring.join("\n"))
+    }
+}
+
+pub fn get_item_path_markdown(item: &ItemKind, hir_db: &dyn HirDb) -> Option<String> {
```
Y-Nak, line 39: ditto

crates/language-server/src/functionality/item_info.rs
```diff
@@ -0,0 +1,77 @@
...
+    item.scope()
+        .pretty_path(hir_db)
+        .map(|path| format!("```fe\n{}\n```", path))
+}
+
+pub fn get_item_definition_markdown(item: &ItemKind, db: &dyn SpannedHirDb) -> Option<String> {
```
Y-Nak, line 45: ditto

Y-Nak (commented):
I looked through the PR. I'll probably add more comments later.

crates/language-server/src/server.rs
Y-Nak: Oh, I missed the part;
> Proof of concept stream-based LSP event handling on the backend side
Intended for dealing with LSP event handler execution order, cancellation, and debouncing of expensive event handlers
 Simple use case: aggregating document updates and handling them with a single on_change handler
See https://github.com/ebkalderon/tower-lsp/issues/284 for examples of more complex scenarios that could arise

But the macro does not seem so good...

Y-Nak: Still, I don't think we need one-to-one correspondence between request kinds.

> 1. I would like to avoid `async` in our implementation. LSPs are mostly
computationally bound rather than I/O bound, and `async` adds a lot of
complexity to the API, while also making harder to reason about
execution order. This leads into the second reason, which is...
> 2. Any handlers that mutate state should be blocking and run in the
event loop, and the state should be lock-free. This is the approach that
`rust-analyzer` uses (also with the `lsp-server`/`lsp-types` crates as a
framework), and it gives us assurances about data mutation and execution
order. `tower-lsp` doesn't support this, which has caused some
[issues](https://github.com/ebkalderon/tower-lsp/issues/284) around data
races and out-of-order handler execution.
> 3. In general, I think it makes sense to have tight control over
scheduling and the specifics of our implementation, in exchange for a
slightly higher up-front cost of writing it ourselves. We'll be able to
fine-tune it to our needs and support future LSP features without
depending on an upstream maintainer.

This is the reasoning about why ruff chose `lsp-server` instead of `tower-lsp`.
It seems reasonable to me. 
https://github.com/astral-sh/ruff/pull/10158

Y-Nak (commented):
I've added a review for the implementation of goto. 

---
An additional note for future reference:

For optimization purposes, it would be beneficial to narrow the possible range before collecting actual paths when we know the passed cursor is included in `ItemKind::Body`. 

This is because:

* The Body contains numerous paths (even local variables are defined as paths in HIR).
* `Hir` maintains `Expr <=> Span` and `Stmt <=> Span` mapping

Therefore, it would be more efficient to narrow down the possible range before collecting paths. Implementing `find_enclosing_stmt`  and `find_enclosing_expr` would be required for this((I think these are necessary in the near future either way).

This narrowing-down-range feature should probably be generalized, e.g.,
```rust
/// Find the closest HIR element that includes the cursor, by traversing the HIR element hierarchy.
fn find_closest_hir_elem<T: HirElem>(db: &dyn LanguageServerDb, cursor: Cursor) -> Option<T> {
    ...
}
```

crates/language-server/src/functionality/streams.rs
```diff
@@ -0,0 +1,107 @@
...
+            Some((_, responder)) = shutdown_stream.next() => backend.handle_shutdown(responder).await,
+            Some(params) = did_delete_watch_file_stream.next() => backend.handle_deleted(params, tx_needs_diagnostics.clone()).await,
+            Some(params) = change_stream.next() => backend.handle_change(params, tx_needs_diagnostics.clone()).await,
+            Some(files_need_diagnostics) = diagnostics_stream.next() => backend.handle_diagnostics(files_need_diagnostics).await,
+            Some((params, responder)) = hover_stream.next() => backend.handle_hover(params, responder).await,
+            Some((params, responder)) = goto_definition_stream.next() => backend.handle_goto_definition(params, responder).await,
+        }
+        tokio::task::yield_now().await;
+    }
+}
```
sbillig, line 107: I think the `fork()` business adds unnecessary complexity.

```suggestion
pub struct FileChange {
    pub uri: url::Url,
    pub kind: ChangeKind,
}
pub enum ChangeKind {
    Open(String),
    Create,
    Edit(Option<String>),
    Delete,
}

pub async fn setup_streams(backend: &mut Backend, mut receivers: MessageReceivers) {
    info!("setting up streams");
    let mut initialized_stream = receivers.initialize_stream.fuse();
    let mut shutdown_stream = receivers.shutdown_stream.fuse();

    let mut change_stream = (
        receivers
            .did_change_watched_files_stream
            .map(|params| futures::stream::iter(params.changes))
            .flatten()
            .fuse()
            .map(|event| {
                let kind = match event.typ {
                    FileChangeType::CHANGED => ChangeKind::Edit(None),
                    FileChangeType::CREATED => ChangeKind::Create,
                    FileChangeType::DELETED => ChangeKind::Delete,
                    _ => unreachable!(),
                };
                FileChange {
                    uri: event.uri,
                    kind,
                }
            }),
        receivers.did_open_stream.fuse().map(|params| FileChange {
            uri: params.text_document.uri,
            kind: ChangeKind::Open(params.text_document.text),
        }),
        receivers.did_change_stream.fuse().map(|params| FileChange {
            uri: params.text_document.uri,
            kind: ChangeKind::Edit(Some(params.content_changes[0].text.clone())),
        }),
    )
        .merge()
        .fuse();

    let (tx_needs_diagnostics, rx_needs_diagnostics) =
        tokio::sync::mpsc::unbounded_channel::<String>();

    let mut diagnostics_stream = UnboundedReceiverStream::from(rx_needs_diagnostics)
        .chunks_timeout(500, std::time::Duration::from_millis(30))
        .fuse();

    let mut hover_stream = (&mut receivers.hover_stream).fuse();
    let mut goto_definition_stream = (&mut receivers.goto_definition_stream).fuse();

    info!("streams set up, looping on them now");
    loop {
        tokio::select! {
            Some((params, responder)) = initialized_stream.next() => backend.handle_initialized(params, responder).await,
            Some((_, responder)) = shutdown_stream.next() => backend.handle_shutdown(responder).await,
            Some(change) = change_stream.next() => backend.handle_change(change, tx_needs_diagnostics.clone()).await,
            Some(files_need_diagnostics) = diagnostics_stream.next() => backend.handle_diagnostics(files_need_diagnostics).await,
            Some((params, responder)) = hover_stream.next() => backend.handle_hover(params, responder).await,
            Some((params, responder)) = goto_definition_stream.next() => backend.handle_goto_definition(params, responder).await,
        }
        tokio::task::yield_now().await;
    }
}
```

crates/language-server/src/functionality/handlers.rs
```diff
@@ -0,0 +1,137 @@
...
+        let contents = Some(doc.text);
+        if let Some(contents) = contents {
+            let input = self
+                .workspace
+                .touch_input_for_file_path(&mut self.db, path)
+                .unwrap();
+            let _ = input.sync_from_text(&mut self.db, contents);
+        }
+        let _ = tx_needs_diagnostics.send(path.to_string());
+    }
```
sbillig, line 84: (Associated change for the removed of the stream forking)

```suggestion
    pub(super) async fn handle_change(
        &mut self,
        change: FileChange,
        tx_needs_diagnostics: tokio::sync::mpsc::UnboundedSender<String>,
    ) {
        let path = change.uri.to_string();

        match change.kind {
            ChangeKind::Open(contents) => {
                info!("file opened: {:?}", &path);
                self.update_input_file_text(&path, contents);
            }
            ChangeKind::Create => {
                info!("file created: {:?}", &path);
                let contents = tokio::fs::read_to_string(&path).await.unwrap();
                self.update_input_file_text(&path, contents)
            }
            ChangeKind::Edit(contents) => {
                info!("file edited: {:?}", &path);
                let contents = if let Some(text) = contents {
                    text
                } else {
                    tokio::fs::read_to_string(&path).await.unwrap()
                };
                self.update_input_file_text(&path, contents);
            }
            ChangeKind::Delete => {
                info!("file deleted: {:?}", path);
                self.workspace
                    .remove_input_for_file_path(&mut self.db, &path)
                    .unwrap();
            }
        }
        tx_needs_diagnostics.send(path).unwrap();
    }
    
    fn update_input_file_text(&mut self, path: &str, contents: String) {
        let input = self
            .workspace
            .touch_input_for_file_path(&mut self.db, path)
            .unwrap();
        let _ = input.sync_from_text(&mut self.db, contents);
    }

```

sbillig: I used `tokio::fs::read_to_string` here, but that might introduce the possibility of changes being handled out of order.

I share Yoshi's skepticism of using async because of these tricky situations, but I remain open minded. If we did drop async, I think the "streamy" essence of the code could remain, we'd just eg explicitly spawn a thread (or "actor") that sequentially handles all change events that arrive on a channel.

crates/language-server/src/backend/workspace.rs
```diff
@@ -303,52 +395,82 @@ impl Workspace {
...
         } else {
             self.standalone_ingot_context
-                .top_mod_from_file_path(db, path)
+                .remove_input_for_file_path(db, path)?;
+            Ok(())
         }
     }
 }
 
 pub trait SyncableInputFile {
```
sbillig, line 470: This trait seems unnecessary.

sbillig: I meant the SyncableInputFile trait.

crates/hir-analysis/src/ty/def_analysis.rs
```diff
@@ -889,7 +889,7 @@ enum DefKind {
     Adt(AdtDef),
     Trait(TraitDef),
     ImplTrait(Implementor),
-    Impl(HirImpl, TyId),
+    Impl(HirImpl, #[allow(dead_code)] TyId),
```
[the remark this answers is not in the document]
micahscopes: When I upgraded rust with `rustup` clippy complained about `TyId` being unused.  I added this temporarily but wasn't sure how to proceed.

crates/hir/src/diagnostics.rs
```diff
@@ -24,6 +24,7 @@ pub trait DiagnosticVoucher: Send {
     fn error_code(&self) -> GlobalErrorCode;
     /// Makes a [`CompleteDiagnostic`].
     fn to_complete(&self, db: &dyn SpannedHirDb) -> CompleteDiagnostic;
+    fn clone_box(&self) -> Box<dyn DiagnosticVoucher>;
```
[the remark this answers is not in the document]
micahscopes: I'm not sure if it's the correct way of doing it, but this allowed me to to implement snapshotting (the `ParallelDatabase` trait) without the compiler complaining about not being able to clone stuff.

crates/language-server/src/backend/db.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl LanguageServerDatabase {
+    pub fn as_language_server_db(&self) -> &dyn LanguageServerDb {
+        <Self as salsa::DbWithJar<Jar>>::as_jar_db::<'_>(self)
+    }
+}
+
+impl salsa::Database for LanguageServerDatabase {
+    fn salsa_event(&self, _: salsa::Event) {}
```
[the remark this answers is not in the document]
micahscopes: The logger has been showing some salsa events already even without implementing this, I guess from the other crates?  I actually disabled them temporarily [here](https://github.com/micahscopes/fe/blob/04b3fc2fa0c4019eb6d99cd5464597821c0d5e27/crates/language-server/src/logger.rs#L55-L59) in order to focus on messages from the language-server crate.

It could be cool to have separate outputs for these or some other way of toggling.

crates/language-server/src/server.rs
micahscopes: :sweat_smile:  I see how the macro isn't exactly a simple solution, and the motivation for it was also messy (getting around those tower-lsp `LanguageServer` trait limitations).  My intention was to reduce the surface area of boilerplate when creating these channels to bridge to the `Backend`, but the tradeoff is opacity and complexity.  Plus, even with this macro there's still a lot of boilerplate in setting up all the stream handlers.

As fun as it was to write I agree that it'd be good not to have it if possible.

As you mentioned on discord one possibility could be to use the [support-mutable-methods branch](https://github.com/ebkalderon/tower-lsp/tree/support-mutable-methods), at least until they figure out a more permanent solution.  I like that suggestion a lot, it'd seriously simplify things.  That way there'd be no requirement to make a channel/stream just to handle an LSP event directly.  We could still get the flexibility and control of the async `Stream` ecosystem as needed and the lock-free state but without needing to set up all those channels/streams by default.  I also appreciate that the `tower-lsp` maintainers seem to be taking this issue seriously.  Another possibility is this [async-lsp](https://github.com/ebkalderon/tower-lsp/issues/284#issuecomment-1709154394) crate which seems quite solid, even though the API is less readable.

As for the decision of `lsp-server` vs `tower-lsp`:  I really like `tower-lsp`'s high level API and the built-in task management inherent to rust async.  Using `lsp-server` would require us to build some kind of task manager regardless, which to me seems like an inherently complex thing to reason about and maintain.

The thing that keeps bothering me is how LSP is an async protocol by design ([1](https://github.com/microsoft/language-server-protocol/issues/306#issuecomment-333776338), [2](https://github.com/microsoft/language-server-protocol/issues/1693)) in nuanced ways that I'm not convinced can be adequately dealt just by faithfully processing events in the order they're received...  LSP events don't always have a well defined total order.  Processing some events in order is essential (e.g. `didChange` events), but there are situations when a single user action can trigger multiple loosely coupled LSP events more or less concurrently.  For example, renaming a file can trigger:
- `willRename` (a request that the client waits on a response for before renaming)

Followed by:
- `didRename` (after the client renames)
- `didChangeWatchedFiles` (from the filesystem watcher with two changes, a create and delete, seems to be delayed by ~1 second on my computer) 
- `didOpen` (when the file with the new name is opened)
- `didClose` (when the file of the previous name is closed)

Great care needs to be taken to ensure that each of these things is doing exactly what they need to do without overlap or conflict, especially in cases where any of these mutates state.  But I'm anticipating edge cases that will be tricky to reason about imperatively, where we may want to condition event handling on the presence or absence of several other recent or pending events, perhaps over a given debounce period.  Ideally we won't need to do this, but it [doesn't seem unlikely](https://github.com/rust-lang/rust-analyzer/issues/119), especially when we get into code actions or refactoring features.

The idea of trying to manage these complexities imperatively/synchronously with a custom task manager makes me nervous, regardless of how faithfully we adhere to the order these events arrive in.

In spite of the current rigidity of `tower-lsp` and the potential complexity of choosing rust async, I do think the async ecosystem (especially streams) could bring a lot of flexibility in managing execution order and in making sense of concurrent events, if we use it carefully.

---

A few more notes about it before letting that proc macro go (RIP 🪦):
- It only makes channels for trait methods that are implemented, so only for the those that are used
- I thought about using an enum or doing some kind of "multiplexing" over a single channel but my the amount of boilerplate code it would take to define the enums and join the data and split the channel separate streams seemed comparable to the amount it would take to just have a single channel per LSP event type.  Which either way felt like too much.
- The macro was inspired by the one used to generate the routes in `tower-lsp` ([here](https://github.com/ebkalderon/tower-lsp/blob/master/tower-lsp-macros/src/lib.rs)) but I admit that it got quite a bit more complex, a work of art

crates/language-server/src/functionality/item_info.rs
micahscopes: Nice, thank you

crates/language-server/src/backend/workspace.rs
micahscopes: I think it's vestigial!

micahscopes: Good eye

crates/language-server/src/functionality/handlers.rs
micahscopes: I haven't tried it yet but this looks great.  Can you help me understand how these changes might get handled out of order?  My thought is that by merging all those change events into a single stream we will enforce that they're processed serially in the order they were merged, at least relative to one another.

micahscopes: (By the way, I shared my complex thoughts on the async stuff here: https://github.com/ethereum/fe/pull/979#discussion_r1545977502)

crates/language-server/src/functionality/handlers.rs
micahscopes: Yup, this is also vestigial, I'd been handling hover in a worker but took it out and forgot to remove this.

crates/language-server/src/functionality/goto.rs
```diff
@@ -0,0 +1,447 @@
...
+        let scope = ctxt.scope();
+        self.path_map.insert(span, (path, scope));
+        walk_path(self, ctxt, path);
+    }
+
+    fn visit_ident(
+        &mut self,
+        ctxt: &mut VisitorCtxt<'_, hir::visitor::prelude::LazySpanAtom>,
+        ident: hir::hir_def::IdentId,
+    ) {
```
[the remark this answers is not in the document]
micahscopes: Ah yes, I see. That will be more straight to the point.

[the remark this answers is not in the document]
micahscopes: How can I get the exact span for a segment when iterating manually?

crates/language-server/src/functionality/goto.rs
```diff
@@ -0,0 +1,447 @@
...
+    let mut segments = cursor_path.segments(db.as_hir_db()).clone();
+    let is_partial = if let Some(pos) = segments.iter().position(|ident| match ident {
+        Partial::Present(ident) => *ident == cursor_ident,
+        Partial::Absent => false,
+    }) {
+        segments.truncate(pos + 1);
+        segments.len() < cursor_path.segments(db.as_hir_db()).len()
+    } else {
+        false
+    };
```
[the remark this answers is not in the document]
micahscopes: This is awesome, I just implemented it and it works great.  I especially like the part about not walking the whole top module!

micahscopes:
Closing in favor of #1022

PR #980, opened 2024-02-01
Dependency resolution via Git

cburgdorf:
### What was wrong?

We currently have no way to specify dependencies that do not already sit on the local file system (e.g. need to be resolved through Git)

### How was it fixed?

This PR implements MVP like support for dependencies to be resolved through Git.

Dependencies can now have an additional `source`  and `rev` field to be used like this:

```
[dependencies]
fpmath = {  version = "1.0", source="https://github.com/bilgin-kocak/felang-fixedpoint", rev="0eb57d267f7f174419ee257f97d7e02eb3469712" }
```

Note that:

1. The Git repository is cloned *shallow*  which means that even for a very large repository we are only fetching the bare minimum that is required to get the files from the revision we want.

2. The dependency is only downloaded once and then treated as a local dependency from then on.

3.  Dependencies are downloaded to whatever to `FE_HOME/.deps` where `FE_HOME` is either set via environment variable or else defaults to `.fe` in the user directory. I believe that is how cargo handles it as well.

4. A dependency can not be specified without `rev` which is by design and especially important since we do not yet support having a lock file to pin down exact versions. So in order to reduce the danger of supply chain attacks a git dependency always needs to be specified together with a `rev`.


g-r-a-n-t (commented):
LGTM

g-r-a-n-t (approved):

Issue #978, opened 2024-01-29
Add support for transient storage


g-r-a-n-t:
### What is wrong?

There is currently no way to use `tstore` or `tload`  in Fe.

see https://docs.soliditylang.org/en/latest/yul.html#evm-dialect

### How can it be fixed

- add `__tstore` and `__tload` [intrinsics](https://github.com/ethereum/fe/blob/d1f633273952b2edec3a31d1a584a9d0992d34bc/crates/analyzer/src/builtins.rs#L40)
- add `evm::tstore` and `evm::tload` to the [std lib](https://github.com/ethereum/fe/blob/d1f633273952b2edec3a31d1a584a9d0992d34bc/crates/library/std/src/evm.fe)

MukulKolpe:
Hey @g-r-a-n-t, can I work on this issue?

Hrushi20:
Hey, I'm working on this issue. Was busy with personal works.

g-r-a-n-t:
hi @MukulKolpe, are you interested in working on transient storage specifically?

MukulKolpe:
Hi @g-r-a-n-t, I'm open to working on other areas too. Is there anything else I can start with?

elielnfinic:
I am open to work on this task if needed.
Thank you.

g-r-a-n-t:
the implementation provided here https://github.com/ethereum/fe/pull/982 is good, but the implementer ran into some issues with the build after updating the solc bindings. this is usually the case when updating such a dependency and it's not fun to fix.

elielnfinic:
I agree, would you like it to be built without bindings or shall I try to fix the issue?

PR #982, opened 2024-02-04
transientStorage functions

Hrushi20:
Pull Request adds TransientStorage functions to fe based on issue #978


g-r-a-n-t:
@Hrushi20 you can test locally using `cargo test` or `cargo test --features solc-backend`

I forgot to mention that solc-rust will need to be updated from `0.8.22` to `0.8.24`. this is usually a pretty straight forward task

crates/analyzer/src/builtins.rs
```diff
@@ -73,6 +73,8 @@ pub enum Intrinsic {
     __sload,          // (p)
     __sstore,         // (p, v) -> ()
     __msize,          // ()
+    __tstore,         // ()
```
g-r-a-n-t, line 76: the parameter and return count should be added to the functions defined in this module

Hrushi20:
Sorry for the late fix. Got caught up with other works.

Hrushi20:
Hey, how do I fix the broken tests?

PR #984, opened 2024-02-08
Dummy file path conversions to fix CI + WASM build

micahscopes:
### What was wrong?
The WASM build was failing due to `Uri::to_file_path` and `Uri::from_file_path` not being implemented for WASM targets.

### How was it fixed?
I made a trait defined only for WASM targets which implements dummy functions


sbillig (approved):
👍

PR #986, opened 2024-02-20
ingot resolution

g-r-a-n-t:


Y-Nak:
I think it'd be better to have a discussion on this PR to align with.
The things that I'd like to talk about are
- Definition of the traits in `ops::cmp` and `ops::logical`, and `Plus` trait.
- `abort` definition and its usage 
- std module structure
- driver's setting

g-r-a-n-t:
added in #1068

PR #990, opened 2024-02-28
parser: Allow `mut` occurrence only in path pattern

Y-Nak:
This PR restricts 
1. the `mut` keyword occurrence in a pattern only before `Path`,
2. Add a field to `Hir::Pat::Path` to indicate the mutability of the path.

The main reason for the change was that I hesitated to make all HIR pattern kinds include mutability flags.


sbillig (approved):
👍

PR #991, opened 2024-02-29
v2 `unsafe` block parsing

g-r-a-n-t:
Added `unsafe` block parsing.

```
unsafe {
  // unsafe code
}
```


Y-Nak (commented):
`unsafe` should be applied to `block` expression.  Please consider this example.
```rust
let x = unsafe {
    unsafe_fn()
}
```
So basically, what you need to do is
1. Modify `expr` parsing
2. Add `unsafe_kw` method to `ast::BlockExpr`
3. Modify `hir::Expr::Block(Vec<StmtId>)` to `hir::Expr::Block(Vec<StmtId>, bool)` or `hir::Expr::Block{ stmts: Vec<StmtId>, is_unsafe: bool }`
4. Modify `hir::lower::expr::Expr::lower_ast`
5. Define `LazyBlockExprSpan` and add a method to be able to get `unsafe` span

crates/parser2/src/syntax_kind.rs
```diff
@@ -284,6 +284,8 @@ pub enum SyntaxKind {
     IfExpr,
     /// `match x { pat => { .. } }`
     MatchExpr,
+    /// `unsafe { foo() }`
+    UnsafeExpr,
```
g-r-a-n-t, line 288: remove

crates/parser2/src/parser/item.rs
```diff
@@ -71,6 +71,13 @@ impl super::Parse for ItemScope {
     fn parse<S: TokenStream>(&mut self, parser: &mut Parser<S>) {
         use crate::SyntaxKind::*;
 
+        // if parser.peek_two() == (Some(UnsafeKw), Some(LBrace)) {
```
g-r-a-n-t, line 74: remove

crates/parser2/src/parser/expr_atom.rs
```diff
@@ -83,8 +84,21 @@ impl super::Parse for BlockExprScope {
                 .map(SyntaxKind::is_item_head)
                 .unwrap_or_default()
             {
-                parser.parse(ItemScope::default())?;
-                continue;
+                let is_item = parser.dry_run(|parser| {
+                    if parser.bump_if(SyntaxKind::UnsafeKw) {
+                        parser
+                            .current_kind()
+                            .map(SyntaxKind::is_item_head)
```
g-r-a-n-t, line 91: `unsafe unsafe`

crates/parser2/src/parser/expr_atom.rs
```diff
@@ -83,8 +84,22 @@ impl super::Parse for BlockExprScope {
...
+                            && !parser.bump_if(SyntaxKind::UnsafeKw)
+                    } else {
+                        true
+                    }
+                });
+
+                if is_item {
+                    parser.parse(ItemScope::default())?;
+                    continue;
+                }
```
sbillig, line 102: `is_item_head` should be changed to not include `UnsafeKw`, because that's no longer a sure indicator of item-ness.

`dry_run` has some overhead; it would be better to only do a `dry_run` when the first token is known to be `unsafe`.

g-r-a-n-t: I've addressed the second point, but left `is_item_head` and `is_expr_atom_head` as they are.

g-r-a-n-t:
I will add unsafe expr checking to this as well.

g-r-a-n-t:
> It's still unclear to me what the `unsafe` actually means in Fe. Do you have any concrete thoughts on this?

In general, code should be labeled as `unsafe` if it has the potential to introduce undefined behavior during runtime. For instance, an invocation of `evm::mstore(..)` could alter the value of data that is exclusively used in safe code, thus necessitating the `unsafe` label.

Y-Nak:
Ok. So, in the v1 implementation, there are functions that don't cause undefined behavior (UB) but are still marked as unsafe, like `RawCallBuffer::offset`, which merely returns an offset. Am I correct in understanding that these should not be marked as unsafe? I felt that unsafe was being somewhat abused in v1, so I want to eliminate any ambiguity in my understanding.

g-r-a-n-t:
That is correct. I agree that `unsafe` was overused in v1 and that it should be used more sparingly in the v2 implementation.

Y-Nak:
Ok, thanks!

PR #997, opened 2024-04-23
Fix a bug in constraitns collection when `Self` appears in where clause

Y-Nak:


sbillig (approved):
👍 (A more diligent reviewer than I would have caught this)

PR #998, opened 2024-04-24
Fix incorrect instantiation in constraint solving

Y-Nak:
I broke the constraint solver when I introduced `Binder`.
This PR also adds the `Callable` information in the type checker, which will be necessary for MIR-lowering and the WF check for types that are implicitly introduced in method resolution.


sbillig (approved):
LGTM

PR #1000, opened 2024-05-02
Add support for development in both Docker and Github Codespace.

afsec:
### What was wrong?

There is no support for Devcontainer

### How was it fixed?

Added `.devcontainer/devcontainer.json` file.

### To-Do

- [X] Clean up commit history

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+{
+	"name": "fe-lang-v2",
+	"image": "mcr.microsoft.com/devcontainers/rust:latest",
+	"features": {
+		"ghcr.io/devcontainers/features/node:1": {},
+		"ghcr.io/devcontainers/features/github-cli:1": {},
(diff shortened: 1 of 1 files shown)


sbillig (approved):
@afsec Thanks! I haven't used gh codespaces (until testing this), but it seems useful to have this defined. Maybe I'll start reviewing PRs in codespaces and further tie myself to gh 🕺. I have a slight feeling that we should keep the set of installed extensions and customizations as minimal as possible, but I just tried this and decided that it seems fine for now. Someone can modify and submit another PR if they feel strongly about it.

Someday when the compiler is ready to be used by people for real things, maybe we'll want to be more careful about such things. Minimizing the attack surface is prudent, even though it's unlikely that someone compromises a popular vscode extension in order to inject some code into the fe compiler to change people's contracts to steal funds or whatever.

(Ignoring the CI failures; they're all unrelated and have been fixed elsewhere)

PR #1001, opened 2024-05-02
Fix trait method selection

Y-Nak:
This PR fixes the bug that @g-r-a-n-t  reported today and makes clippy(v1.78.0) happy.
(I want to clean up the ty-check test; the output is really messy...). But for now, I want to keep it as is.


Y-Nak:
Hmm, CI fails. It seems to be a v1 issue rather than a v2 issue. I'll look into this tomorrow.

sbillig:
> Hmm, CI fails. It seems to be a v1 issue rather than a v2 issue. I'll look into this tomorrow.

Fixed by @g-r-a-n-t here: #1003

sbillig (approved):
👍

PR #1003, opened 2024-05-03
Update revm

g-r-a-n-t:
### What was wrong?

The test runner crate was using version 3 of revm, which caused CI issues.

### How was it fixed?

Switched to the latest version of revm and made the test-runner compatible.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/crates/analyzer/tests/errors.rs
+++ b/crates/analyzer/tests/errors.rs
@@ -340,10 +340,10 @@ test_ingot! { mainless_ingot }
 test_ingot! { bad_visibility }
 test_ingot! { missing_dep }
 test_ingot! { misconfigured_dep }
-#[cfg(not(target_arch = "wasm32"))]
-test_ingot! { dep_unavailable_rev }
-#[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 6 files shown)


sbillig (approved):
👍

crates/analyzer/tests/errors.rs
```diff
@@ -340,10 +340,10 @@ test_ingot! { mainless_ingot }
...
 test_ingot! { missing_dep }
 test_ingot! { misconfigured_dep }
-#[cfg(not(target_arch = "wasm32"))]
-test_ingot! { dep_unavailable_rev }
-#[cfg(not(target_arch = "wasm32"))]
-test_ingot! { dep_unavailable_source }
+// #[cfg(not(target_arch = "wasm32"))]
+// test_ingot! { dep_unavailable_rev }
+// #[cfg(not(target_arch = "wasm32"))]
+// test_ingot! { dep_unavailable_source }
```
g-r-a-n-t, line 346: it looks like the git2 crate we're using for dep resolution isn't returning errors for non-existent repositories

PR #1004, opened 2024-05-04
Update revm on fe-v2 branch

sbillig:
This is #1003 cherry-picked to the fe-v2 branch.


sbillig:
Scratch that, this commit doesn't apply nicely because fe-v2's test-runner crate is behind master's. I'm tentatively inclined to do one last master -> fe-v2 merge before we delete all the old stuff on the v2 branch.

PR #1014, opened 2024-06-10
Ensure termination in trait solving

Y-Nak:
This PR ensures the termination in trait solving by adding a bound on type depth.
As the comments in the code describe, we'd need to revisit the trait solver when v2 is deployed.
This limitation doesn't break the soundness of the trait solver itself, but in some cases, users might need to add extra type annotations in their code.

### DIFF
--- a/crates/hir-analysis/src/ty/trait_resolution/proof_forest.rs
+++ b/crates/hir-analysis/src/ty/trait_resolution/proof_forest.rs
@@ -13,12 +13,21 @@ use crate::{
         canonical::{Canonical, Canonicalized},
         fold::{TyFoldable, TyFolder},
         trait_def::{impls_for_trait, Implementor, TraitInstId},
+        ty_def::{TyData, TyId},
         unify::PersistentUnificationTable,
+        visitor::{TyVisitable, TyVisitor},
     },
     HirAnalysisDb,
 };
-
-const MAXIMUM_SOLUTION_NUM: usize = 3;
+const MAXIMUM_SOLUTION_NUM: usize = 2;
+/// The maximum depth of any type that the solver will consider.
+///
+/// This constant defines the upper limit on the depth of types that the solver
+/// will handle. It is used as a termination condition to prevent the solver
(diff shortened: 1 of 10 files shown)


crates/hir-analysis/src/ty/trait_resolution/proof_forest.rs
```diff
@@ -439,3 +449,86 @@ impl ConsumerNode {
...
+/// - `db`: A reference to the HIR analysis database.
+/// - `ty`: The type for which the depth is to be computed.
+///
+/// # Returns
+/// The depth of the type as a `usize`.
+///
+/// # Note
+/// This function is a stop gap solution to ensure termination when the solver
+/// encounters coinductive cycles. It serves as a temporary solution until the
+/// parser can properly handle coinductive cycles.
```
sbillig, line 470: "parser"?

sbillig (approved):
Looks good. A max depth of 256 ends up being fairly slow (1.25s to run the vec clone uitest, vs <0.05s for other trait solver tests). Maybe a lower number would be better? (128 takes 0.35s for example, 64 is 0.13s; dunno what a reasonable limit is). This extra analysis time is only a problem in the LSP case, of course.

crates/hir-analysis/src/ty/trait_resolution/proof_forest.rs
Y-Nak: oh... fixed it. thanks.

Y-Nak:
@sbillig I tweaked the trait resolution process so that the solver prioritize consumer nodes that are closer to the original goal. This would mitigate the overhead since it hits the maximum solution limit of the original goal before hitting the depth limit(in the vec case, the solver takes 0.03s). This is of course an ugly stop-gap solution, but assuming that we improve the parser so that it can handle these cases properly without a depth limit, I think it'd be fine for now.

sbillig:
Looks great, this is a much better solution than my suggestion 😅

PR #1015, opened 2024-06-26
Update salsa

Y-Nak:
This PR updates salsa to salsa 3.0. The major changes are
1. Tracked structs and interned structs are required to have `'db` lifetime.
2. The actual Id(or field) of tracked/interned structs is a pointer to a data location now. This means we can't prefill the keyword in salsa 3.0(I simply removed the 'define_keyword` macros. I just removed the `fe-macro` crate.
3. We need to implement the `salsa::update::Update` trait to store a type in another tracked type. In normal cases, we can use `salsa::Update` derive macro. The only types that I implemented the trait manually are `Partial` and `IndexSet/IndexMap`.

Please refer to https://github.com/salsa-rs/salsa/pull/490 for more details.

@micahscopes I decided to leave the lsp crate as is, so it can't be compiled right now. Also, I temporarily excluded the language server crate from the workspace in https://github.com/ethereum/fe/pull/1015/commits/dce1e09e976ec88a2a346c549504cf53fd46ef00 to build/test the compiler itself. If you have any trouble migrating to salsa 3.0, please let me know.

@g-r-a-n-t This change would affect your driver implementation as well. So please take a quick look at the new driver implementation in this PR.


sbillig (approved):
👍 
That's a whole lotta lifetime annotations.

PR #1017, opened 2024-07-29
Add support for specifying generic arg in path segment

### LINKED ISSUE #988: Add support for specifying generic arg in path segment
Currently, it's not allowed to specify type args for each path segment.
This problem makes it impossible to call methods in some cases.
e.g., 
```rust
struct Foo<T> {
    t: T
}

impl<T> Foo<T> {
    fn method() -> i32 {
        1
    }
}

fn foo() {
    // Deciding the `Foo` type is not possible without a type argument for `Foo`.
    let x = Foo::method()
    
    // We need this! 
    let x = Foo<i32>::method()
}
```

sbillig:
fixes #988


Y-Nak (approved):
Great! (sorry for the delay)

PR #1019, opened 2024-09-17
Fix missing occurence check

Y-Nak:
This PR fixes a bug of missing an occurrence check in the proof tree implementation.


sbillig (approved):
👍

PR #1023, opened 2024-11-30
Appease clippy's new rules

micahscopes:
### What was wrong?
Clippy has been complaining about unnecessary explicit lifetimes.  In order to please clippy I found that I had to fix quite a few other things as well.

~I'm not sure that they're actually unnecessary though: https://github.com/rust-lang/rust-clippy/issues/12908~

~I'm opening this PR to find out.~

### How was it fixed?
`clippy --fix ...`


crates/hir-analysis/src/ty/ty_check/mod.rs
```diff
@@ -306,58 +358,6 @@ impl<'db> TyCheckerFinalizer<'db> {
     }
 
     fn check_unknown_types(&mut self) {
-        impl<'db> Visitor<'db> for TyCheckerFinalizer<'db> {
```
micahscopes, line 309: Clippy didn't like this trait being implemented inside of this function.

crates/test-utils/src/lib.rs
```diff
@@ -728,19 +729,17 @@ fn execute_runtime_functions(executor: &mut Executor, runtime: &Runtime) -> (Exi
...
         evm_runtime::CreateScheme::Legacy {
             caller: address(DEFAULT_CALLER),
         },
         U256::zero(),
         bytecode,
         None,
-    ) {
-        (reason, output)
-    } else {
-        panic!("EVM trap during test")
```
micahscopes, line 742: Clippy complained that this `else` block never gets reached and that the surrounding `if let` was unnecessary

crates/hir/src/hir_def/module_tree.rs
```diff
@@ -110,6 +110,7 @@ impl<'db> ModuleTree<'db> {
 /// top level modules. This function only depends on an ingot structure and
 /// external ingot dependency, and not depends on file contents.
 #[salsa::tracked(return_ref)]
+#[allow(elided_named_lifetimes)]
```
micahscopes, line 113: Clippy and the rust compiler were battling back and forth, on the one hand Clippy would complain about `needless_lifetimes` and change it to the elided version, but then the rust compiler would complain about `elided_named_lifetimes`.  So I added this allow rule here and one other place.

crates/hir/src/lib.rs
```diff
@@ -117,6 +117,7 @@ impl<'db> ModuleAnalysisPass<'db> for ParsingPass<'db> {
 // The reason why this function is not a public API is that we want to prohibit users of `HirDb` to
 // access `InputIngot` directly.
 #[salsa::tracked(return_ref)]
+#[allow(elided_named_lifetimes)]
```
micahscopes, line 120: another place where clippy was battling the compiler

sbillig (approved):
🙏

PR #1024, opened 2024-12-18
additional LSP diagnostics + fixups

micahscopes:
### language server changes
- more analysis passes in the language server
- default fallback handlers for unused LSP notifications/events
- graceful handling of sub-diagnostics with no spans


sbillig (approved):
👍

PR #1025, opened 2024-12-23
Fix language server in stdio mode

micahscopes:
### What was wrong?
The language server was crashing in stdio mode.  It turned out there was a `println` statement somewhere clogging up stdio, which should be reserved for JSON-RPC messages when running the language server in stdio mode.

### How was it fixed?
I just changed the `println` to a `debug` statement.  The VSCode extension now uses stdio without problems.  This will make it easier to support other editors.


sbillig (approved):
👍

PR #1026, opened 2024-12-24
Initial Zed editor support

micahscopes:
This relies on [`fe-lang/tree-sitter-fe`](https://github.com/fe-lang/tree-sitter-fe) which is basically a clone of [`tree-sitter-rust`](https://github.com/tree-sitter/tree-sitter-rust).

In the future we'll want to modify `tree-sitter-fe` to more faithfully adhere to Fe language syntax and features but for now this seems fine.

Currently we also rely on `fe-language-server` being in the path.  In the future it'd be nice if this could automatically build the language server.


sbillig:
Works great!

The editor stuff (and generally anything that isn't directly using compiler code) could probably live in one or more repos in the fe-lang org. I don't have strong feelings about this, but it feels appropriate. (We should get the website dir out of this repo too)

(I merged the other lsp PRs and rebased this, btw)

micahscopes:
Sounds great I'll do that now

micahscopes:
@sbillig any objections to extracting the new vscode extension out into the main branch here? https://github.com/fe-lang/vscode-fe

I'd move the existing contents of `vscode-fe#main` into an `old-version` branch or something like that.

sbillig:
Sounds good to me.

PR #1028, opened 2024-12-25
Remove editor extensions from monorepo

micahscopes:
See here:
- [vscode-fe](https://github.com/fe-lang/vscode-fe)
- [zed-fe](https://github.com/fe-lang/zed-fe)
- [nvim-fe](https://github.com/fe-lang/nvim-fe)


sbillig (approved):
🙏

PR #1029, opened 2024-12-31
fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0357

Also updated yanked crate `bytes`.


g-r-a-n-t (approved):
thank you

PR #1033, opened 2025-01-13
salsa 3.0

sbillig:
@Y-Nak I removed the InputIngot field from InputFile to avoid the infinite loop in the salsa object debug printing. This was a purely mechanical exercise of plumbing the appropriate InputIngot value around where it was needed in place of looking it up from the InputFile. Some of the impacted code could probably use some refactoring.

@micahscopes would you look at the latest commit's changes in the language server code? I added a `TODO` about a possibly incorrect IngotId being used for subdiagnostics. We probably need to look up the subdiagnostic's span's file's ingot manually there instead.


sbillig:
I'd be fine with this being merged as-is, btw. The subdiagnostic ingot thing could be addressed separately.

micahscopes:
@sbillig I took a look and it seems reasonable.  With respect to the subdiagnostic ingot issue, I'm fine with merging as is and adding a followup issue as there is another related cleanup chore that needs to happen: I need to ensure that `InputFile`s always have the correct path set, a relative path to the input ingot.

Y-Nak (approved):
👍

PR #1044, opened 2025-01-20
Add skeleton cli crate

sbillig:
@g-r-a-n-t All this does is `fe check foo.fe`. I had imagined loading a multi-file project, but I don't want to write code that would likely be replaced by your resolver stuff.


g-r-a-n-t (approved):
:+1:

PR #1048, opened 2025-01-24
Diagnostic refactor

sbillig:
Goal:

- Diagnostic enums should contain the (salsa) ids of the things involved in the error (eg the `TyId`s of the types for a type mismatch). These enum variants should generally be constructed directly, rather than via constructor functions that might generate strings, etc.
- The complexity of constructing helpful messages/labels for the user should be consolidated in the `to_complete(..) -> CompleteDiagnostic` functions.
- We should add convenience functions to the (salsa) id types where it's helpful.

I scrapped the `LazyDiagnostic` idea that I showed during the meeting, in favor of just moving the diagnostic stuff from the hir crate to the hir-analysis crate, so now `DiagnosticVoucher::to_complete` takes a `SpannedHirAnalysisDb` instead of a `SpannedHirDb`, etc. Its useful to have concrete span locations when generating the diagnostic text, so we can sort the labels by location and use text like "`x` defined first here", "`x` redefined here", etc and have it make sense.

I'll submit some follow-up PRs that actually improve diagnostics in the future, and to do a little more code shuffling. I'd like to merge this because it touches a lot of lines and could lead to conflict hassles for other stuff.


g-r-a-n-t (approved):
:+1:

PR #1050, opened 2025-01-27
Bump salsa to latest (includes accumulator bugfix), and use salsa::tracked methods instead of separate _impl fns

sbillig:
Salsa 3 supports tracked methods, so I inlined the standalone tracked `_impl` functions. No functionality changes.


micahscopes (approved):

micahscopes:
Awesome!

PR #1059, opened 2025-02-17
fix: Return exit code 1 when `fe-driver2` encounters errors

Turupawn:
This PR ensures that `fe-driver2` commands exits with code `1` when errors are detected, preventing silent failures.

This changes can help building CI automations [like this one](https://github.com/Turupawn/fe-examples) that I'm building now.

PR #1060, opened 2025-02-18
fix: Return exit code 1 when `fe check` encounters errors (continuation)

Turupawn:
replaces #1059

This PR ensures that `fe check` command exits with code 1 when errors are detected, preventing silent failures.

This changes can help building CI automations [like this one](https://github.com/Turupawn/fe-examples) that I'm building now.


sbillig (approved):
Thank you!

PR #1061, opened 2025-02-20
Salsa 3, part 3

sbillig:
Bumping the salsa revision again to pick up the latest breaking changes

- "coarse" tracked structs (struct fields aren't tracked as independent entities by default)
- tracked fn return types must impl `salsa::Update` trait
- `derive(salsa::Supertype)` allows us to avoid using `salsa::interned` wrapper structs around enums-of-salsa-ids (eg AdtRefId)


sbillig:
This has been updated to the latest salsa master, and is ready to merge.

Note that I had to replace IndexMap with OrderMap (because the latter implements the Hash trait), which is a light wrapper around IndexMap that changes some behavior to guarantee that the order remains consistent.

PR #1063, opened 2025-02-22
[experimental, for study purposes] salsa tracked core ingot input

micahscopes:
I couldn't help but try this out, for study purposes.

### DIFF
--- a/crates/language-server/src/backend/get_core.rs
+++ b/crates/language-server/src/backend/get_core.rs
@@ -1,0 +1,85 @@
+use camino::Utf8PathBuf;
+use common::{
+    input::{IngotKind, Version},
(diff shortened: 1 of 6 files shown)


crates/language-server/src/backend/get_core.rs
```diff
@@ -0,0 +1,85 @@
...
+    InputIngot::new(
+        db,
+        "core", // this should be handled by the driver crate default behavior
+        IngotKind::Core,
+        Version::new(0, 0, 1),
+        Default::default(),
+    )
+}
+
+pub fn init_core_ingot(db: &mut dyn InputDb) -> InputIngot {
```
micahscopes, line 23: This would be better implemented in the driver of course, I just kept it here for this sketch

PR #1065, opened 2025-03-07
Core library support in the language server.

g-r-a-n-t:
Using rust-embed and moved static core setup to `common::input`. Atm users get an error when goto selecting something in the core library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,9 +101,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.95"
+version = "1.0.97"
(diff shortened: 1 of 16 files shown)


sbillig (approved):
I'm fine with merging this as-is, but I think we need to be more careful with `InputIngot::core` (at least when we have multiple ingots in a single program). Each call to `InputIngot::core` returns a new InputIngot, so if we create two `LocalIngotContext`s via `LocalIngotContext::new`, they'll have different instances of the core ingot, and there will be two incompatible instances of the `Option` type, for example.

PR #1067, opened 2025-03-15
fix: perform `fe check` on the current path `.`

Turupawn:
This makes possible perform a `fe check .` by converting the both the root file and all the files in the core library directory to it's absolute paths when looking for the library entrypoint root file. (`core/src/lib.fe`).


g-r-a-n-t (approved):
looks good

PR #1068, opened 2025-03-28
Resolution handler

g-r-a-n-t:


crates/common/src/config.rs
```diff
@@ -96,54 +100,59 @@ impl Config {
...
-#[derive(Debug, Clone)]
-pub enum DependencyDescription {
-    Path(Utf8PathBuf),
-    PathWithArguments {
-        path: Utf8PathBuf,
-        arguments: IngotArguments,
-    },
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct DependencyDescription {
+    pub url: Url,
```
micahscopes, line 111: This might not be expressive enough to handle both relative and absolute dependency paths.

crates/common/src/config.rs
```diff
@@ -96,54 +100,59 @@ impl Config {
...
+    pub arguments: Option<IngotArguments>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Dependency {
```
micahscopes, line 116: It could be really nice if `Dependency` is just a flat thing, something like:

```rs
Dependency {
  alias: SmolStr,
  location: Either<Url, Utf8PathBuf>,
  arguments: Option<IngotArguments>
}

File: crates/common/src/ingot.rs [hunk 3]
```diff
@@ -87,22 +85,59 @@ impl Ingot<'_> {
...
+            db.workspace().items_at_base(db, self.base(db))
+        }
+    }
+
+    #[salsa::tracked]
+    pub fn config(self, db: &'db dyn InputDb) -> Option<Config> {
+        db.workspace()
+            .containing_ingot_config_file(db, self.base(db))
+            .map(|config_file| Config::from_string(config_file.text(db).clone()))
+    }
```
micahscopes, line 105: Here's where the magic happens, where we just parse the config from the config's `File` input into a `Config` object.  I added all the derivations to `Config` so that it can be salsa tracked.

crates/resolver/src/ingot.rs
```diff
@@ -47,16 +47,23 @@ impl ResolutionHandler<FilesResolver> for BasicIngotNodeHandler {
             return config
                 .dependencies
                 .into_iter()
-                .map(|dependency| match dependency.description {
-                    DependencyDescription::Path(path) => (
-                        ingot_path.join(path).canonicalize_utf8().unwrap(),
```
micahscopes: Seems like the resolver was handling relative paths quite well before.  If `Dependency`/`DependencyDescription` can only store a `Url` this gets trickier.

crates/hir/src/hir_def/mod.rs
```diff
@@ -80,11 +83,18 @@ impl<'db> HirIngot<'db> for Ingot<'db> {
...
             .iter()
-            .filter_map(|(name, url)| {
-                self.index(db)
-                    .containing_ingot(db, url)
-                    .map(|ingot_description| (IdentId::new(db, name), ingot_description))
-            })
+            .filter_map(
+                |Dependency {
+                     alias,
+                     description: DependencyDescription { url, .. },
```
micahscopes, line 89: The oversimplified `DependencyDescription` with its handy `Url` is making this a little too easy.  In the end we need to be able to consistently map a `Dependency` to a URL.  In the case of relative path dependencies, there needs to be some way of injecting a base URL so that if the dependency points to a relative path we can resolve an absolute URL for that relative path.

Maybe we can have a method like:
```rust
impl Dependency {
  fn url(&self, base_url: Url) -> Url {
    let flattened_absolute_url = match self.location ... // handle path vs url
    return flattened_absolute_url
}
```

The idea would be that the dependency itself carries around either a url or a path and that this can be resolved given a base url.

crates/resolver/src/ingot.rs
g-r-a-n-t: I don't think the dependency struct should hold onto a url, unless it is a git repo or something. it should just be a simple representation of what is in the config file. it's pretty easy for a function in common::ingot to map the path `DependencyDescription` to a file url given the parent ingot.

crates/common/src/ingot.rs
```diff
@@ -87,22 +85,59 @@ impl Ingot<'_> {
...
+
+    #[salsa::tracked]
+    pub fn version(self, db: &'db dyn InputDb) -> Option<Version> {
+        self.config(db).map(|config| config.ingot.version).flatten()
+    }
+
+    #[salsa::tracked]
+    pub fn dependencies(self, db: &'db dyn InputDb) -> Vec<Dependency> {
```
g-r-a-n-t, line 113: this could just return the aliases and urls and the dependency descriptions can remain relative

sbillig:
> Readd support for --core flag.

I think we should drop support for specifying a custom core lib, and only support a custom std lib (when we figure out what the std lib should be). Core should be limited to things that wouldn't be interesting to replace. The compiler depends on core containing a certain set of traits and impls.

crates/common/src/tree.rs
micahscopes, line 1: What's your thinking for keeping this in `common` vs in `fe` or in the resolver?

crates/common/src/graph.rs
micahscopes, line 1: The naming in here seems pretty generic, any reason not to have more specific naming, even just at the module level?  So that we could have something like `dependencies::Graph` or `dependency_info::Graph`? Or were you thinking this could potentially be useful for creating lots of different kinds of graphs, not just for dependencies?

crates/language-server/src/functionality/handlers.rs
micahscopes, line 1: Loving all these consolidations :sunglasses:

micahscopes (approved):
Overall looks good to me!  Only changes I'd suggest are organizational:
- keeping the tree structure out of `common` crate since it's part of a pretty specific feature
- using less generic names for the dependency graph stuff in common

crates/common/src/tree.rs
g-r-a-n-t: this is used in both the `driver` and `fe` crates to format trees. `fe` is off the table due to `driver` needing it and seeing as this module just deals with the printing of dependency trees I don't think it belongs in the `resolver` crate.

crates/common/src/graph.rs
g-r-a-n-t: I think this makes sense. the dependency tree formatting stuff could also be moved into that module, which would maybe make things a bit cleaner.

PR #1070, opened 2025-03-31
Use salsa tracking to link `InputFile` to `InputIngot`

micahscopes:

PR #1072, opened 2025-04-01
Remove `InputIngot` constructor helper

micahscopes:
I'm having second thoughts on how useful this is :sweat_smile:


sbillig:
> I'm having second thoughts on how useful this is 😅

It's only useful if you also remove the subsequent `.set_files` calls.

micahscopes:
Ahh, right!  Thank you

micahscopes:
Hmm, I'm realizing now that this would clash with the changes in #1070

micahscopes:
I moved over to here: #1073

PR #1078, opened 2025-04-09
lsp: fix hover/goto for type aliases

sbillig:
Type aliases are now resolved as `PathRes::TyAlias`, so the language server can get the proper span to use for hover and goto.

Best to look at the commits separately. The second commit is cleanup that I missed in #1074.


micahscopes (approved):
Awesome, thank you!

PR #1081, opened 2025-04-18
Ingot file management part 1

### LINKED ISSUE #1076: Make file/ingot inputs more granular, relying more on salsa queries
Grant expressed the need for a better way of marking a config file in its ingot. Instead of just making another input type, we discussed the possibility of instead making `InputFile` a tracked struct derived from ingot, path and contents inputs. Using this approach, we could have a salsa query that returns the role of a file given these ingredients, and another that maintains an index of paths in an ingot.

micahscopes:
Fixes #1076

Todo:
- [x] Tests passing
- [ ] Form better understanding of import salsa cycle
- [ ] Better error handling
  - [ ] don't print to stderr/stdout
  - [ ] use `Result`; add error types
- [ ] settle on file index naming (Workspace?)
- [ ] implement language server "sync" functionality with resolver


g-r-a-n-t (approved):
looks goof

PR #1086, opened 2025-05-20
Update README.md

MonkeyKing44:
fix old link, might be problem with old one, so change  it to a new one


micahscopes:
Looks like both links are working now.

Appreciate your taking a look though @MonkeyKing44.

PR #1087, opened 2025-05-23
Driver debugging session with Grant

micahscopes:


crates/driver/src/lib.rs
```diff
@@ -20,7 +20,8 @@ pub fn run(opts: &Options) {
     match &opts.command {
         Command::Build => eprintln!("`fe build` doesn't work at the moment"),
         Command::Check { path, core } => {
-            let base_url = Url::from_directory_path(path).expect("failed to parse base URL");
+            let base_url = Url::from_directory_path(path.canonicalize().unwrap())
```
g-r-a-n-t, line 23: after making this change we get the hanging behavior

PR #1088, opened 2025-05-23
Pattern matching reimplementation

micahscopes:
This is a very heavily LLM driven (gemini + claude) reimplementation of pattern matching and decision tree stuff from #770.

To-do:
- [x] Distilled implementation to closely match #770
- [x] Scrutinize and consolidate test cases


micahscopes:
yay! thank you for all your help

PR #1094, opened 2025-05-30
Fix `fe check` for standalone ingots

### LINKED ISSUE #1093: `fe check singlefile.fe` doesn't work
Dunno how important this is, but `fe check foo.fe` is convenient for testing simple code.

```
fe check ~/code/eth/result.fe
ingot root file is missing
Files in ingot:
  file:///Users/sean/code/eth/result.fe: Some("result.fe")

thread 'main' panicked at crates/hir/src/hir_def/module_tree.rs:243:13:
ingot root file is missing
```

micahscopes:
fixes #1093

### DIFF
--- a/crates/driver/src/lib.rs
+++ b/crates/driver/src/lib.rs
@@ -112,7 +112,7 @@ pub fn run(opts: &Options) {
                             Some(content),
                         );
                     }
-                    base_url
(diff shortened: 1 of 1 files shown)


sbillig (approved):
🙏

PR #1099, opened 2025-06-19
Don't print to stdio/stderr; use tracing

micahscopes:
The language server supports stdio connections, and printing to stdio/stderr can break the pipe.  Let's use tracing instead!

### Changes:
- [x] Clippy won't let you log to stdio directly, not with `println`, not with `eprintln`, so don't even think about itt
- [x] Logging everywhere has been switched to tracing
- [x] Enable default log level in the Fe CLI
  - `RUST_LOG` level is respected by Fe CLI too 

Now if you want to print something in a test, you'll need to do e.g. `tracing::debug!("bla bla bla")`. Use `RUST_LOG=debug cargo test` to see the logs from tests.


crates/bench/src/main.rs
```diff
@@ -1,3 +1,5 @@
+use tracing::info;
+
 fn main() {
-    eprintln!("run `cargo bench`");
+    info!("run `cargo bench`");
```
sbillig:
```suggestion
    error!("run `cargo bench`");
```

crates/driver/src/lib.rs
```diff
@@ -14,11 +14,12 @@ use resolver::{
     ingot::{source_files::SourceFiles, Ingot, IngotResolver},
     Resolver,
 };
+use tracing::{error, warn};
 use url::Url;
 
 pub fn run(opts: &Options) {
     match &opts.command {
-        Command::Build => eprintln!("`fe build` doesn't work at the moment"),
+        Command::Build => warn!("`fe build` doesn't work at the moment"),
```
sbillig:
```suggestion
        Command::Build => error!("`fe build` doesn't work at the moment"),
```

crates/driver/src/lib.rs
```diff
@@ -144,7 +145,7 @@ pub fn run(opts: &Options) {
                 std::process::exit(1);
             }
         }
-        Command::New => eprintln!("`fe new` doesn't work at the moment"),
+        Command::New => warn!("`fe new` doesn't work at the moment"),
```
sbillig:
```suggestion
        Command::New => error!("`fe new` doesn't work at the moment"),
```

PR #1104, opened 2025-07-15
Use resolver and driver db in language server

### LINKED ISSUE #1075: move ingot loading code from language-server to driver, for use in the fe cli

micahscopes:
(please merge #1103 and #1102 first)

resolves #1075

- Uses the resolver as is instead of all that workspace sync stuff
- Removes the language server salsa db, instead implementing diagnostics functionality on top of the driver db


crates/driver/src/lib.rs
```diff
@@ -19,14 +19,14 @@ use url::Url;
...
-        Command::Build => eprintln!("`fe build` doesn't work at the moment"),
+        Command::Build => tracing::error!("`fe build` doesn't work at the moment"),
         Command::Check { path, core } => {
             let mut db = DriverDataBase::default();
             let mut ingot_resolver = IngotResolver::default();
 
             let core_url = if let Some(core_path) = core {
                 if !core_path.exists() {
-                    eprintln!("the core path `{core_path}` does not exist");
+                    tracing::error!("the core path `{core_path}` does not exist");
```
g-r-a-n-t: why should this be changed? it's my understanding that the user needs to set the RUST_LOG env variable to see tracing

micahscopes: I set up a tracing subscriber in the Fe CLI that prints error level logs by default

micahscopes: (Basically this is for appeasing the new clippy rules, see #1099)

micahscopes: (Tldr the idea is to avoid directly using stdout/stderr by default in crates upstream of the `fe` and `language-server` crates, which each need well defined stdout/stderr behavior)

g-r-a-n-t: I think `eprintln` is better suited for this. tracing is meant for understanding runtime behavior and these are just messages generated by the CLI to help the user. I understand the need for tracing, but changing these messages results in what I think is cluttered output. it also doesn't seem like the LSP should be calling this function, so it shouldn't be a problem.

micahscopes: I definitely hear your concerns about the cluttered output and wanting to distinguish between output/logging designed for end users to see vs runtime debugging.  Two ideas:

1. Tracing subscriber formatting is highly configurable, we could make the CLI output look really simple and clean, equivalent to simple print statements
2. I've come to see tracing itself as being not only a runtime debugging logger, and moreso a unified logging manager. My intuition was that for the CLI output, setting the subscriber to print error and info level logging statements would be appropriate for outputs aimed at the end user, whereas debug level log entries would be hidden by default. Alternatively we could have a custom tracing span designed to be printed in the CLI. 

I really like the idea of a well organized unified logging system, but some more alternatives could be:
- overriding clippy here
- moving this stuff to the `fe` crate which doesn't have the prohibition on eprintln/println

micahscopes: Just learned about tracing's "target" filter, which seems perfect for this

micahscopes: I took these tracing statements out for now, will make a followup next week to explore the options

PR #1111, opened 2025-08-19
Add support for generic parameter defaults

sbillig:
```
struct Foo<T = i32> {
  inner: T
}

trait Add<T: Self> {
  type Output = Self
  fn add(self, _ other: T) -> Self::Output
}
```

I also moved the type alias analysis into DefAnalyzer, where the other item definitions are analyzed. This fixes some missing type alias diagnostics.


micahscopes (approved):
Looks good

PR #1112, opened 2025-08-19
Reverse lazy span and language server goto/hover improvements

micahscopes:


micahscopes:
@sbillig still not quite there but this one is much improved

PR #1117, opened 2025-09-02
Fix language server content update handler

micahscopes:


micahscopes:
@g-r-a-n-t this will make the language server actually work

g-r-a-n-t (approved):

PR #1119, opened 2025-09-03
add support for operator overloading via ops traits

sbillig:
Notes:
- Unary plus is currently hardcoded as a no-op. I think we should remove parser support for it. I dunno what it would do. (rust doesn't support it, fwiw)
- The index op on arrays can't be implemented in fe yet; `impl<T, const N: usize> Index<usize> for [T; N]` fails on the use of `N` in `[T; N]` (we only int literals in that position right now) (issue #1120)
- Unlike rust, we currently have a BitNot trait in addition to Not. We could merge them.

eg
```rust
use core::ops::Add
impl Add<i32> for Point {
  fn add(self, _ n: i32) -> Point {
    Point { x: self.x + n, y: self.y + n }
  }
}

fn f(p: Point) -> Point {
  p + 100
}
```


sbillig:
Ignore the fmt issue plz, it's addressed in #1121

micahscopes (approved):

PR #1142, opened 2025-09-30
`const` fixes

### LINKED ISSUE #1120: add support for using const ty param as array size
for example:

```rust
struct A<const N: usize> {
  inner: [u8; N] // error
}
struct B<const N: usize> {
  inner: A<N> // this is fine
}
```
```
error[3-0015]: the expression is not supported yet in a const type context
  ┌─ foo.fe:2:15
  │
2 │   inner: [u8; N]
  │               ^ only literal expression is supported
```

sbillig:
Two things:
- fixes #1120 
- fixes use of const defs in const arg position
eg
```rust
const SIZE: usize = 10
type MyString = String<SIZE> // 
```

Shoulda been separate commits, but oh well.

### DIFF
--- a/crates/hir-analysis/src/name_resolution/path_resolver.rs
+++ b/crates/hir-analysis/src/name_resolution/path_resolver.rs
@@ -2,7 +2,7 @@ use common::indexmap::IndexMap;
 use either::Either;
 use hir::{
     hir_def::{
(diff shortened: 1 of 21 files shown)


g-r-a-n-t (approved):
:+1:

PR #1147, opened 2025-10-15
effects lite, phase 1

sbillig:
This implements parsing and analysis for trait and type effects/capabilities.

```rust
pub struct Storage {
    value: u64,
}

pub fn mint(_ amount: u64)
  uses (mut st: Storage)
{
    st.value += amount
}

pub fn example() {
    let mut st = Storage { value: 0 }
    with (Storage = st) {
        mint(5)
        mint(10)
    }
}
```

```rust
pub trait Logger {
    fn log(mut self, _ msg: String<10>)
}

pub struct Console {}
impl Logger for Console {
    fn log(mut self, _ msg: String<10>) { ... }
}

pub fn log_hello()
  uses (mut logger: Logger)
{
    logger.log("hello")
}

pub fn do_some_logging() {
    with (Logger = Console {}) {
        log_hello()
        log_hello()
    }
}
```

### DIFF
--- a/crates/hir-analysis/src/diagnostics.rs
+++ b/crates/hir-analysis/src/diagnostics.rs
@@ -3,6 +3,7 @@
 //! implement [`DiagnosticVoucher`] which defines the conversion into
 //! [`CompleteDiagnostic`].
 
+use crate::ty::trait_def::TraitInstId;
 use crate::{
     HirAnalysisDb,
     name_resolution::diagnostics::{ImportDiag, PathResDiag},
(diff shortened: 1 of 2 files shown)


cburgdorf (approved):
Looks great!

The only thing that I spotted (ok, I'll be honest Codex spotted it, but I confirmed it) is that the `check_callable_effects` isn't yet called for method calls.

So, this code should fail but doesn't:

```
pub struct Storage {
    value: u64,
}

pub struct Worker {}

impl Worker {
    pub fn run(self)
      uses (mut storage: Storage)
    {
        storage.value += 1
    }
}

pub fn exercise(worker: Worker) {
    worker.run()
}
```

Should probably also add another snap for that.

PR #1148, opened 2025-11-02
Add support for associated consts

sbillig:
rust-style trait-associated consts. Eg
```rust
trait Foo {
  const BAR: u8 = 10 // optional default value
}
impl Foo for MyStruct {
  const BAR: u8 = 100
}
fn f() {
  let x = MyStruct::BAR
  let y: [u32; MyStruct::BAR]
}
```

### DIFF
--- a/crates/uitest/fixtures/ty_check/ambiguous_assoc_const.snap
+++ b/crates/uitest/fixtures/ty_check/ambiguous_assoc_const.snap
@@ -5,7 +5,7 @@ expression: diags
 input_file: fixtures/ty_check/ambiguous_assoc_const.fe
 ---
 error[2-0018]: multiple trait candidates found
-   ┌─ ambiguous_assoc_const.fe:21:10
+   ┌─ ambiguous_assoc_const.fe:20:10
    │
  1 │ trait C1 {
    │       -- candidate: `<Foo as C1>::X`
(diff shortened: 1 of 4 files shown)


crates/hir-analysis/src/name_resolution/path_resolver.rs
```diff
@@ -696,6 +741,26 @@ where
...
+                                trait_insts: traits,
+                            },
+                            path,
+                        ));
+                    }
+                    AssocConstSelection::NotFound => {}
+                }
+            }
             // Fast path: `<A as Trait>::Assoc` — return the projection directly.
             if let TyData::QualifiedTy(trait_inst) = ty.data(db)
```
cburgdorf: As this handles associated *types* for explicitly qualified traits, should this also handle associated consts of that form e.g. `<A as Trait>::TIMEOUT`?

crates/uitest/fixtures/ty_check/ambiguous_assoc_const.snap
```diff
@@ -0,0 +1,19 @@
...
+ 1 │ trait C1 {
+   │       -- candidate: `<Foo as C1>::X`
+   ·
+ 5 │ trait C2 {
+   │       -- candidate: `<Foo as C2>::X`
+   ·
+20 │     Foo::X
+   │          ^ `X` is ambiguous
+   │
+   = hint: specify the trait explicitly: `<Foo as C1>::X`
```
cburgdorf: There doesn't seem to be a test that exercises in the explicit form that the hint suggests. I'm also not sure if it works already (hence my comment earlier). I haven't tried the PR locally. Still need to get my Rust setup working again.

PR #1150, opened 2025-11-07
Introduce first parts of MIR + YUL Codegen

cburgdorf:
This aims to be the first deliverable of bringing mir + codegen back to eventually be able to produce deployable bytecode. There's **a lot** left to do but I'm thinking it would be better to get this reviewed and merged so that we have a starting point for further collaboration on that front.

**Scope**
- Added the standalone `fe-mir` crate with MIR data structures, lowering logic, control-flow metadata, and call tracking so every HIR function now lowers into basic blocks plus loop/match annotations.
- Introduced the `fe-codegen` crate that consumes MIR and emits minimal Yul: locals, arithmetic/logical ops, loops, literal `match` switches, and intra-module function calls are rendered into readable Yul via a small document builder.
- Extended `fe check` with `--dump-mir` and `--emit-yul-min`, wiring these flags into the new MIR → Yul pipeline and adding the `mir` / `codegen` dependencies to the CLI.
- Added a suite of snapshot tests that run the new emitter across small Fe programs (arithmetic, bit ops, control flow, matches, calls, etc.) to lock in the emitted Yul text.

**MIR + Yul Status**
- MIR handles `let`, assignments, aug-assign, expression statements, returns, `while`, `if`, literal-only `match`, and records `break` / `continue` destinations plus match arm metadata; calls keep their resolved `Callable` and argument values for downstream use.
- Lowered loops carry `LoopInfo` so the emitter can recreate structured Yul `for` loops, and literal matches become `switch` terminators with per-arm details.
- The emitter rebuilds identifiers, lowers arithmetic/logical/unary expressions, supports tuples and simple blocks, emits `if` / `while` / `switch` constructs, and turns resolvable function or method calls into Yul calls, storing the final result in `ret`.
- `fe check --emit-yul-min` prints each function’s Yul body for inspection, while the tests assert the textual Yul for every sample program.

**Still To Do**
1. Lower and emit `for` loops plus any other statements currently rejected (they still panic during MIR lowering).
2. Broaden `match` coverage beyond unique literal arms plus a single wildcard so richer patterns and enums work.
3. Support destructuring in patterns and assignment targets within the emitter to unlock tuple/struct bindings.
4. Expand expression support (struct literals, member access, indexing, storage/memory ops, etc.) which currently raise “unsupported” errors.
5. Allow codegen to reference runtime/contract functions, builtins, and external ABIs when emitting calls, not just HIR functions defined in the same module.
6. Integrate the emitted Yul into the real build pipeline with ABI/memory handling; today it’s only a minimal `function … -> ret` form dumped for debugging.

### DIFF
--- a/crates/codegen/tests/fixtures/match_literal.snap
+++ b/crates/codegen/tests/fixtures/match_literal.snap
@@ -13,6 +13,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 2
       }
+    ret := v0
   }
 }
 {
@@ -28,6 +29,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -43,6 +45,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -58,6 +61,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -73,6 +77,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -88,6 +93,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -103,6 +109,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -118,6 +125,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
@@ -133,6 +141,7 @@ input_file: tests/fixtures/match_literal.fe
       default {
         v0 := 3
       }
+    ret := v0
   }
 }
 {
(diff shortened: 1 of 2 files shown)


crates/codegen/tests/fixtures/match_literal.snap
```diff
@@ -0,0 +1,197 @@
...
+    let v0 := 0
+    switch e
+      case 1 {
+        v0 := 1
+      }
+      default {
+        v0 := 2
+      }
+  }
+}
```
sbillig, line 18: This should assign `ret := v0` (or equivalent). (Same problem in all match tests)

sbillig (approved):
🤘 great start

crates/codegen/tests/fixtures/match_literal.snap
cburgdorf: Good catch. Pushed another commit to fix that

cburgdorf:
> > fe check --emit-yul-min prints each function’s Yul body for inspection, while the tests assert the textual Yul for every sample program.
> 
> @cburgdorf this might be a good opportunity to start building out the `build` subcommand.

Yes, I agree that this is what we eventually want to land at. There's just so much left to be done to get to the point where we can compile a meaningful Fe example down to a deployable contract in bytecode that I don't bother about the fine details for now.

We also want something like `fe test` sooner than later. Maybe you'd be interested taking that on? I'll probably hack something together for intermediate use because I'm getting to a point where I would really be able to run the YUL code that I produce to test it for correctness. At the moment I just look at the generated YUL and just check correctness in my head but I don't want to code in that "correctness vacuum" for too long.

PR #1157, opened 2025-11-17
fe build

g-r-a-n-t:
This change introduces a `fe build` command for compiling Fe contracts to EVM bytecode. It supports two backends: Sonatina (default, no solc needed) and a Yul backend that invokes solc. You can target a standalone .fe file, an ingot directory, a workspace root, or a workspace member by name; optionally filter to a single contract; choose optimization level; and control the output directory.

Here is a comprehensive writeup on the current CLI behavior: https://github.com/g-r-a-n-t/fe/blob/73e58ee102a18d25bd2d65aca25e4ce31ed7435a/CLI.md


sbillig:
@codex review

PR #1158, opened 2025-11-17
 Remote ingot dependencies

g-r-a-n-t:
Added remote dependency support.                                                                                                                                                                       
                                                                                                                                                                                              
  - Introduced resolver/git plumbing (crates/resolver/src/git.rs, git_stub.rs) with new fixtures; extended                                                                                                                                      
    resolver graph/ingot logic.                                                                                                                                                               
  - Expanded common dependency graph and cache to track remote locations; reorganized display/tree code under                                                                                                                                     
    common/src/dependencies/tree.rs.                                                                                                                                                          
  - Driver now handles remote ingots (new ingot_handler), with updated dependency tree diagnostics and config handling.                                                                                                                                                                                 
  - `fe tree` simplified.

### DIFF
--- a/crates/resolver/src/git.rs
+++ b/crates/resolver/src/git.rs
@@ -144,12 +144,23 @@ impl GitResolver {
                 error,
             }
         })?;
-        let object =
-            repo.find_object(oid, None)
-                .map_err(|error| GitResolutionError::RevisionLookup {
+        let object = match repo.find_object(oid, None) {
+            Ok(object) => object,
+            Err(error) if description.source.scheme() == "file" => {
+                return Err(GitResolutionError::RevisionLookup {
                     rev: description.rev.clone(),
                     error,
-                })?;
+                });
+            }
+            Err(_) => {
+                self.fetch_default_remote(repo, description)?;
+                repo.find_object(oid, None)
+                    .map_err(|error| GitResolutionError::RevisionLookup {
+                        rev: description.rev.clone(),
+                        error,
+                    })?
+            }
+        };
         let mut builder = CheckoutBuilder::new();
         builder.force();
         repo.checkout_tree(&object, Some(&mut builder))
@@ -162,6 +173,26 @@ impl GitResolver {
                 rev: description.rev.clone(),
                 error,
             })?;
+        Ok(())
+    }
+
+    fn fetch_default_remote(
+        &self,
+        repo: &Repository,
+        description: &GitDescription,
+    ) -> Result<(), GitResolutionError> {
+        let mut remote =
+            repo.find_remote("origin")
+                .map_err(|error| GitResolutionError::FetchRepository {
+                    source: description.source.clone(),
+                    error,
+                })?;
+        remote.fetch::<&str>(&[], None, None).map_err(|error| {
+            GitResolutionError::FetchRepository {
+                source: description.source.clone(),
+                error,
+            }
+        })?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


micahscopes:
can you think of any obvious ways to break up tree.rs into a few smaller modules just to make it easier to navigate? not super important

Cargo.toml
```diff
@@ -25,6 +25,7 @@ derive_more = { version = "1.0", default-features = false, features = [
 ] }
 dir-test = "0.4"
 glob = "0.3.2"
+indicatif = "0.17"
```
g-r-a-n-t: remove

crates/common/src/dependencies/tree.rs
```diff
@@ -35,6 +35,7 @@ pub fn display_tree(
     graph: &DiGraph<Url, (SmolStr, DependencyArguments)>,
```
g-r-a-n-t, line 105: This should be moved to `DependencyGraph`. Maybe add a separate `DependencyTree` type.

crates/driver/Cargo.toml
```diff
@@ -23,3 +23,7 @@ resolver.workspace = true
 url.workspace = true
 tracing.workspace = true
 mir.workspace = true
+indicatif.workspace = true
```
g-r-a-n-t: remove

crates/common/src/ingot.rs
```diff
@@ -142,7 +143,16 @@ impl<'db> Ingot<'db> {
             Some(config) => config
                 .dependencies(&base_url)
                 .into_iter()
-                .map(|dependency| (dependency.alias, dependency.url))
+                .map(|dependency| {
+                    let url = match &dependency.location {
+                        DependencyLocation::Remote(remote) => db
+                            .dependency_graph()
+                            .local_for_remote_git(db, remote)
```
g-r-a-n-t, line 150: This should be tracked in `workspace`.

crates/driver/src/lib.rs
```diff
@@ -3,80 +3,127 @@
...
+pub struct SpinnerProgress;
 
-        // Root ingot resolution should never fail since directory existence is validated earlier.
-        // If it fails, it indicates a bug in the resolver or an unexpected system condition.
-        if let Err(err) = ingot_graph_resolver.graph_resolve(&mut handler, ingot_url) {
-            panic!(
-                "Unexpected failure resolving root ingot at {ingot_url}: {err:?}. This indicates a bug in the resolver since directory existence is validated before calling init_ingot."
-            );
-        }
+impl SpinnerProgress {
```
g-r-a-n-t: remove

crates/common/src/ingot.rs
g-r-a-n-t: @micahscopes I think little organizational things like this can wait for a later pr.

crates/resolver/src/git.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+        Ok(())
+    }
+
+    fn ensure_checkout(
+        &self,
+        description: &GitDescription,
+        checkout_path: &Utf8Path,
+    ) -> Result<CheckoutStatus, GitResolutionError> {
+        if checkout_path.exists() {
```
cburgdorf, line 105: Maybe I'm reading it wrong but if the `checkout_path` already exists then we attempt to checkout the revision without fetching from the remote first. That could lead to a scenario where the revision isn't yet available in the local checkout, no?

crates/common/src/ingot.rs
micahscopes: sounds good to me!

PR #1159, opened 2025-11-17
Introduce crate to interact with solc to compile YUL to deployable binary code

cburgdorf:
This introduces a new crate that interacts with `solc` to compile YUL code to deployable binary code. The API is largely identical to what Fe had at its last release. The main difference is that `solc` is expected to be exposed on the system as `FE_SOLC_PATH` and then invoked there. The main motivation to do that is to simply not deal with all the C++ stuff for now. A secondary benefit is that windows support basically would come for free (assuming there aren't any other windows blocker). I'm not saying this is the best long term solution, but I consider it good enough for now.

For CI we pin a specific solidity version and cache them so that we don't redownload it over and over with every CI run. I also added a test that uses `revm` to execute a method on a small contract and tests the result. This will probably move elsewhere when we bring back a proper `fe test` command.

The runner isn't yet hooked up with our YUL generation pipeline but it shouldn't be long to get there.


sbillig (approved):
LGTM

PR #1162, opened 2025-11-20
Traversal API

micahscopes:
This is a start at flipping the internal API of the hir/hir-analysis crates inside-out, so that our hir analysis queries and program traversal logic can be nicely reused without requiring developers to have an intricate understanding of how to use each query, and without duplicating visitor logic.

The goal is support followups on language server features, code transformations, and hopefully even codegen for graph-based IRs.

It's sort of like a "virtually materialized HIR interface", since our middle layer is query based we don't have a concrete materialization of the HIR, so this is kind of like an in-between solution.

There's a bunch of stuff in here:
- hir + hir-analysis have been merged
- a new `core` module keeps syntactic hir accessors contained from analysis logic
  - provides a semantic-level public interface  
- the analysis::def_analyzer has been fully metabolized into this new interface (others ought to follow)
- a new `Diagnosable` trait provides proof of concept for a pull-based diagnostic gathering pattern (which could be a really good model in designing interfaces to collect info for debugging/language server/graph output features)
- lots of reworked/dissolved indirection layers from the old hir-analysis crate, more to come

Needs to be rebased due to rapid developments elsewhere in the codebase!

Will update this later today or tomorrow with more info.

Built from hard lessons learned in #1163 #1131 #1112

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,7 @@ env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
   RUST_BACKTRACE: full
+  SOLC_VERSION: v0.8.30
 
 jobs:
   lint:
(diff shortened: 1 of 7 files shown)


crates/hir/src/core/hir_def/semantic.rs
sbillig, line 1: defunct

crates/hir/src/core/hir_def/item.rs
```diff
@@ -617,12 +609,13 @@ pub struct Func<'db> {
...
 
     pub name: Partial<IdentId<'db>>,
-    pub attributes: AttrListId<'db>,
-    pub generic_params: GenericParamListId<'db>,
-    pub where_clause: WhereClauseId<'db>,
+    pub(in crate::core) attributes: AttrListId<'db>,
+    pub(in crate::core) generic_params: GenericParamListId<'db>,
+    pub(in crate::core) where_clause: WhereClauseId<'db>,
+    // TODO: Add semantic view for FuncParamListId and restrict visibility
     pub params: Partial<FuncParamListId<'db>>,
```
sbillig: This TODO is partially addressed by `Func::param_views`, could be completed

crates/hir/src/diagnosable.rs
sbillig, line 1: Should this be outside of `hir::core`, so that it can't use the hir-core-only fields? It seems like it should only rely on public accessors, not (semi-)private fields.

crates/hir/src/core/semantic/diagnostics.rs
```diff
@@ -0,0 +1,1803 @@
...
+        out.extend(self.diags_parameters(db));
+        out.extend(self.diags_param_types(db));
+
+        // Defer body/type-check diagnostics to BodyAnalysisPass; here we only validate the
+        // explicit return type annotation when present.
+        out.extend(self.diags_return(db));
+
+        // Where-clause diagnostics (trait bounds and kind checks on where predicates)
+        {
+            let clause = WhereClauseView {
```
sbillig: Ugly pattern sneaking in because this file is inside `hir::core`. `self.where_clause()` should return a `WhereClauseView`.

crates/hir/src/core/semantic/diagnostics.rs
```diff
@@ -0,0 +1,1803 @@
...
+        }
+        out
+    }
+}
+
+impl<'db> Diagnosable<'db> for TypeAlias<'db> {
+    type Diagnostic = TyDiagCollection<'db>;
+
+    fn diags(self, db: &'db dyn HirAnalysisDb) -> Vec<Self::Diagnostic> {
+        let mut out = self.analyze(db);
```
sbillig: I think we should just inline these `analyze()` functions and eliminate them; they're only called once, and it adds another jump to find the actual logic.

crates/hir/src/core/semantic/diagnostics.rs
```diff
@@ -0,0 +1,1803 @@
...
+    }
+
+    /// Compare impl methods vs. trait methods and report missing/mismatched ones.
+    pub(crate) fn diags_method_conformance(
+        self,
+        db: &'db dyn HirAnalysisDb,
+        implementor: Binder<ImplementorId<'db>>,
+    ) -> Vec<TyDiagCollection<'db>> {
+        implementor.skip_binder().diags_method_conformance(db)
+    }
```
sbillig: remove?

sbillig (approved):
🥳

 I think there's more cleanup we could do, but I'd be happy for this to be merged now.

PR #1165, opened 2025-11-25
Improved reference <> definition support

micahscopes:
Builds on #1162 (traversal API) to add a reference view API at the HIR semantic layer, then uses it to power a bunch of LSP functionality.

#### HIR semantic layer (`hir::semantic::reference`):
  - ReferenceView abstraction for tracking symbolic references (paths, field accesses, method calls, use paths)
  - Target enum that unifies module-level scopes and local bindings
  - TargetResolution for handling ambiguous references gracefully (returns all candidates)
  - HasReferences trait + collector for finding all references to a target across a module or ingot
  - Proper resolution for use paths, including decomposed use foo::{bar, baz} statements and child file modules

#### LSP handlers (`language-server`):
  - goto-definition - expanded to handle path segments, where clauses, locals, ambiguous refs
  - find-all-references - ingot-wide search for both scopes and locals
  - document highlight
  - rename symbol (non-module)
  - go to type definition
  - hover - now works on local bindings, shows multiple candidates for ambiguous refs
  - semantic tokens (basic)

#### Fixes along the way:
  - Local bindings now properly shadow module-level items
  - Rename replaces just the identifier, not the whole path (Foo::Bar → Foo::Baz not just Baz)
  - Reference spans return the token not the whole expression
  - File module goto works (was using relative path instead of file:// URL)
  - Use path resolution starts from the correct scope

#### What's left for followup
  - Module renaming (needs to handle file renames)
  - Cross-ingot references (@g-r-a-n-t can we take a look at this together?)
  - Semantic tokens could use more granularity
  - Integration tests for the LSP handlers beyond snapshot tests

### DIFF
--- a/crates/language-server/src/functionality/references.rs
+++ b/crates/language-server/src/functionality/references.rs
@@ -18,7 +18,8 @@ fn find_references_at_cursor<'db>(
     cursor: Cursor,
 ) -> Vec<async_lsp::lsp_types::Location> {
     // Use the simplified API to get the target at cursor
-    let Some(target) = top_mod.target_at(db, cursor) else {
+    let resolution = top_mod.target_at(db, cursor);
+    let Some(target) = resolution.first() else {
         return vec![];
     };
 
@@ -34,7 +35,7 @@ fn find_references_at_cursor<'db>(
                     continue;
                 }
                 let mod_ = map_file_to_mod(db, file);
-                for ref_view in mod_.references_to_target(db, &target) {
+                for ref_view in mod_.references_to_target(db, target) {
                     if ref_view.span().resolve(db).is_some()
                         && let Ok(location) = to_lsp_location_from_lazy_span(db, ref_view.span())
                     {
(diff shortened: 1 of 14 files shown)


crates/language-server/src/functionality/references.rs
```diff
@@ -0,0 +1,384 @@
...
+    backend::Backend,
+    util::{to_lsp_location_from_lazy_span, to_lsp_location_from_scope, to_offset_from_position},
+};
+
+use super::goto::Cursor;
+
+/// Find an item definition at the cursor position.
+///
+/// Returns the target scope if the cursor is on an item's name token.
+fn find_item_definition_at_cursor<'db>(
```
micahscopes: unnecessary indirection

crates/hir/src/core/semantic/reference/mod.rs
```diff
@@ -0,0 +1,525 @@
...
+        let span = typed_body.expr_binding_def_span(func, expr_id)?;
+        let ty = typed_body.expr_ty(db, expr_id);
+
+        Some(Target::Local {
+            span,
+            ty,
+            func,
+            expr: expr_id,
+        })
+    }
```
sbillig: PathView as it's defined, and this function (and find_expr_for_path) don't suffice for variable names in function bodies, where scoping rules rule.

micahscopes: I'm working on this, still not satisfied though

micahscopes: should be addressed now, not in the most elegant way, but effectively

PR #1166, opened 2025-11-27
Support handling of complex enums

cburgdorf:
- Lowering and codegen now support enums with payloads: variant constructors allocate space, write discriminants plus fields, and match arms can bind and read variant data when switching
    on enums.
  - Block/state caches reuse Yul temps so bound variant fields stay in scope per match arm.
  - Core library gained `enum_repr.fe` helpers (store_discriminant, store_variant_field, get_variant_field) and re-exported them for contracts.
  - New fixtures and harness tests cover constructing and matching enums with data on EVM to prove end-to-end behavior.

  Example Fe code now handled:

```
  enum MyOption {
      None,
      Some(u64),
  }

  pub fn match_option(opt: MyOption) -> u64 {
      match opt {
          MyOption::Some(value) => value,
          MyOption::None => 0,
      }
  }
```

### DIFF
--- a/crates/codegen/src/yul/emitter/control_flow.rs
+++ b/crates/codegen/src/yul/emitter/control_flow.rs
@@ -327,37 +327,52 @@ impl<'db> FunctionEmitter<'db> {
 
         let mut default_body = None;
         for arm in &match_info.arms {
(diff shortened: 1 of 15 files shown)


sbillig (approved):
👍

PR #1167, opened 2025-11-28
Rework contract codegen (`[#contract_init(Foo)]` etc, and `code_region_offset/len` functions)

sbillig:
We now compile fe code like this:

```rust
#[contract_init(Foo)]
fn init() {
    let len = code_region_len(runtime)
    let offset = code_region_offset(runtime)
    codecopy(dest: 0, offset, len)
    return_data(0, len)
}

#[contract_runtime(Foo)]
fn runtime() {
  ...
}
```

into yul like this:
```
object "Foo" {
  code {
    init()
    function init() {
      let v0 := datasize("Foo_deployed")
      let v1 := dataoffset("Foo_deployed")
      codecopy(0, v1, v0)
      return(0, v0)
    }
  }

  object "Foo_deployed" {
    code {
      runtime()
      function runtime() {
        ...
```

I removed all the MIR/codegen handling of the `contract` object. For now, please pretend that `contract` doesn't exist. I'm working on the new contract object implementation, and it will desugar into something like the above.

Couple other changes:
- FuncSignature is its own syntax node now, mostly to improve parser code, but also for some possible diagnostic improvements.
- Extern functions can be generic now. This enables `code_region_offset`/`len`, and will enable future `fn size_of<T>()` etc

### DIFF
--- a/crates/codegen/src/yul/emitter/control_flow.rs
+++ b/crates/codegen/src/yul/emitter/control_flow.rs
@@ -143,7 +143,9 @@ impl<'db> FunctionEmitter<'db> {
         match terminator {
             Terminator::Return(Some(val)) => self.emit_return_with_value(*val, ctx.docs, ctx.state),
             Terminator::Return(None) => {
-                ctx.docs.push(YulDoc::line("ret := 0"));
+                if self.returns_value() {
+                    ctx.docs.push(YulDoc::line("ret := 0"));
+                }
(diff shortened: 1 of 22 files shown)


cburgdorf (approved):
Good stuff :+1: 
The copilot remarks about the return type of the added intrinsics might be valid concerns but I'll rage merge nonetheless.

PR #1168, opened 2025-11-28
 Use lang items to resolve MIR deps and expose attr value spans

cburgdorf:
This builds on top of #1166 

It's inspired by Rust: https://doc.rust-lang.org/beta/unstable-book/language-features/lang-items.html

  - add Hir lang item registry that prefers local/core defs for alloc/get_field/etc.
  - switch MIR lowering to lang-item lookups (incl. get_discriminant) instead of hardcoded paths
  - mark core helpers with #[lang = "..."] and update enum match snapshots
  - teach parser/HIR spans about name-value attrs and test #[lang = "alloc"]

### DIFF
--- a/crates/codegen/tests/fixtures/enum_variant_contract.snap
+++ b/crates/codegen/tests/fixtures/enum_variant_contract.snap
@@ -63,7 +63,7 @@ object "EnumContract" {
           store_variant_field__u256__3271ca15373d4483(v4, 0, 0, v3)
           let v5 := v4
           let v2 := 0
-          switch mload(v5)
+          switch get_discriminant(v5, 0)
             case 1 {
               let v6 := get_variant_field__u256__3271ca15373d4483(v5, 0, 0)
               v2 := v6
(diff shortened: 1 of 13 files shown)


sbillig:
I'm not sure I like this direction. I agree that the old code wasn't ideal, and maybe this is an improvement on that. But either way, we have a bunch of verbose logic in the compiler to look up library functions. Regarding the use of "lang" in particular: these library functions aren't at all an integral part of the language that the compiler must know about. The `Add` trait etc are, for example.

The direction I think I'd like to go in is to introduce a desugaring/lowering pass, in which we specify fe -> fe transformations. The lower-level fe would, eg, replace `foo.x` field access with `core::read_field<T: Target>(foo, 0)` (or we might be able to drop the `0` there and compute it at the type level, but that's a tangent). Each lowered version would still be valid fe code, and could/would be run through the analyzer, where library functions would be resolved by name.

Of course, this idea is just in my head for now, and I don't want to block progress. That said, I would strongly prefer that we just add something like a utility function or macro that let's you look up library functions by name (eg just collapse all of the old copypasta `resolve_store_discriminant_via_path` etc functions into one utility function that takes a path), and avoid these explicit lists of library functions that we're gonna use, and all the lang item table machinery, etc.

PR #1171, opened 2025-12-03
Split MIR lowering into cohesive modules

cburgdorf:
There were a few mini refactorings in the first commit but the second commit is a pure code split. It refactors mir/src/lower into focused submodules (expr, prepass, match, intrinsics, variants, aggregates, contract) with a thin mod.rs entrypoint and builder scaffold.

I wanted to get this done before adding more features to MIR and to make it easier for others to contribute on the MIR code.


sbillig (approved):
👍

PR #1173, opened 2025-12-03
Handle storage usage in codegen

cburgdorf:
This PR adds support for storage access via the effect system, allowing contract state to be read and written through effect parameters.

  ## Key changes:

  - Field assignments: Support assigning to struct fields (e.g., p.x = value)
  - Enum discriminant handling: Always use get_discriminant for enum matching, even for payloadless enums, ensuring consistent codegen
  - Terminating match arms: Properly handle match arms that return early, avoiding incorrect control flow after the match
  - Storage effects: Track address spaces (memory vs storage) through MIR lowering and emit correct mload/sload and mstore/sstore operations based on whether values come from effect parameters

  Contract entrypoints with effect parameters now automatically get storage pointers synthesized in the function prologue, while regular functions receive effect bindings as additional
  parameters.
  
  There's a comprehensive storage demo that is deployed and tested against.
  
https://github.com/argotorg/fe/blob/f9483e4c2719eda0dc010407a4df8643b088013c/crates/codegen/tests/fixtures/storage.fe#L95-L130
  
  ## Caveats

  The integration with the current uses effect system is intentionally minimal and temporary. This implementation is meant to evolve incrementally alongside @sbillig  's ongoing work on the effect system. As that design matures, the storage access mechanism will be revised to align with the final effect semantics.

### DIFF
--- a/crates/codegen/src/yul/emitter/expr.rs
+++ b/crates/codegen/src/yul/emitter/expr.rs
@@ -1,10 +1,13 @@
 //! Expression and value lowering helpers shared across the Yul emitter.
 
 use hir::hir_def::{
-    CallableDef, Expr, ExprId, LitKind, Stmt, StmtId,
+    Attr, CallableDef, Expr, ExprId, Func, ItemKind, LitKind, Stmt, StmtId,
     expr::{ArithBinOp, BinOp, CompBinOp, LogicalBinOp, UnOp},
 };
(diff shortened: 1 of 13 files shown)


sbillig (approved):
👍

When we desugar contract code, we'll generate explicit storage pointers, and we can drop the assumption that all effects are storage effects (which obviously won't be the case).

PR #1174, opened 2025-12-05
decision tree codegen and projection abstraction

micahscopes:
This PR introduces a projection abstraction for navigating into data structures. Field access and pattern matching now work via semantic paths rather than scattered offset calculations.

  ### What changed

  The projection system flows through the whole pipeline:
  - **HIR** has a generic `Projection<Ty, Var, Idx>` vocabulary used by the decision tree
  - **MIR** uses `Place { base, projection }` for both field access and match lowering
  - **Codegen** walks projections through one centralized pathway (`lower_place_address`) to emit pointer arithmetic

This replaces the old approach of calling `GetField`/`GetVariantField` helpers everywhere, which made it hard to reason about layout correctness.

  ### Changes / To-do

  - [x] Generic projection module (`hir/src/projection.rs`) with `Projection<Ty, Var, Idx>`
  - [x] HIR uses `Idx = Infallible` so pattern matching can't construct dynamic indexing
  - [x] Decision tree --> nested MIR switches
  - [x] MIR `Place { base, projection, address_space }` with `PlaceLoad`/`PlaceRef`
  - [x] Field access uses `Place` abstraction
  - [x] Single codegen choke point (`lower_place_address`)
  - [x] Layout centralized in HIR with snapshot tests
  - [x] Shared match temp (one `let tmp := 0` for nested switches)
  - [x] Direct memory loads with masking/signextend
  - [x] Nested enum bindings work end-to-end (`match_nested_enum.fe`)
  - [x] Fix off-by-one in `field_offset_bytes` / `variant_field_offset_bytes`
  - [x] Make "unknown size --> 32 bytes" explicit in layout API instead of `unwrap_or(32)` in codegen
  - [x] MIR should route default --> wildcard leaf explicitly (codegen shouldn't rediscover it)
  - [x] Add fixtures for wildcard + nested defaults

  ### Future directions

  **Near-term**
  - `storage_size_bytes` / `field_storage_size_bytes` in layout API
    - make the "unknown types are 32-byte words" rule explicit rather than scattered `unwrap_or(32)`

  **Medium-term**
  - `Index`/`Deref` projection variants (for arrays, references)
  - `may_alias()` on projection paths, needed for SROA, escape analysis
  - Span tracking for editability (i.e. IDE refactoring support)
  - `Editable`/`EditablePath` traits (stubbed out, not yet implemented)
  - HIR semantic API: `FieldAccessView::as_projection()`
    - unify field access tracking with projection vocabulary
    - investigate consolidating with broader HIR semantic traversal API
    - projection abstraction could be super useful for e.g. the doc engine / language server

  **Long-term**
  - On-demand graph view of MIR for optimization passes (e.g. SROA, escape analysis)
  - Target-agnostic MIR
    - remove `AddressSpaceKind` from `Place`, add explicit `Alloca`
    - let codegen decide memory vs storage based on target
  - Producing sea of nodes and other similar graph representations i.e. projections as graph edges rather than address-computation instructions

### DIFF
--- a/crates/hir/src/projection.rs
+++ b/crates/hir/src/projection.rs
@@ -11,8 +11,6 @@
 //!   other type representation via the `Ty`, `Var`, and `Idx` type parameters.
 //! - **Graph IR compatible**: Can be used as instruction operands (CFG+SSA) or
 //!   as edge labels (Sea of Nodes) with the same vocabulary.
-//! - **Editability-ready**: Designed to support future refactoring features via
-//!   the `Editable` trait (interface defined, implementation deferred).
 
 use smallvec::SmallVec;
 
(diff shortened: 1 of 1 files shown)


sbillig:
### Code review

Found 1 issue:

1. Missing `insert_place_expr` call in default arm binding emission. The regular switch cases (lines 345-350) cache the place expression with `insert_place_expr`, but the default arm (lines 402-406) does not include this caching. This inconsistency means bindings in default arms won't have their place expressions cached, which could break future reference-semantics codegen for default arm bindings.

https://github.com/argotorg/fe/blob/b6f676385336fe42832ab2dc82a2d969c0fb31db/crates/codegen/src/yul/emitter/control_flow.rs#L401-L407

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sbillig:
More input from Claude:

  When computing the byte offset of a field in a struct/tuple, there are two approaches in the codebase:

  1. Centralized API (field_offset_bytes_or_word_aligned): When ANY preceding field has an unknown size (like an enum), it falls back to word-aligned layout: field_idx * 32 bytes. This ensures all fields start at word boundaries.
  2. Manual calculation in lower_place_address(): Sums individual field sizes using ty_size_bytes_or_word(), which returns 32 bytes per unknown field.

  Why this matters:

  Consider a struct { a: MyEnum, b: u8 } where MyEnum has unknown size:
                                                                                                                                   | Approach           | Field a offset | Field b offset     |
  |--------------------|----------------|--------------------|
  | Centralized API    | 0              | 32 (1 * 32)        |
  | Manual calculation | 0              | 32 (sum of a's 32) |

  These happen to match here. But for { a: u8, b: MyEnum, c: u16 }:

  | Approach           | Field c offset                                                      |
  |--------------------|---------------------------------------------------------------------|
  | Centralized API    | 64 (2 * 32, because field b is unknown)                             |
  | Manual calculation | 1 + 32 = 33 (sum of actual sizes with 32-byte fallback per unknown) |

PR #1177, opened 2025-12-09
Support MethodCall in if/else branches

Turupawn:
* `MethodCall` is now supported in if/else branches as returned vales 
* Added a `lower_and_format_call` function to encapsulate repeated code

### DIFF
--- a/crates/codegen/src/yul/emitter/expr.rs
+++ b/crates/codegen/src/yul/emitter/expr.rs
@@ -200,11 +200,6 @@ impl<'db> FunctionEmitter<'db> {
                     lowered_args.push(self.lower_expr(arg.expr, state)?);
                 }
 
-                if let CallableDef::Func(func_def) = callable.callable_def {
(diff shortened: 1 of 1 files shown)


cburgdorf:
Can you add a test file to crates/codegen/tests/fixtures that demonstrates the code that now passes but before didn't?

crates/codegen/tests/fixtures/if_else_methodcall.snap
```diff
@@ -0,0 +1,37 @@
...
+}
+function teststruct_get_value2(self, x) -> ret {
+  ret := mul(x, 24)
+}
+function test_if_else(x) -> ret {
+  let v0 := alloc(0)
+  let v1 := v0
+  let v2 := 0
+  if gt(x, 10) {
+    v2 := get_value1(v1)
```
cburgdorf: It looks like the name of the functions being called `get_value1` does not match the actual name of the function `teststruct_get_value1`

sbillig:
Closing as defunct

PR #1183, opened 2025-12-12
contract lowering

sbillig:


crates/codegen/src/yul/emitter/function.rs
```diff
@@ -299,10 +296,59 @@ impl<'db> FunctionEmitter<'db> {
...
-        layout::ty_size_bytes(self.db, ty).ok_or_else(|| {
+        self.ty_storage_slots(ty).ok_or_else(|| {
             YulError::Unsupported(format!(
                 "cannot determine storage size for effect `{binding_name}`: unsupported type"
             ))
         })
     }
+
+    /// Best-effort slot size computation for types in storage.
+    fn ty_storage_slots(&self, ty: TyId<'db>) -> Option<u64> {
```
cburgdorf: I think this method should handle `enum` and probably `tuple` types as well (even though tuples will be rewritten as structs so I'm not entirely sure if it is needed here)

crates/codegen/tests/fixtures/effect_ptr_domains.snap
```diff
@@ -0,0 +1,57 @@
+---
+source: crates/codegen/tests/yul.rs
+expression: output
+input_file: tests/fixtures/effect_ptr_domains.fe
+---
+function bump(foo) {
```
cburgdorf, line 6: I believe this should be named `bump_eff_stor` .

cburgdorf (approved):
Looks great. I left two comments inline.

PR #1199, opened 2025-12-31
mir: newtype peeling

sbillig:
This (recursively) replaces single-field structs with their field, while retaining the original type information from the. Result is many fewer memory allocations, and free "newtype" wrapper structs like `struct Wei { inner: u256 }`.

### DIFF
--- a/crates/codegen/tests/fixtures/create_contract.snap
+++ b/crates/codegen/tests/fixtures/create_contract.snap
@@ -1,6 +1,5 @@
 ---
 source: crates/codegen/tests/yul.rs
-assertion_line: 42
 expression: output
 input_file: tests/fixtures/create_contract.fe
 ---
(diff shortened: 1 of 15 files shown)


crates/codegen/tests/fixtures/erc20.desugared.snap
```diff
@@ -295,12 +300,12 @@ fn spend_allowance(owner: Address, spender: Address, amount: u256) uses (mut sto
 
 struct TokenStore {
     total_supply: u256,
-    balances: StorageMap<Address, u256>,
-    allowances: StorageMap<(Address, Address), u256>,
+    balances: StorageMap<Address, u256, BALANCES_SALT>,
```
cburgdorf, line 315: I assume having to supply these salts manually is meant as a temporary solution.

cburgdorf (approved):
Looks good. I assume the manual salt for the maps is just a temporary solution though.

crates/codegen/tests/fixtures/erc20.desugared.snap
sbillig: Yeah, we can improve this. The previous scheme wasn't actually working.

Issue #1207, opened 2026-01-08
support trait-based `for` loop iteration


sbillig:
Simple proposal that will suffice for now (ignoring `mut`):
(from zulip https://fe-lang.zulipchat.com/#narrow/channel/458688-fe-compiler-dev/topic/for-loop.20iteration/with/562861650)

```rust
// in core lib
trait Seq<T> {
    fn len(self) -> usize;
    fn get(self, i: usize) -> T; // or just use the core::ops::Index trait for this
}
```

```rust
struct CalldataView<T>
where
  T: Decode,
{
  offset: usize,
  len: usize,
}
impl<T> Seq<T> for CalldataView<T>
where T: Decode
{
  fn len(self) -> usize { self.len }
  fn get(self, i: usize) -> T {
    T::decode_from(ptr: CalldataPtr<T>::new(self.offset + i))
  }
}

fn example() -> u64 {
  let nums = CalldataView<u64>::at_offset(4)
  let mut sum = 0
  for n in nums {
    sum += n
  }
  sum
}
// desugars to:
let len = nums.len() // or Seq::len(nums)
let mut i = 0
while i < len {
  sum += nums.get(i) // or Seq::get(nums, i)
  i += 1
}
```

PR #1213, opened 2026-01-13
Seq iteration

micahscopes:
Implements generic iteration via a `Seq<T>` trait, following Sean's proposal in #1207. Also adds `..` range operator syntax and fixes const generics in impl target types.

  ### Changes

  #### Parser/HIR
  - **Range operator** - adds `..` as `ArithBinOp::Range`, parsed and lowered through HIR
  - Formatter and codegen adjusted to print/emit the new expression form

  #### Core library (`library/core/`)
  - **`Seq<T>` trait** - `fn len(self) -> usize` and `fn get(self, i: usize) -> T`
  - **`Range` / `DynRange`** - range types with `Seq` implementations
  - **Array `Seq` impl** - `impl<T, const N: usize> Seq<T> for [T; N]`

  #### Type checking (`hir/analysis/ty/`)
  - **For-loop Seq resolution** - loops typed via `Seq` trait lookup instead of hardcoded array/range cases
  - **`ForLoopSeq`** - stores pre-resolved `Callable` for `len`/`get` so MIR can emit direct calls
  - **Inference hygiene** - snapshot/rollback when probing candidate impls to prevent pollution
  - **Effect preservation** - resolved effect args carried through synthesized method calls

  #### MIR lowering
  - **Seq-based desugaring** - `for x in seq` becomes index loop using resolved `Seq::len`/`Seq::get`
  - **Continue correctness** - `continue` now targets increment block, not condition (fixes infinite loop bug)
  - **`CallOrigin.expr`** - now `Option<ExprId>` to represent synthesized calls without fake expr IDs
  - Fallback paths remain for arrays/ranges when Seq resolution unavailable

  #### Const generics fix
  - **Impl target scoping** - `Impl::lower_ast` and `ImplTrait::lower_ast` now enter preliminary scope first, lower generics, *then* lower target type
  - Enables `impl<T, const N: usize> ... for [T; N]` patterns that previously failed resolution

  ### Tests
  - `const_generic_impl_target_array.fe` / `const_generic_impl_regression.fe` - uitest for const generic fix
  - `for_continue.fe` / `for_continue.mir.snap` - MIR snapshot for continue-in-for correctness

### Questions / follow-up
- how to combine with #1211?
- `Range` vs `DynRange`?  Do we need this distinction?

### DIFF
--- a/library/core/src/range.fe
+++ b/library/core/src/range.fe
@@ -36,7 +36,11 @@ pub struct Range<S: Bound, E: Bound> {
 /// Fully const range: `Known<S>..Known<E>`
 impl<const S: usize, const E: usize> Seq<usize> for Range<Known<S>, Known<E>> {
     fn len(self) -> usize {
-        E - S
+        if E < S {
+            0
+        } else {
+            E - S
+        }
     }
 
     fn get(self, i: usize) -> usize {
@@ -47,7 +51,11 @@ impl<const S: usize, const E: usize> Seq<usize> for Range<Known<S>, Known<E>> {
 /// Const start, dynamic end: `Known<S>..n` (common `0..n` pattern)
 impl<const S: usize> Seq<usize> for Range<Known<S>, Unknown> {
     fn len(self) -> usize {
-        self.end - S
+        if self.end < S {
+            0
+        } else {
+            self.end - S
+        }
     }
 
     fn get(self, i: usize) -> usize {
@@ -58,7 +66,11 @@ impl<const S: usize> Seq<usize> for Range<Known<S>, Unknown> {
 /// Dynamic start, const end: `n..Known<E>`
 impl<const E: usize> Seq<usize> for Range<Unknown, Known<E>> {
     fn len(self) -> usize {
-        E - self.start
+        if E < self.start {
+            0
+        } else {
+            E - self.start
+        }
     }
 
     fn get(self, i: usize) -> usize {
@@ -69,7 +81,11 @@ impl<const E: usize> Seq<usize> for Range<Unknown, Known<E>> {
 /// Fully dynamic range: `start..end`
 impl Seq<usize> for Range<Unknown, Unknown> {
     fn len(self) -> usize {
-        self.end - self.start
+        if self.end < self.start {
+            0
+        } else {
+            self.end - self.start
+        }
     }
 
     fn get(self, i: usize) -> usize {
--- a/crates/codegen/tests/fixtures/range_bounds.snap
+++ b/crates/codegen/tests/fixtures/range_bounds.snap
@@ -75,7 +75,18 @@ function $sum_dynamic($start, $end) -> ret {
   leave
 }
 function $range_known_const_s__usize___known_const_e__usize___seq_len__0_4__f9605efabd18106c($self) -> ret {
-  ret := sub(4, 0)
+  let v0 := 0
+  switch lt(4, 0)
+    case 1 {
+      v0 := 0
+    }
+    case 0 {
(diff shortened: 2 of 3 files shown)


library/core/src/range.fe
```diff
@@ -0,0 +1,22 @@
...
+pub struct Range<const N: usize> {
+    pub start: usize,
+}
+
+/// A range with runtime-determined length: `start..end`
+pub struct DynRange {
+    pub start: usize,
+    pub end: usize,
+}
+
```
sbillig, line 23: It's probably useful to have a type-level Range, but why not `Range<const Start: usize, End: usize>`? Type-level Range will introduce more monomorphization code bloat (one instance of each method for each (Start, End) pair), but with inlining it'll boil down to tiny code anyway.

sbillig: Or maybe `Range<const Start: usize> { end: usize }`, because a common case is `0..n`, where n is not statically known.

Pleasant side-effect of using static Range with size <= 1 word where possible is that in our current level of compiler sophistication (low), DynRange will be stored in memory, and each loop iteration will involve a lot of mload overhead.

micahscopes:
@sbillig @Turupawn ah wait, just kidding, a few more things to do.  I'll let you know when it's ready to look at.

micahscopes:
Okay, better now! c.c. @Turupawn @sbillig

crates/mir/src/lower/expr.rs
```diff
@@ -1428,6 +1499,696 @@ impl<'db, 'a> MirBuilder<'db, 'a> {
...
+                self.assign(None, None, Rvalue::Load { place });
+                start_val
+            } else {
+                // Fallback: just use 0
+                self.synthetic_u256(BigUint::from(0u64))
+            };
+
+        // Extract the end value from the range
+        let end_value =
+            if let Some(info) = self.field_access_info(range_ty, FieldIndex::Ident(end_ident)) {
```
sbillig: Better to unwrap here and remove the fallback. Same goes for the 'start' case above.

library/core/src/range.fe
```diff
@@ -0,0 +1,78 @@
...
+pub struct Range<S: Bound, E: Bound> {
+    pub start: <S as Bound>::Repr,
+    pub end: <E as Bound>::Repr,
+}
+
+/// Fully const range: `Known<S>..Known<E>`
+impl<const S: usize, const E: usize> Seq<usize> for Range<Known<S>, Known<E>> {
+    fn len(self) -> usize {
+        E - S
+    }
```
sbillig, line 44: This should use `saturating_sub` (though sadly that doesn't exist yet), or we should add a hir check to ensure E >= S at least for the sugar syntax. (or we just make an issue to modify this when saturating_sub is available).

crates/mir/src/lower/expr.rs
```diff
@@ -1428,6 +1499,696 @@ impl<'db, 'a> MirBuilder<'db, 'a> {
...
+            });
+            let elem_value = self.emit_seq_get_call(
+                iterable_value,
+                idx_for_get,
+                elem_ty,
+                &seq_info.get_callable,
+                &seq_info.get_effect_args,
+            );
+
+            self.assign(None, Some(elem_local), Rvalue::Value(elem_value));
```
sbillig: This doesn't work for pattern bindings, eg `for Point { x, y } in points`.

I think all you need to do is:
```rust
// remove the `if let Some(elem_local)` guard
let idx_for_get = ...
let elem_value = ...
self.bind_pat_value(pat, elem_value)
// carry on with the `// Execute the body` code below.
```

PR #1215, opened 2026-01-16
Implement proper type casting

cburgdorf:
- Implements first-class as casting for Fe with clearly defined precedence/associativity and typechecking rules that only allow lossless casts (including lossless `uN` -> `iM` when `M >= N+1`).
  - Adds transparent “newtype” wrapper casting (single-field struct / single-element tuple) so wrappers can cast to/from their underlying integer type.
  - Introduces a standard integer downcasting API with checked, truncating, saturating, and unchecked variants (unchecked included for performance-critical, provably-safe use cases even without unsafe today).
  - Removes the legacy name-based cast shim behavior now that proper casting is supported end-to-end, and updates lowering/codegen to treat casts as representation-preserving operations.
  - Migrates core/std usage to the new casting system; rewrites many `downcast_unchecked()` sites to `downcast_truncate()` where behavior and runtime cost are equivalent.
  - Updates diagnostics to direct users toward downcast APIs for narrowing cases.

### DIFF
--- a/crates/uitest/fixtures/ty_check/cast_invalid.snap
+++ b/crates/uitest/fixtures/ty_check/cast_invalid.snap
@@ -4,15 +4,6 @@ assertion_line: 27
 expression: diags
 input_file: fixtures/ty_check/cast_invalid.fe
 ---
-error[8-0031]: type annotation is needed
-   ┌─ cast_invalid.fe:14:5
-   │
-14 │     256 as u8
-   │     ^^^
-   │     │
-   │     type annotation is needed
-   │     no default type is provided for an integer type. consider giving integer type
-
 error[8-0055]: cast is not provably lossless
   ┌─ cast_invalid.fe:2:5
   │
--- a/crates/hir/src/analysis/ty/ty_check/expr.rs
+++ b/crates/hir/src/analysis/ty/ty_check/expr.rs
@@ -293,10 +293,16 @@ impl<'db> TyChecker<'db> {
         {
             let value = int_id.data(self.db);
             if self.int_literal_fits_in_ty(value, to) {
+                // Unify the literal's type variable with the target leaf type
+                // so it doesn't remain unresolved.
+                let leaf = self.peel_transparent_newtypes(to);
+                let _ = self.table.unify(from, leaf);
                 return ExprProp::new(to, true);
             }
 
             let leaf = self.peel_transparent_newtypes(to);
+            // Unify to prevent a spurious "type annotation needed" error.
+            let _ = self.table.unify(from, leaf);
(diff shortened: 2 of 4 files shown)


crates/codegen/tests/fixtures/name_collisions.snap
```diff
@@ -76,7 +76,7 @@ function $__u32_as_u256($x) -> ret {
   leave
 }
 function $test_cast_shim_collision() -> ret {
-  let v0 := 42
+  let v0 := $__u32_as_u256(42)
```
cburgdorf, line 79: This was the remaining name collision that was implicitly fixed by these change.

crates/uitest/fixtures/ty_check/cast_invalid.snap
```diff
@@ -0,0 +1,46 @@
...
+input_file: fixtures/ty_check/cast_invalid.fe
+---
+error[8-0031]: type annotation is needed
+   ┌─ cast_invalid.fe:14:5
+   │
+14 │     256 as u8
+   │     ^^^
+   │     │
+   │     type annotation is needed
+   │     no default type is provided for an integer type. consider giving integer type
```
sbillig: Ty checker could probably be made to allow this case.

sbillig: Err, I mean the case `255 as u8`, but maybe that already works?

library/core/src/num.fe
```diff
@@ -24,7 +25,7 @@ extern {
 
     fn __bitnot_u8(_: u8) -> u8
 
-    pub fn __u8_as_u256(_: u8) -> u256
+    fn __u8_as_u256(_: u8) -> u256
```
sbillig: Do we still need these functions? Seems like they can be deleted.

sbillig (approved):
👍

library/core/src/num.fe
cburgdorf: I just  [pushed a commit](https://github.com/argotorg/fe/pull/1215/commits/8668bcedab7539102c2303d24aabe5e309232772) to collapse all these 30 functions into a single generic `__bitcast<From, To>(value: From) -> To`.

That one is still needed. The public casting API needs to convert any integer to/from the EVM's 256-bit word, but the type system forbids these conversions via `as`. ` __bitcast` bridges this gap as the sole intrinsic the compiler recognizes for unchecked bit reinterpretation.

crates/uitest/fixtures/ty_check/cast_invalid.snap
cburgdorf: Yep, literals can now be casted if they fit :+1:

PR #1218, opened 2026-01-20
Sonatina backend: end-to-end contract compilation and test infrastructure

micahscopes:
Adds a second compilation backend that lowers Fe source to EVM bytecode through Sonatina IR. All 21 contract test fixtures (erc20, escrow, vault, AMM, etc.) compile and pass on both backends.

#### Sonatina codegen (`crates/codegen/src/sonatina/`)
- Full MIR-to-Sonatina IR lowering: functions, control flow, memory, storage, calls, intrinsics
- Native struct types with GEP-based field addressing
- Contract object emission with init/runtime sections
- Sub-word masking, address space routing, projection handling

#### Backend selection (`--backend sonatina` or `--backend yul`)
- `Backend` trait abstracts over codegen pipelines
- Optimization levels (`--opt-level 0/1/2`) mapped to each backend's optimizer

#### Test and report tooling
- `fe test --report` captures sources, IR, bytecode, and EVM traces as `.tar.gz`
- Gas comparison between backends with opcode breakdowns
- `test_fe_test_both_backends` runs every fixture on both backends and asserts they agree

#### Testing
- [x] Sonatina IR snapshot tests
- [x] "both-backends" integration tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1936,7 +1936,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3010,7 +3010,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3345,7 +3345,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


crates/codegen/src/sonatina/mod.rs
```diff
@@ -0,0 +1,2148 @@
...
+    ///
+    /// Objects define how code is organized for compilation. Each object
+    /// has sections (like "runtime" and "init") that contain function entries.
+    ///
+    /// For contract runtime entrypoints: The entry function emits `evm_stop` instead of
+    /// internal `Return` since contract dispatchers handle return data via `evm_return`.
+    ///
+    /// For simple test files (no explicit contract): A wrapper function calls the entry
+    /// and then does `evm_stop`, preserving internal function call semantics.
+    fn create_objects(&mut self) -> Result<(), LowerError> {
```
sbillig, line 424: - all functions reachable via calls from `%entry` will be automatically included. `%include` is for including functions that don't appear in the call graph
- to find the contract runtime and ini functions, you should look for the `contract_init(Foo)`/`contract_runtime(Foo)` attributes (or whatever they are). Look elsewhere in MIR/codegen for how these are found, and share code.
- contract init objects aren't currently being created

crates/codegen/Cargo.toml
```diff
@@ -9,7 +9,11 @@ mir = { path = "../mir", package = "fe-mir" }
 hir.workspace = true
 common.workspace = true
 hex = "0.4"
+num-bigint.workspace = true
 rustc-hash.workspace = true
+sonatina-ir = { path = "../../../sonatina/crates/ir" }
+sonatina-triple = { path = "../../../sonatina/crates/triple" }
+sonatina-codegen = { path = "../../../sonatina/crates/codegen" }
```
sbillig: Point to github plz

PR #1223, opened 2026-01-28
Read-only embedded ingot files written to temp for language server navigation

g-r-a-n-t:


g-r-a-n-t:
> Did you look into any alternative LSP features for displaying a file from the LSP server? In the past I thought I saw something like this, but it being a fairly new and not-yet-supported feature.

I looked into it a bit at first and got the impression that that approach would be more complicated due to each editor handling virtual files differently (or not at all). In any case, this behavior is a good fallback and maybe we can add virtual file support one editor at a time.

micahscopes (approved):
Looks great.

Yeah, looks like `workspace/getReferenceDocument` will come in 3.18.  We should target that eventually but not in the near term, I totally agree.

The main followup that could be cool in the nearer term would be to generalize this machinery to be used for displaying Yul/Sonatina IR and maybe even representations of bytecode.

PR #1236, opened 2026-02-04
mut/ref/own and borrow checking

sbillig:


crates/common/src/diagnostics.rs
```diff
@@ -171,6 +173,7 @@ impl DiagnosticPass {
             Self::TraitSatisfaction => 6,
             Self::MethodDefinition => 7,
             Self::TyCheck => 8,
+            Self::Mir => 10,
```
micahscopes: `EventLower` is also 10, should this be 11?

crates/uitest/fixtures/mir_check/use_after_move.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/uitest/tests/mir_check.rs
+expression: res
+input_file: fixtures/mir_check/use_after_move.fe
+---
+error[8-0000]: type mismatch
+  ┌─ use_after_move.fe:6:5
+  │
+6 │     value.value
+  │     ^^^^^^^^^^^ expected `()`, but `u256` is given
```
micahscopes: should this be `let _ = value.value`? to trigger the correct error?

crates/uitest/fixtures/mir_check/move_conflict_borrow.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/uitest/tests/mir_check.rs
+expression: res
+input_file: fixtures/mir_check/move_conflict_borrow.fe
+---
+error[8-0000]: type mismatch
+  ┌─ move_conflict_borrow.fe:6:5
+  │
+6 │     value.value
+  │     ^^^^^^^^^^^ expected `()`, but `u256` is given
```
micahscopes: Seems like this type error is masking the desired error here as well.

Cargo.toml
```diff
@@ -51,5 +51,36 @@ sonatina-verifier = { git = "https://github.com/fe-lang/sonatina", rev = "e6d85a
 # Set to 0 to make the build faster and debugging more difficult.
 debug = 0
 
-[profile.test.package.sonatina-codegen]
-opt-level = 3
+[profile.dev.package]
+revm = { opt-level = 3 }
```
micahscopes, line 56: you can also increase the number of parallel jobs for `fe test`, right now I think I set it to 4 because 32 was destroying my computer

PR #1243, opened 2026-02-08
feat(fe-cli): generate shell completions 

vuvoth:
# Description 
-   Generate shell completions for the `fe` cli in.
```bash
./target/debug/fe completion <shell>` # bash, elvish, fish, powershell, zsh
```

You should add the generated script above to the shell completion folders to make it work perfectly. 
Example in oh my zsh: 
```
./target/debug/fe completion zsh > ~/.oh-my-zsh/completions/_fe
```
Then you can use \<Tab\> to complete the commands. 

For other shells, please check the instructions on the internet :D.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1071,6 +1071,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.5.65"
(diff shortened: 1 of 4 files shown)


sbillig (approved):
Nice, thanks @vuvoth!

PR #1247, opened 2026-02-11
Make fe new use simple counter example

cburgdorf:
This makes `fe new` create a simple `Counter` example accompanied with a test that the user can run.


sbillig (approved):
👍

PR #1261, opened 2026-02-18
speedup tests

sbillig:


micahscopes (approved):
looks good, the fixture consolidation and binary caching are nice wins. I built on top of this on the test-speedup branch with the mock test fixes.

PR #1267, opened 2026-02-19
Revive release pipeline

cburgdorf:
Summary
                                                                                                                                                                                                                    
  - Switch default backend to Sonatina: Fe no longer depends on solc for code generation. The Sonatina backend (direct EVM bytecode) is now the default for both fe check and fe test.
  - Expand release matrix to 5 targets: Linux (amd64, arm64), macOS (arm64, amd64), and Windows (amd64). Each target uses proper cross-compilation with explicit --target flags.
  - Modernize release workflow: Update softprops/action-gh-release from v1 to v2, add required permissions: contents: write, remove unused libboost-all-dev dependency, and add vendored-openssl feature chain for
  cross-compilation targets that can't link system OpenSSL.

  - Verified release workflow produces all 5 binaries with tag [v26.0.0-alpha.3](https://github.com/argotorg/fe/releases/tag/v26.0.0-alpha.3)
  
  Known issues:
  
  - Sonatina backend is currently unable to compile our `fe new` example
  
  
<img width="1333" height="641" alt="image" src="https://github.com/user-attachments/assets/3b97376a-ce88-4b85-90a6-1bd7f07afdfc" />

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2071,7 +2071,7 @@ dependencies = [
 
 [[package]]
 name = "fe"
-version = "0.26.0"
+version = "26.0.0-alpha.5"
 dependencies = [
  "camino",
  "clap",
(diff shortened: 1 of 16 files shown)


sbillig:
#1268

PR #1269, opened 2026-02-20
`alloc` intrinsic

g-r-a-n-t:
This PR makes `std::evm::mem::alloc` a compiler intrinsic instead of a normal stdlib function, so each backend can lower allocation in a backend-safe way.

It includes the following changes:

  - `std::evm::mem::alloc` is now `extern` and recognized as an intrinsic during MIR lowering.
  - Added MIR opcode `IntrinsicOp::Alloc` and plumbing (formatting + intrinsic mapping).
  - Contract init lowering now emits the `alloc` intrinsic directly (avoids generating calls to an `extern` function).
  - Codegen lowering:
      - Yul backend emits the bump-pointer/free-memory-pointer sequence inline (no `$alloc` helper).
      - Sonatina backend lowers to `evm_malloc` (avoids relying on Solidity’s `0x40` free-memory scheme).
  - Fixed a subtle Yul miscompile for self-assign patterns like `x = alloc(x)` by caching the size in a temp before overwriting the destination.
  - Updated Yul + Sonatina IR snapshots and added a regression fixture for the self-assign case.

### DIFF
--- a/crates/codegen/src/yul/emitter/statements.rs
+++ b/crates/codegen/src/yul/emitter/statements.rs
@@ -550,11 +550,7 @@ impl<'db> FunctionEmitter<'db> {
         }
         docs.push(YulDoc::block(
             format!("if iszero({ptr}) "),
-            vec![YulDoc::line(format!("{ptr} := msize()"))],
-        ));
-        docs.push(YulDoc::block(
-            format!("if lt({ptr}, 8192) "),
-            vec![YulDoc::line(format!("{ptr} := 8192"))],
+            vec![YulDoc::line(format!("{ptr} := 0x80"))],
         ));
         docs.push(YulDoc::line(format!("mstore(64, add({ptr}, {size}))")));
         Ok(())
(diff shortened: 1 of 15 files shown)


crates/codegen/src/yul/emitter/statements.rs
```diff
@@ -515,6 +519,47 @@ impl<'db> FunctionEmitter<'db> {
...
+            docs.push(YulDoc::line(format!("{ptr} := mload(64)")));
+        }
+        docs.push(YulDoc::block(
+            format!("if iszero({ptr}) "),
+            vec![YulDoc::line(format!("{ptr} := msize()"))],
+        ));
+        docs.push(YulDoc::block(
+            format!("if lt({ptr}, 8192) "),
+            vec![YulDoc::line(format!("{ptr} := 8192"))],
+        ));
```
sbillig, line 554: This should just be `if iszero({ptr}) {ptr} := 0x80`

I think the msize and 8192 things were unnoticed ai workarounds for sonatina's non-sol memory model.

PR #1272, opened 2026-02-21
sona alloca

sbillig:
We now emit typed alloca instructions for non-escaping allocations, based on MIR escape analysis. This opens up more sonatina optimizations.

This branch also targets the latest changes in sonatina, with some big gas savings.


sbillig:
@codex review

PR #1290, opened 2026-02-26
Fuzz fix

g-r-a-n-t:
- Diagnostics no longer assume every diagnostic has a span: `CompleteDiagnostic::primary_span` now returns `Option<Span>` (`crates/common/src/diagnostics.rs:33`), and all diagnostic sorters/LSP conversion were updated to handle `None` (`crates/driver/src/db.rs:136`, `crates/hir/src/analysis/diagnostics.rs:268`, `crates/language-server/src/util.rs:119`, `crates/language-server/src/lsp_diagnostics.rs:37`, plus test harnesses like `crates/hir/src/test_db.rs:259` and `crates/hir/tests/pattern_matching.rs:33`).
- HIR printing/typecheck made robust to malformed/partial input found by fuzzing: `pretty_print_signature` uses placeholders instead of panicking (`crates/hir/src/core/print.rs:1045`), and unnamed params now get a `_` ident instead of unwrap-panic (`crates/hir/src/analysis/ty/ty_check/env.rs:1356`).
- MIR lowering/monomorphization hardened to report errors instead of panicking/returning incomplete MIR:
  - Unary `ref`/`mut` lowering no longer panics if the operand can’t be treated as a place (`crates/mir/src/lower/expr.rs:294`).
  - Monomorphizer now defers and propagates instantiation / trait-resolution failures as `MirLowerError` (instead of `panic!`), including the “empty worklist but deferred error exists” edge case (`crates/mir/src/monomorphize.rs:340`, `crates/mir/src/monomorphize.rs:1280`).
  - Effect-arg lowering no longer silently defaults to `0` (removes miscompile); MIR panics on violated HIR invariants and HIR now rejects invalid `(pass_mode, arg)` combos up front (`crates/mir/src/lower/expr.rs:1263`, `crates/hir/src/analysis/ty/ty_check/expr.rs:1301`). Also adds a fallback effect-local name `effect{idx}` (`crates/mir/src/lower/mod.rs:2460`).
- New fuzz regression harness for CLI: `crates/fe/tests/crash_regressions.rs:81` runs `fe check --standalone` over `crates/fe/tests/fixtures/crash_regressions/README.md:1` + 7 minimal `.fe` inputs to ensure “no panics/ICEs”, only diagnostics.

### DIFF
--- a/crates/common/src/diagnostics.rs
+++ b/crates/common/src/diagnostics.rs
@@ -48,6 +48,25 @@ impl CompleteDiagnostic {
     }
 }
 
+pub fn cmp_complete_diagnostics(
+    lhs: &CompleteDiagnostic,
+    rhs: &CompleteDiagnostic,
+) -> std::cmp::Ordering {
+    match lhs.error_code.cmp(&rhs.error_code) {
+        std::cmp::Ordering::Equal => {
+            let lhs_span = lhs.primary_span();
+            let rhs_span = rhs.primary_span();
+            match (lhs_span, rhs_span) {
+                (Some(lhs_span), Some(rhs_span)) => lhs_span.cmp(&rhs_span),
+                (Some(_), None) => std::cmp::Ordering::Less,
+                (None, Some(_)) => std::cmp::Ordering::Greater,
+                (None, None) => std::cmp::Ordering::Equal,
+            }
+        }
+        ord => ord,
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct GlobalErrorCode {
     pub pass: DiagnosticPass,
--- a/crates/driver/src/db.rs
+++ b/crates/driver/src/db.rs
@@ -4,7 +4,10 @@ use codespan_reporting::term::{
     termcolor::{BufferWriter, ColorChoice},
 };
 use common::file::File;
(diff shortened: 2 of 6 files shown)


crates/mir/src/lower/expr.rs
```diff
@@ -292,15 +292,16 @@ impl<'db, 'a> MirBuilder<'db, 'a> {
...
+                        if let Some(place) = self.place_for_borrow_expr(*inner) {
+                            let space = self.value_address_space(place.base);
+                            let value_ty = self.builder.body.value(value_id).ty;
+                            self.builder.body.values[value_id.index()].origin =
+                                ValueOrigin::PlaceRef(place);
+                            self.builder.body.values[value_id.index()].repr =
+                                self.value_repr_for_ty(value_ty, space);
+                        } else {
+                            let _ = self.lower_expr(*inner);
+                        }
```
sbillig, line 304: This looks like a silent mis-compile, which is worse than a panic. HIR analysis should check this, and it's reasonable for MIR to panic when HIR invariants are violated.

g-r-a-n-t: https://github.com/argotorg/fe/pull/1290/changes/87342693567a21fdba1ae704f136c72aa14500b9#diff-37993866307802c64c8d6ef31174bf4e0b4c424e510662153a5ed0a92eb69e3eR1301-R1317

PR #1293, opened 2026-02-26
Fix const array mem usage

sbillig:
Previously, each read from a const array would copy the entire array into memory. Ok for small arrays, horrible for big ones. We now do a code copy for each element. Sometime later we'll decide what to do based on the size.

This includes a special case for so that arrays of u8 and bool are backed, but for now all other int sizes are stored as 32 byte words. This isn't ideal of course, and needs to be addressed with some properly defined layout rules. Strings also need layout attention.


sbillig:
@codex review

PR #1304, opened 2026-02-27
Fix test discovery for ingots where lib.fe has no tests

cburgdorf:
## Summary                                                                                                                                                                                                        
                                               
  - `run_tests_ingot` only checked the root module (`lib.fe`) for `#[test]` functions via `has_test_functions(db, root_mod)`. If `lib.fe` contained no tests, the entire ingot was silently skipped with "No tests  
  found" — even when other modules had tests.                                                                                                                                                                       
  - Add `ingot_has_test_functions` that checks all modules in the ingot via `Ingot::all_funcs`, and use it in place of the root-module-only check.                                                                  
  - Add a regression test with an ingot fixture whose `lib.fe` has no tests while a sibling module (`helper.fe`) does.

  ## Reproduction

  Create an ingot where `lib.fe` only exports helpers and all `#[test]` functions live in other modules:

  src/lib.fe      ← pub fn add(a: u256, b: u256) -> u256 { ... }  (no tests)
  src/helper.fe   ← #[test] fn test_add() { ... }

  $ fe test .
  Warning: No tests found in .

  After this fix, `fe test` correctly discovers and runs `test_add`.

  ## Test plan

  - [x] New test `test_cli_test_ingot_discovers_tests_in_non_root_modules` passes with fix, fails without
  - [x] All existing `fe_test` / `fe_test_runner` tests pass (49/49)

### DIFF
--- a/crates/fe/src/test/mod.rs
+++ b/crates/fe/src/test/mod.rs
@@ -21,6 +21,7 @@ use colored::Colorize;
 use common::{
     InputDb,
     config::{Config, WorkspaceMemberSelection},
+    ingot::Ingot,
 };
 use contract_harness::{CallGasProfile, EvmTraceOptions, ExecutionOptions, RuntimeInstance};
 use driver::DriverDataBase;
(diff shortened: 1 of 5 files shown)


cburgdorf:
@codex review

PR #1305, opened 2026-02-27
Fix cross-ingot create2 by generating dependency contract templates

cburgdorf:
## Summary                                                                                                                                                                                                        
                                               
  - Fix compiler panic when using `create2<Contract>` where the contract is defined in a different ingot within the same workspace                                                                                  
  - Generate synthetic MIR templates for dependency ingot contracts so cross-ingot deployment works                                                                                                                 
  - Defer `ContractInitCodeOffset`/`ContractInitCodeLen` templates to avoid Sonatina verifier errors for contracts that are never deployed in their defining ingot                                                  
                                                                  
  ## Problem

  Using `create2<SomeContract>` in ingot B when `SomeContract` is defined in ingot A caused:

  failed to instantiate synthetic MIR for ContractInitCodeLen

  `lower_contract_templates` only generated synthetic MIR for contracts in the current ingot's modules. Dependency ingot contracts were never lowered, so the monomorphizer couldn't find their templates.

  ## Approach

  Three changes work together:

  1. **Dependency contract template generation** (`contracts.rs`, `mod.rs`): `lower_ingot` now iterates `resolved_external_ingots` and generates templates via a new `lower_dependency_contract_templates` function.
   It uses the host ingot's `TargetContext` so `std::evm::EvmTarget` resolves correctly.

  2. **Deferred seeding** (`ir.rs`, `contracts.rs`, `monomorphize.rs`): Dependency templates are marked `defer_root: true` so the monomorphizer skips them in `seed_roots`. `ContractInitCodeOffset` and
  `ContractInitCodeLen` are deferred for *all* contracts — they emit `sym_addr`/`sym_size` referencing the init entrypoint as a Sonatina embed symbol, which only exists when `create2` triggers deployment.

  3. **Sibling co-instantiation** (`monomorphize.rs`): When a deferred template is first instantiated on-demand, all other synthetic templates for the same contract are co-instantiated. This is needed because
  contract entrypoints call `init_handler` and `recv_arm_handlers` via `call_symbol` (pre-resolved names), which the monomorphizer cannot trace through `resolve_call_target`.

  ## Test plan

  - [x] New regression test `test_cli_test_cross_ingot_create2` — workspace with provider/consumer ingots exercising cross-ingot `create2`

### DIFF
--- a/crates/mir/tests/fixtures/transient_field_effects.mir.snap
+++ b/crates/mir/tests/fixtures/transient_field_effects.mir.snap
@@ -53,16 +53,6 @@ fn __Guard_runtime() -> ():
     eval __Guard_recv_0_0((), v0)
     terminate return_unit__Evm_hef0af3106e109414__3af54274b2985741(())
 
-fn __Guard_init_code_offset() -> u256:
-  bb0:
-    v0: u256 = code_region_offset(func_item(__Guard_init))
-    ret v0
-
-fn __Guard_init_code_len() -> u256:
-  bb0:
-    v0: u256 = code_region_len(func_item(__Guard_init))
-    ret v0
-
 fn bump_lock__TStorPtr_u256___e67d09080c0f1393(v0: u256) -> ():
   bb0:
     v1: u256 = load tstor[v0]
--- a/crates/codegen/tests/fixtures/sonatina_ir/by_ref_trait_provider_storage_bug.snap
+++ b/crates/codegen/tests/fixtures/sonatina_ir/by_ref_trait_provider_storage_bug.snap
@@ -120,18 +120,6 @@ func private %__ByRefTraitProviderStorageBug_runtime() {
         evm_invalid;
 }
 
-func private %__ByRefTraitProviderStorageBug_init_code_offset() -> i256 {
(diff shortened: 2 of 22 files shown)


crates/mir/tests/fixtures/transient_field_effects.mir.snap
```diff
@@ -54,16 +54,6 @@ fn __Guard_runtime() -> ():
     eval __Guard_recv_0_0((), v0)
     terminate return_unit__Evm_hef0af3106e109414__3af54274b2985741(())
 
-fn __Guard_init_code_offset() -> u256:
```
cburgdorf, line 56: It looks like these were previously just eagerly emitted even if they were not actually used, Now they are only emitted if they are actually used.

cburgdorf:
Eek..was a bit too fast to ping for a review. Will need to do a follow-up

sbillig:
> Defer ContractInitCodeOffset/ContractInitCodeLen templates to avoid Sonatina verifier errors for contracts that are never deployed in their defining ingot

We should just modify the sonatina verifier, not try to work around it.

PR #1310, opened 2026-03-01
Compile all test objects at once to avoid quadratic codegen cost

cburgdorf:
Previously, emit_test_module_sonatina called compile_object separately for each test. Each call invokes prepare_section on the EVM backend, which mutates shared function IR (inserting free-ptr restore instructions, splitting critical edges). Since all tests share the same module, these mutations accumulated across compilations — each successive test saw larger functions, making codegen progressively slower until it effectively hung for ingots with many tests.

Replace the per-test compile_object loop with a single compile_all_objects call that builds one section cache for the entire module. Each shared function is lowered and mutated exactly once, then the pre-compiled artifacts are distributed to their respective test objects. This reduces registry test compilation from hanging (>5 minutes with no end in sight) to ~1 second.


cburgdorf:
@codex review

PR #1311, opened 2026-03-02
Fuzz fix 2

g-r-a-n-t:
| # | Commit | Area | Why It Exists | What It Changes | Main Coverage |                                                                                       
|---|---|---|---|---|---|                                                                                                                                     
| 1 | `eff91898e` `hir: avoid self-assumptions in trait headers` | HIR trait headers | Trait headers could recurse through invalid self-assumptions involving associated projections. | Stops trait-header self-assumptions and adds direct regression coverage. | `trait_header_assoc_projection_recursive_bound.fe` |                                                                                                                                                                     
| 2 | `accc364f6` `hir: recover trait-ref and assoc-bound cycles as diagnostics` | HIR trait lowering and constraint expansion | Several recursive trait and assoc-bound cases could hang or panic instead of surfacing a normal error. | Adds cycle recovery in `lower_trait_ref`, makes recursive bound expansion safer, keeps unnamed impl methods from poisoning trait collection, and covers the affected crash-regression paths. | `499d4ead`, `1670ec70`, `acb16143`, `a8d7d78c`, `assoc_ty_cycle` |
| 3 | `11316942b` `hir: harden associated-type resolution and normalization` | HIR assoc-type resolution | Assoc-type lookup and normalization could reuse bad inference state or hit panic paths, especially around qualified traits. | Hardens canonicalization, normalization, unify remapping, and qualified-trait / assoc-type resolution. | `b02aa0c9`, `c2482ead`, `7856b3f9` |                                                                                                      
| 4 | `f2250a523` `hir: harden malformed impl and pattern analysis` | HIR pattern analysis and ops selection | Malformed impls and malformed patterns could still reach panic or `unreachable!()` paths during analysis. | Turns ops-selection failures into diagnostics and makes malformed pattern cases recoverable instead of fatal. | `a64bec5e`, `76e16ce9`, `25d24f34`, `b8fc85b7` |                                                                                                 
| 5 | `e81f3c594` `mir/index: avoid scalar address-space lookup on by-value indexing` | MIR indexing / repr queries | MIR could query scalar address-space information for by-value indexing cases where that lookup is not meaningful. | Narrows the MIR-side fix to the address-space/indexing issue only, without keeping the older invalid-HIR fallback machinery. | by-value indexing / repr-address-space lowering path |
| 6 | `d51fafedf` `mir: short-circuit diagnostics and lowering on invalid HIR` | Driver, CLI, LSP, MIR entry points | MIR work was still being attempted after HIR had already failed, and the behavior could diverge by entry point. | Centralizes MIR diagnostics flow and stops MIR lowering/diagnostics from running on already-invalid HIR. This commit also owns the invalid-HIR crash fixtures that used to motivate the old MIR fallback. | `mir_borrow_operand_not_place.fe`, `mir_match_lowering_non_ref_projection_d58afcf7.fe`, `mir_value_address_space_fallback_143863d4.fe`, `monomorphize_failed_instantiate_from_parse_error.fe`, `monomorphize_unlowered_hir_expr_survives.fe`, `monomorphize_unresolved_trait_method.fe` |
| 7 | `6195de534` `tests: minimize coverage-only crash fixtures` | Regression test corpus | A few leftover crash fixtures were still coverage-only cleanup rather than owning a distinct semantic fix. | Keeps the cleanup isolated to the residual tests-only reductions that did not clearly belong to one owning fix commit. | `hir_pretty_print_missing_param_name_5ecc2069.fe` and related coverage-only fixture cleanup |
| 8 | `8cadca252` `tests: refresh inline Fe snippets for current parser syntax` | Test maintenance only | Some inline Fe snippets and parser-facing fixtures were stale after the history cleanup and caused otherwise unrelated tests to fail. | Updates codegen/MIR/parser-facing test inputs to current Fe syntax. This is a tests-only maintenance commit, not a fuzz-fix semantic change. | `crates/codegen/src/sonatina/tests.rs`, `crates/mir/src/transform.rs`, parser-facing fixtures |


crates/fe/tests/fixtures/crash_regressions/trait_constraints_missing_core_effect_ref_c5d7f979.fe
sbillig, line 1: Why is this test failing to resolve core lib traits?

I don't think this is the right fix; we did this previously and core lib issues went undetected for a long time. A core lib invariant check some time earlier (non-panicking, but failing the compilation) would be fine, but I think the panic in this code should stay.

crates/mir/src/ir.rs
sbillig, line 1: This panic is revealing a real issue in the code leading up to this point, and should stay. This issue should be fixed sometime earlier in the pipeline.

crates/mir/src/lower/match_lowering.rs
```diff
@@ -455,16 +455,8 @@ impl<'db, 'a> MirBuilder<'db, 'a> {
...
-        else {
-            panic!(
-                "{context} requires `Ref` scrutinee (ty={}, repr={:?}, path_len={})",
-                scrutinee_ty.pretty_print(self.db),
-                scrutinee_repr,
-                path.len()
-            );
-        };
+        let current_ty =
+            crate::repr::peel_transparent_field0_projection_path(self.db, scrutinee_ty, path)?;
```
sbillig: I think this is papering over a real bug. This panic should probably stay.

crates/hir/src/analysis/ty/unify.rs
```diff
@@ -665,6 +673,12 @@ where
 
             TyData::ConstTy(cty) => match cty.data(db) {
                 ConstTyData::TyVar(var, ty) if !self.var_stack.contains(&var.key) => {
+                    if var.key.0 as usize >= self.table.len() {
+                        // This type variable originates from a different
+                        // unification table; degrade gracefully instead of
+                        // panicking on out-of-bounds access.
+                        return TyId::invalid(db, InvalidCause::Other);
+                    }
```
sbillig, line 674: I think these should be panics, not TyId::Invalid fallbacks; wrong unification is a useful bug signal.

crates/hir/src/analysis/ty/binder.rs
```diff
@@ -115,11 +115,17 @@ impl<'db> TyFolder<'db> for InstantiateFolder<'db, '_> {
...
+                }
+                return ty;
             }
             TyData::ConstTy(const_ty) => {
                 if let ConstTyData::TyParam(param, _) = const_ty.data(db) {
-                    return self.args[param.idx];
+                    if let Some(arg) = self.args.get(param.idx) {
+                        return *arg;
+                    }
+                    return ty;
```
sbillig: I think this fallback is incorrect; these panics are useful bug detection.

crates/hir/src/analysis/ty/trait_resolution/mod.rs
sbillig, line 1: This could silently ignore valid (but deep) constraints. A proper cycle detection here (or at some earlier stage) would be ideal.

crates/mir/src/ir.rs
g-r-a-n-t: https://github.com/g-r-a-n-t/fe/commit/7616b99b74721652f13f77cfb2f867295a5649b0

crates/hir/src/analysis/ty/unify.rs
g-r-a-n-t: https://github.com/g-r-a-n-t/fe/commit/7df51879de7441cd6f32878f086b6ce5d0387d58

crates/hir/src/analysis/ty/trait_resolution/mod.rs
g-r-a-n-t: https://github.com/g-r-a-n-t/fe/commit/877673802f4c4c40abf50c1db041ee201634ff49

crates/hir/src/analysis/ty/binder.rs
g-r-a-n-t: https://github.com/g-r-a-n-t/fe/commit/93a595bb434fcb3cf5d4453ab4ed2a2d334f2623

crates/fe/tests/fixtures/crash_regressions/trait_constraints_missing_core_effect_ref_c5d7f979.fe
g-r-a-n-t: I think I will drop this regression from the PR and deal with it later.

crates/mir/src/lower/match_lowering.rs
g-r-a-n-t: https://github.com/g-r-a-n-t/fe/commit/e3d7c15946118f5109e90befe217257047bb4550

g-r-a-n-t:
> @g-r-a-n-t Do you mind if I push a few commits to this branch?

Please do.

PR #1319, opened 2026-03-05
checked arithmetic

cburgdorf:
## Summary                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  - Arithmetic operators (`+`, `-`, `*`, unary `-`) now perform overflow/underflow checks at runtime for all integer types, reverting the transaction on overflow — matching Solidity 0.8+ behavior                 
  - Adds `__checked_{add,sub,mul,neg}_<type>` extern intrinsics in `core::num` that the MIR lowers operator expressions into, with overflow detection handled directly in the Sonatina and Yul backends             
  - Introduces `WrappingAdd`/`WrappingSub`/`WrappingMul` and `SaturatingAdd`/`SaturatingSub`/`SaturatingMul` traits in `core::ops` as opt-in escape hatches for intentional wrapping/saturating semantics
  - Supports both signed and unsigned types across all bit widths (u8–u256, i8–i256, usize, isize), including proper sub-word masking and sign extension for types smaller than 256 bits
  - Adds compile-time overflow detection for const-context arithmetic (division by zero, overflow in const expressions)
  - Adds signed division (`sdiv`), signed modulo (`smod`), and arithmetic right shift (`sar`) support in the Sonatina backend
  - Adds comprehensive test coverage: `checked_arithmetic_methods.fe`, `checked_arithmetic_regressions.fe`, `signed_arithmetic.fe` with both happy-path and `should_revert` tests

  ## Architecture: layered approach across backends

  The implementation uses a layered strategy that keeps overflow semantics consistent while respecting the different needs of the Yul and Sonatina backends:

  1. **Core stdlib (`core::num`)**: Operator trait impls (e.g. `Add for u8`) now call `__checked_add_u8` instead of `__add_u8`. These `__checked_*` functions are declared as `extern` intrinsics — they have no Fe
  body. The unchecked `__add_*` variants remain available and are used by the `Wrapping*`/`Saturating*` trait impls.

  2. **MIR**: When the MIR lowers a call to a `__checked_*` extern, it attaches a `CheckedIntrinsic` metadata struct (carrying the `CheckedArithmeticOp` and the concrete `TyId`) to the `CallOrigin`. This gives
  backends typed information about what overflow check is needed without embedding any particular lowering strategy at the MIR level.

  3. **Sonatina backend**: The `lower_checked_intrinsic` function pattern-matches on the `CheckedIntrinsic` to emit the arithmetic instruction followed by an inline overflow test (e.g. `result < lhs` for unsigned
   add, sign-flag analysis for signed add) and a conditional branch to a revert block. All overflow logic is emitted as native Sonatina IR — no function calls, no runtime helpers.

  4. **Yul backend**: The YUL backend takes a fundamentally different approach. The overflow logic is implemented in Fe itself via `core::num_yul`
  — a set of generic helper functions (`checked_add_unsigned_impl`, `checked_sub_signed_impl`, etc.) that use the `IntWord` trait to operate on the word-level representation of any integer type. These generic
  helpers are monomorphized per type via thin wrappers (e.g. `checked_add_u8`, `checked_neg_i256`). When the Yul emitter encounters a `__checked_*` extern call, it resolves it to the corresponding
  `core::num_yul::checked_*` function, which gets compiled to Yul like any other Fe function. This means the Yul backend gets overflow checking "for free" through normal Fe compilation, without needing any
  special intrinsic handling.

@sbillig That part :point_up:, I'm most unsure about. I remember from the last time we spoke that you set we should keep what we have implement in Fe but keep it YUL specific so this is why now the YUL backend uses `num_yul.fe` for all the safemath stuff whereas Sonatina doesn't go that extra route.

### DIFF
--- a/crates/hir/src/analysis/ty/const_ty.rs
+++ b/crates/hir/src/analysis/ty/const_ty.rs
@@ -260,17 +260,17 @@ pub(crate) fn evaluate_const_ty<'db>(
     ) -> Result<BigInt, ConstIntError> {
         match expr {
             Expr::Block(stmts) => {
-                let Some(last) = stmts.last() else {
-                    return Err(ConstIntError::Overflow);
+                let [stmt] = stmts.as_slice() else {
+                    return Err(ConstIntError::NotIntExpr);
                 };
-                let Partial::Present(stmt) = last.data(db, body) else {
-                    return Err(ConstIntError::Overflow);
+                let Partial::Present(stmt) = stmt.data(db, body) else {
+                    return Err(ConstIntError::NotIntExpr);
                 };
                 let Stmt::Expr(expr_id) = stmt else {
-                    return Err(ConstIntError::Overflow);
+                    return Err(ConstIntError::NotIntExpr);
                 };
                 let Partial::Present(inner) = expr_id.data(db, body) else {
-                    return Err(ConstIntError::Overflow);
+                    return Err(ConstIntError::NotIntExpr);
                 };
                 eval_int_expr(db, body, inner, expected)
             }
--- a/crates/uitest/fixtures/ty_check/const_block_checked_overflow.fe
+++ b/crates/uitest/fixtures/ty_check/const_block_checked_overflow.fe
@@ -1,0 +1,10 @@
+fn bad_call() {}
+
+struct U8<const N: u8> {}
+
+const BAD: u8 = {
+    bad_call()
+    1
+}
+
+fn bad(_x: U8<BAD>) {}
--- a/crates/uitest/fixtures/ty_check/const_block_checked_overflow.snap
+++ b/crates/uitest/fixtures/ty_check/const_block_checked_overflow.snap
@@ -1,0 +1,20 @@
+---
+source: crates/uitest/tests/ty_check.rs
+expression: diags
+input_file: fixtures/ty_check/const_block_checked_overflow.fe
+---
+error[3-0023]: the expression cannot be evaluated at compile time
+  ┌─ const_block_checked_overflow.fe:5:17
+  │  
+5 │   const BAD: u8 = {
+  │ ╭─────────────────^
+6 │ │     bad_call()
+7 │ │     1
+8 │ │ }
+  │ ╰─^ unsupported in const evaluation
+
+error[3-0024]: non-const function call in const context
+  ┌─ const_block_checked_overflow.fe:6:5
+  │
+6 │     bad_call()
+  │     ^^^^^^^^^^ calls in const context must be `const fn`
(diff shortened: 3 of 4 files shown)


crates/mir/src/lower/intrinsics.rs
```diff
@@ -19,6 +36,23 @@ impl<'db, 'a> MirBuilder<'db, 'a> {
...
+            let zero = self.alloc_synthetic_value(u256_ty, SyntheticValue::Int(BigUint::from(0u8)));
+            self.set_current_terminator(Terminator::TerminatingCall {
+                source: crate::ir::SourceInfoId::SYNTHETIC,
+                call: crate::ir::TerminatingCall::Intrinsic {
+                    op: IntrinsicOp::Revert,
+                    args: vec![zero, zero],
+                },
+            });
+            return Some(value_id);
+        }
```
sbillig: This should be a return-type check, not a special case. Any function that has a "never" return type should be treated as terminating.
`fn todo() -> !`

crates/codegen/src/sonatina/lower.rs
```diff
@@ -2341,3 +2363,617 @@ fn bitcast_ptr<'db, C: sonatina_ir::func_cursor::FuncCursor>(
...
+enum SubWordMask {
+    /// Unsigned: apply `and(result, mask)`
+    Unsigned { mask: I256 },
+    /// Signed: truncate to `narrow_ty` then sign-extend back to I256
+    Signed { mask: I256, narrow_ty: Type },
+}
+
+/// Returns the masking info for a sub-256-bit intrinsic, or `None` for 256-bit
+/// types that don't need masking.
+fn sub_word_mask(callee_name: &str) -> Option<SubWordMask> {
```
sbillig: fe should create sonatina ir that uses sonatina int sizes (i8, i16, etc) that match the fe int size, instead of always using i256, and let sonatina handle any masking etc. I don't think it currently does this, so doing it in fe would be fine for now.

cburgdorf: >I don't think it currently does this

Yes, right. The normal arithmetic doesn't do it yet but I now did so for the checked arithmetic. I'll add another commit to expand it to the non-checked arithmetic.

crates/mir/src/lower/intrinsics.rs
cburgdorf: Yep, should be generalized now.

cburgdorf:
@codex review

crates/codegen/tests/fixtures/sonatina_ir/math_ops.snap
```diff
@@ -10,10 +10,36 @@ func private %math_ops() -> i256 {
...
+}
+
+func private %u64_haee7f05a097ffa16_eq_he50383edd273619f_eq(v0.i256, v1.i256) -> i256 {
+    block0:
+        v3.i64 = trunc v0 i64;
+        v4.i64 = trunc v1 i64;
+        v5.i1 = eq v3 v4;
+        v6.i256 = zext v5 i256;
+        return v6;
+}
```
sbillig: - Sonatina codegen replaces all smaller-sized ints with i256; if you change that, we could avoid doing all this truncation noise.

- This `rem` function is manually doing the same work as evm_umodo (checking if the rhs is 0).

- Ideally, instead of emitting these helpers, we could (for each function with arithmetic ops) emit a single `revert` block within the function, and jump to that one block from every checked op.

- Not sure why only `%` is getting this treatment, not all of the other ops.

Issue #1181, opened 2025-12-11
don't attempt parser recovery when compiling/checking via the CLI


sbillig:
The parser should take an option to control whether the parser will attempt to recover from syntax errors. Probably `true` for the language server, `false` for the CLI (controllable via a flag, perhaps).

PR #1327, opened 2026-03-11
Add parser recovery mode option and tests

KiChjang:
Addresses #1181.

Adds an option to the CLI and parser on whether or not the parser would automatically recover from syntax errors. False by default for the CLI, true when it comes from the language server.


KiChjang:
@codex review

PR #1334, opened 2026-03-13
Immutable contract fields

sbillig:
Adds immutable-by-default contract fields.

Contract fields are now code-backed unless declared with `mut`. A field like `x: u256` is treated as immutable contract data, while `mut x: u256` keeps the previous storage-backed mutable behavior.

Immutable fields must be assigned during `init` on every successful initialization path. During `init`, code-backed fields are exposed as temporary mutable memory-backed providers via `uses (mut field)`. After initialization, their values are serialized into a tail section appended to the deployed runtime bytecode and are read back as code-backed read-only providers.


micahscopes:
c.c. @EmanHerawy

micahscopes:
addresses #1204

EmanHerawy:
Hi @micahscopes Thanks for addressing this. I'm not a core dev and haven't worked in Fe code base yet but I used AI to walk me through the PR  I noticed that the way you are handling it here is different than how it is handled in [solidity](https://github.com/argotorg/solidity/blob/27ac306927f09c592f8336b06bb2526f8960b5cc/libevmasm/Assembly.cpp#L1392) which would increase the gas cost as we have to `mload`, `CODECOPY` and `mload` the value while solidity git in one step via `PUSH32`
to summarize: 

in solidity EVERY CALL:
    PUSH32 <value>   
    
in fe     
 EVERY CALL:
    mload(0x40)   
    codecopy(...)   
    mload(p)         

```rust 
//ref https://github.com/argotorg/fe/pull/1334/changes#diff-6a1615171bfe819703f7540b34c2e8a82b5d00d6554f1273b7147603fc42371eR1211
fn code_load_helper_doc() -> YulDoc {
    YulDoc::block(
        "function __fe_code_load(off) -> out ",
        vec![
            YulDoc::line("let p := mload(0x40)"),
            YulDoc::line("codecopy(p, off, 0x20)"),
            YulDoc::line("out := mload(p)"),
        ],
    )
}
```
I believe you are way better than me in this area, just wanted to flag it. Happy to be wrong though

sbillig:
@EmanHerawy You're definitely right. The PUSH32 patching that solc does is trickier, and will require some extra work in sonatina. I think the approach here is fine as a first pass to implement the feature (and it's certainly better than using storage slots), and we can refine it later.

g-r-a-n-t:
@codex review

g-r-a-n-t:
@codex review

g-r-a-n-t:
@codex review

sbillig:
@g-r-a-n-t what's the status of this? We need immutable fields for the uniswap port.

g-r-a-n-t:
@codex review

crates/codegen/tests/fixtures/sonatina_ir/immutable_contract_field_init_set.snap
```diff
@@ -0,0 +1,885 @@
...
+
+    block2:
+        evm_revert 0.i256 0.i256;
+
+    block3:
+        v6.objref<@layout_7> = obj.alloc @layout_7;
+        call %decode_runtime_args v6;
+        v8.i256 = evm_code_size;
+        v10.i256 = add v8 -32.i256;
+        mstore v0 v10 i256;
```
micahscopes: agent:

> This looks like it returns the field's code offset rather than its value: v10 =
> codesize-32 is the handle, it gets mstored here, and the handler (L35) just
> mloads that word back and returns it — no codecopy on the read path. I think
> Code isn't in `default_borrow_transport_set`, so the scalar handle degrades to a
> plain word. The array path materializes via object copy, which is why the array
> test passes; there's no executing test for a scalar get-after-init.

crates/hir/src/analysis/ty/ty_check/contract.rs
```diff
@@ -780,6 +793,1076 @@ pub fn check_contract_recv_arm_body<'db>(
...
+                let continues = iter_flow.continues;
+                let mut normal_exits = vec![iter_flow.normal.clone()];
+                if let Some(iter_state) = iter_flow.normal.clone() {
+                    let body_flow = self.analyze_expr(body, iter_state);
+                    returns.extend(body_flow.returns);
+                    normal_exits.extend(body_flow.breaks.into_iter().map(Some));
+                    // Body continuations target this loop and are consumed here.
+                }
+                AssignmentFlow {
+                    normal: merge_normal_states(normal_exits),
```
micahscopes: agent:

> `body_flow.normal` never makes it into `normal_exits`, so facts written in the loop
> body don't survive to the post-loop state: `t = true; for .. { t = false };
> while t { x = v; break }` is accepted but can leave x unassigned. (While has a
> related issue — single backedge pass, so chained facts like `x = y; y = 0` need
> two passes to converge.) Killing facts for anything assigned in a loop body
> might be the simplest fix.

micahscopes (commented):
agent:

> Related loop-facts case: `cond_value` reads the scrutinee for `Cond::Let` but
> ignores the pattern, so `while let false = b` is treated as `while b` —
> `let b = true; while let false = b { x = v; break }` is accepted even though
> the body never runs.

micahscopes:
agent:

> Edge case in the init serializer: serialize_init_immutables_into_buffer
> (synthetic.rs:688) selects bindings by `RefKind::Provider { space: Memory }` —
> the shape init-time immutables get after the Code→Memory remap. But a genuine
> Memory-space field (e.g. `f: MemPtr<u256>`; nothing in analysis rejects it)
> produces the same shape at init, with slot taken from the Memory-space
> numbering — so alongside Code-backed fields it serializes into the immutables
> buffer at a colliding offset, or past its end (the buffer is sized by Code
> slots only). Related: pre-PR these fields were a hard codegen error ("memory
> contract field handles are not supported"); post-PR they're silently a fresh
> zeroed ObjAlloc per entry. If that's not an intended feature, rejecting
> Memory-space contract fields in analysis would close it; either way,
> identifying init immutables by an explicit flag rather than the
> Memory-provider shape seems safer.

micahscopes:
agent:

> Two smaller things: `ContractHost::immutable_field` / `init_immutable_field`
> (contracts.fe:22-23) don't seem to be consumed anywhere (handles lower via the
> EffectHandle builtin), and since they lack default bodies they're breaking for
> existing ContractHost impls — should they get defaults or be dropped? And the
> init-time semantics (read-before-assign yields zeros, reassignment allowed,
> last write wins — matching Solidity ≥0.8.21) are defensible but worth a line
> in the newsfragment/docs, since the definite-assignment check guarantees a
> value is assigned somewhere on each path, not before each use.

g-r-a-n-t:
@codex review

g-r-a-n-t:
This should be good to go @sbillig @micahscopes

PR #1340, opened 2026-03-18
Add payable

cburgdorf:
## Summary

  This PR adds a `#[payable]` attribute for contract `init` blocks and `recv` arms, addressing two problems:

  **Safety:** Fe contracts currently accept ETH silently on all entrypoints. Users can lose funds by sending ETH to functions that don't handle it. With `#[payable]`, the compiler injects a `CALLVALUE` check that
   reverts non-payable calls.

  **ABI correctness:** Without an explicit payable signal, JSON ABI generation (planned in a follow-up) would have to mark every function as `"stateMutability": "payable"` — even pure views and read-only
  accessors — since the compiler couldn't distinguish intent. With `#[payable]`, the ABI can correctly derive `"payable"`, `"nonpayable"`, `"view"`, and `"pure"`, which is required for proper integration with
  wallets, block explorers, and tools like Foundry or ethers.js.

  ### What's included

  - **Parser:** Attribute lists are now parsed before `init` blocks and `recv` arms/blocks, with lookahead to distinguish field attributes from init/recv attributes
  - **HIR:** `attributes` field propagated on `ContractInit` and `ContractRecvArm` with `is_payable()` helpers
  - **MIR:** New `IntrinsicOp::Callvalue` wired through ir, fmt, escape analysis, intrinsics mapping, and both codegen backends (Sonatina + Yul)
  - **MIR lowering:** `CALLVALUE != 0 → revert` guards emitted in both `lower_init_entrypoint` and `lower_runtime_entrypoint` for non-payable entrypoints
  - **Formatter:** `fe fmt` preserves `#[payable]` placement on init and recv arms
  - **Validation:** Full diagnostic pipeline (`PayableError` → `PayableAttrPass` → `DiagnosticVoucher`) rejects `#[payable]` on unsupported items (fn, struct, enum, msg variant, recv block, const, extern fn,
  etc.) and malformed forms like `#[payable(foo)]` or `#[payable = 1]`
  - **Test harness:** `RuntimeInstance::fund_contract()` gives deployed test contracts ETH balance so tests can exercise payable semantics with non-zero value

  ### Example

  ```fe
  pub contract Vault {
      #[payable]
      init() {}

      recv VaultMsg {
          #[payable]
          Deposit {} { ... }          // accepts ETH → ABI: "payable"

          GetBalance {} -> u256 { 0 } // reverts if ETH sent → ABI: "pure"
      }
  }

  Test plan

  - Non-payable recv arm works with zero value
  - Non-payable recv arm reverts with non-zero value (should_revert)
  - Payable recv arm works with zero value
  - Payable recv arm accepts non-zero ETH
  - Non-payable init reverts with non-zero value (should_revert)
  - Payable init accepts non-zero ETH
  - Contract without init deploys with zero value
  - Contract without init reverts with non-zero value (should_revert)
  - MIR snapshot verifies CALLVALUE guards present/absent as expected
  - UI tests for #[payable] on unsupported items (fn, struct, enum, const, extern fn)
  - UI tests for #[payable] on msg variant (targeted error message)
  - UI tests for #[payable] on recv block (targeted error message)
  - UI tests for malformed #[payable(foo)] and #[payable = 1]
  - All existing tests pass (cargo test --workspace)

### DIFF
--- a/crates/codegen/src/sonatina/lower.rs
+++ b/crates/codegen/src/sonatina/lower.rs
@@ -1527,6 +1527,9 @@ fn lower_intrinsic<'db, C: sonatina_ir::func_cursor::FuncCursor>(
             )))
         }
         IntrinsicOp::Caller => Ok(Some(ctx.fb.insert_inst(EvmCaller::new(ctx.is), Type::I256))),
+        IntrinsicOp::Callvalue => Ok(Some(
+            ctx.fb.insert_inst(EvmCallValue::new(ctx.is), Type::I256),
+        )),
         IntrinsicOp::ReturnData | IntrinsicOp::Revert => Err(LowerError::Internal(
             "terminating intrinsic must be lowered as Terminator::TerminatingCall".to_string(),
         )),
--- a/crates/codegen/src/sonatina/tests.rs
+++ b/crates/codegen/src/sonatina/tests.rs
@@ -1,4 +1,5 @@
 use driver::DriverDataBase;
+use hir::HirDb;
 use hir::hir_def::{ItemKind, TopLevelMod};
 use mir::analysis::{CallGraph, build_call_graph, reachable_functions};
 use mir::{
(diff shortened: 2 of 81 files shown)


cburgdorf:
@codex review

PR #1351, opened 2026-03-22
gh workflow: enable vendored-openssl for release builds for all targets

sbillig:


sbillig:
Codex says this is a bad idea 🤷

PR #1392, opened 2026-04-09
LSP: panic surfacing, logging hygiene, and dispatch-deadlock fix

micahscopes:
## Summary

Fixes a cluster of LSP regressions: silent crashes with no visible error, slow cold-start, duplicate `fe lsp` instances spawned by Zed, and debugging-by-spelunking-through-IDE-logs. No new user-facing features. This PR hardens error surfacing, tightens logging, and adds workspace and instance diagnostics so the next LSP bug report is easier to triage.

## What's in here

Four themes, in dependency order:

### Panic surfacing

Worker-thread panics were previously swallowed by the spawn path and left the LSP running in a broken state. Panics now propagate through `WorkerError`, salsa DB context is captured via `DbPanicContext`, and both tracing and the file log receive the frame. A regression test reproduces the silent-swallow behavior pre-fix. Also demotes the `setup_unhandled` notification/event path from `warn` to `debug`, since it fires for normal traffic (`$/cancelRequest`, `$/setTrace`, custom `experimental/*` methods) and was drowning out real warnings.

### Workspace / instance detection

Startup now logs the workspace discovery path and warns when sibling or stale `fe-lsp` processes are present (the "Zed spawned three language servers" symptom). The detector distinguishes `SiblingLive` (another pid for the same workspace root), `StaleFound` (dead pid record), and `Malformed` (unparseable record). Also fixes a custom-target filter matching bug so per-request activity actually surfaces in filtered logs.

### async-lsp upgrade and dispatch-deadlock fix

Drops `micahscopes/async-lsp pub-inner-type-id` in favor of `fe-lang/async-lsp` main. The fe-lang fork is `oxalica/async-lsp 0.2.3` plus two commits for a `MainLoop::run` dispatch deadlock triggered by bursts of `>max_concurrency` concurrent requests (repro: goto-def burst). The regression test `repro_concurrent_goto_burst_does_not_deadlock` runs in the CI default (promoted from `#[ignore]`). An upstream PR to oxalica with the two fix commits is pending; we'll repoint to crates.io once it lands.

### Logging hygiene

File logs now land in `<workspace>/.fe-lsp/` instead of the global `~/.cache/fe-language-server/`. The shared directory meant one workspace's retention budget could evict another's forensic evidence. Retention is 5 files per workspace (down from 50), with a 50 MB rotation cap per session via a new `RotatingFileWriter`, so ~250 MB worst case per workspace. A dual-subscriber setup captures verbose per-request tracing to disk while keeping stderr quiet by default.

## Dependency note

`crates/language-server/Cargo.toml` pins `async-lsp` at `{ git = "https://github.com/fe-lang/async-lsp", branch = "main" }`. `fe-lang/async-lsp` is a fresh org-level fork directly off `oxalica/async-lsp`; the only delta is the two `MainLoop::run` deadlock fixes on top of 0.2.3.

## Test plan

- [x] `cargo test -p fe-language-server`: 93 passed, 0 failed, 1 ignored
- [x] `cargo clippy -p fe-language-server --all-targets`: clean
- [x] `cargo fmt --all -- --check`: clean
- [x] `repro_concurrent_goto_burst_does_not_deadlock` runs in CI default (promoted from `#[ignore]`)
- [x] Manual Zed smoke: cold start, workspace log dir populated, retention enforced
- [ ] Reviewer sanity check of panic-handling path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ tarpaulin-report.html
 .vscode
 .DS_Store
 **/*.fe-lsp.json
+**/.fe-lsp/
 
 # OpenSpec and AI assistant files
 openspec/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -668,7 +668,7 @@ dependencies = [
  "futures-lite",
  "parking",
  "polling",
- "rustix 1.1.3",
+ "rustix",
  "slab",
  "windows-sys 0.61.2",
 ]
@@ -686,13 +686,13 @@ dependencies = [
 
 [[package]]
 name = "async-lsp"
-version = "0.2.1"
-source = "git+https://github.com/micahscopes/async-lsp?branch=pub-inner-type-id#5ce6e1ba89162d90825bafdd4f9d265fa9909d02"
+version = "0.2.3"
+source = "git+https://github.com/fe-lang/async-lsp?branch=main#ce9c0f26ebf417754247451f5293a1bc182ca6a2"
 dependencies = [
  "futures",
  "lsp-types",
  "pin-project-lite",
- "rustix 0.38.44",
+ "rustix",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
@@ -729,7 +729,7 @@ dependencies = [
  "cfg-if",
  "event-listener 5.4.1",
  "futures-lite",
- "rustix 1.1.3",
+ "rustix",
 ]
 
 [[package]]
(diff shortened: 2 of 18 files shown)


micahscopes:
@cburgdorf, I'd like to get this in before the bountiful launch.

cburgdorf (approved):
It's a pretty gigantic changeset so I just skimmed over the code and took into account what my agents told me. Nothing stands out that would preventing merging imho .

:shipit:

micahscopes:
Any mishaps should be quarantined to the language server at any rate.  @g-r-a-n-t let me know if anything goes wrong when you get a chance to try this out!

PR #1393, opened 2026-04-10
Bump sonatina to eb50941

sbillig:


cburgdorf (approved):
Tested against bountiful. Works 🎉

PR #1394, opened 2026-04-13
Release notes

cburgdorf:
- Adds the newsfragments for the changes that happened between 26.0.0 and current master
- creates a `CHANGLOG.md` in root
- some needed settings tuning


cburgdorf:
Should we also add a validation step to our PRs to require them to have a newsfragment file going forward?

Issue #1348, opened 2026-03-22
lib: better `assert` function


sbillig:
at least `assert(x == y, "some description")`

cburgdorf:
Taking this on

sbillig:
#1471

PR #1399, opened 2026-04-16
Add `assert_msg(cond, "message")` emitting Solidity `Error(string)`

cburgdorf:
Introduces `assert_msg<const N: usize>(b: bool, message: String<N>)` in
the std prelude. On `b == false` it reverts with the 4-byte selector
`0x08c379a0` (keccak256("Error(string)")[..4]) followed by the
ABI-encoded message, matching Solidity's `require(cond, "message")` /
`revert("message")`.
    
Also improves const-generic inference so a string literal can satisfy a
generic `String<N>` parameter without an annotation at the call site:
when a string literal is passed to a parameter of shape `String<N>` and
the inference key for `N` appears only in that one parameter, the
checker pins `N` to the literal's min length. Keys shared across
multiple parameters are left alone so the other arguments (e.g.
`[u8; N]`) still drive inference for `N`.

Implements #1348

### DIFF
--- a/crates/fe/tests/fixtures/fe_test_runner/assert_msg.fe
+++ b/crates/fe/tests/fixtures/fe_test_runner/assert_msg.fe
@@ -1,0 +1,53 @@
+use std::evm::{Address, Call, Create, Evm, RawMem, last_returndata}
+use std::evm::mem
+
+#[test(should_revert, selector = 0x08c379a0)]
+fn test_assert_msg_reverts() uses (evm: mut Evm) {
+    assert_msg(false, "boom")
+}
+
+#[test]
+fn test_assert_msg_passes() uses (evm: mut Evm) {
+    assert_msg(true, "should not revert")
+}
+
+msg ReverterMsg {
+    #[selector = 0xaabbccdd]
+    Trigger,
+}
+
+pub contract Reverter {
+    recv ReverterMsg {
+        Trigger {} {
+            assert_msg(false, "boom")
+        }
+    }
+}
+
+#[test]
+fn test_assert_msg_payload_layout() uses (evm: mut Evm) {
+    let addr = evm.create2<Reverter>(value: 0, args: (), salt: 0)
+
+    // Build calldata containing only the ReverterMsg::Trigger selector.
+    let selector_word = (0xaabbccdd as u256) << 224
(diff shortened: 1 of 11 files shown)


cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

sbillig:
Proposal:

Define assert_msg in terms of Text aka DynString:
```rust
     use super::abi::Text

     pub fn assert_msg(_ b: own bool, _ message: own Text) {
         if !b {
             let (ptr, len) = encode_calldata(0x08c379a0, message)
             ops::revert(offset: ptr, len: len)
         }
     }
```
(this might increase gas usage)

and remove the hir changes. There's an issue in hir to be solved, but the current solution isn't the right one. I think we need to modify the type unifier so that it imposes an `N >= literal_len` constraint for `String<N>`, which probably isn't a huge change, but could be deferred.

cburgdorf:
@sbillig ok, made the requested change.

sbillig:
I think the abi-encoded payload is wrong. I'm rebasing and investigating.

sbillig:
Here's what the agent says:

  Change assert_msg from:

  let (ptr, len) = encode_calldata(0x08c379a0, message)

  to:

  let (ptr, len) = encode_calldata(0x08c379a0, (message,))

  with a short comment explaining why.

  Why: encode_calldata expects its second argument to represent the full ABI payload after
  the selector. Generated msg/custom-error structs already do that. A bare Text/DynString
  encodes as its dynamic payload:

  len || bytes

  but Error(string) has one ABI argument, so it needs the single-argument head first:

  offset(0x20) || len || bytes

  So for "boom", the correct full revert data is:

  08c379a0
  0000000000000000000000000000000000000000000000000000000000000020
  0000000000000000000000000000000000000000000000000000000000000004
  626f6f6d00000000000000000000000000000000000000000000000000000000

cburgdorf:
Ok, I did what the agent said :saluting_face:

cburgdorf:
...and added another test that covers this.

g-r-a-n-t (approved):

PR #1404, opened 2026-04-22
mega mir redo

sbillig:
The big change here is to split MIR into phases. SMIR -> NSMIR -> MIR -> Sonatina/Yul

SMIR ("semantic mir") - this is basically hir in cfg form. Used for const time function evaluation (and thus needs to live in the hir crate to avoid a dependency cycle between hir and mir). This allows us to cleanly expand const fn capabilities, and replaces the previous ugly hir-based implementation.
NSMIR (normalized) - includes concrete borrowing and place/origin information. Used for borrow checking and other diagnostics.
MIR - low-level, but backend neutral representation that's lowered into sonatina/yul. Includes concrete layouts, const/code regions, etc

(Aside: SMIR and NSMIR could maybe be collapsed together. NSMIR just derives a bunch of information that isn't relevant for ctfe)

This fixes a lot of tricky bugs we had previously due to sidecar info alongside the insufficient representation, and information being deduced/inferred in late stages because it wasn't propagated forward by earlier stages.

Unfortunately basically every snapshot has been updated. Name mangling that was previously done in the old MIR is now deferred to codegen, and yul currently uses a simple incremented suffix for conflicting names, which should be changed.

(Aside: I think we should consider replacing the big codegen snapshots with small targeted snapshots.)

TODO, here or in quick follow-up prs
- [ ] remove diagnostic blocks preventing const fn functionality that's already supported (mut, while, ..?)
- [ ] add uninitialized var analysis. most of the machinery for this is in place.
- [x] port over previous naming scheme, or make a new one
- [ ] newsfragment


sbillig:
@codex review

sbillig:
@codex review

cburgdorf:
Not a proper review but looks like the bountiful tests crash against this branch:

```
$ ~/Documents/hacking/fe/fe/target/debug/fe test
running `fe test` for 3 inputs (jobs=8)

COMPILING        games    /home/cburgdorf/Documents/hacking/fe/bountiful/contracts/ingots/games

thread '<unnamed>' (2999015) panicked at crates/mir/src/runtime/lower/classify.rs:1036:17:
runtime call resolution failed during return-class inference for SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>): runtime trait-call resolution failed to resolve a concrete impl body: caller=SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) decl=SemanticInstanceKey(Id(33f04), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) method=encode_payload concrete_inst=Encode<Sol> original_inst=Encode<Sol>
                 games    Sonatina backend panicked while emitting test module: runtime call resolution failed during return-class inference for SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>): runtime trait-call resolution failed to resolve a concrete impl body: caller=SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) decl=SemanticInstanceKey(Id(33f04), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) method=encode_payload concrete_inst=Encode<Sol> original_inst=Encode<Sol>
ERROR [66.845s]  games    Sonatina backend panicked while emitting test module: runtime call resolution failed during return-class inference for SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>): runtime trait-call resolution failed to resolve a concrete impl body: caller=SemanticInstanceKey(Id(33ef3), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) decl=SemanticInstanceKey(Id(33f04), PhantomData<&salsa::interned::Value<fe_hir::analysis::semantic::instance::semantic::SemanticInstanceKey<'_>>>) method=encode_payload concrete_inst=Encode<Sol> original_inst=Encode<Sol>
COMPILING        registry /home/cburgdorf/Documents/hacking/fe/bountiful/contracts/ingots/registry

```

This is against latest bountiful master

sbillig:
> Not a proper review but looks like the bountiful tests crash against this branch:

shiiiiit

sbillig:
The core/std abi changes were meant to be a separate pr that this one depends on. I'll see if I can split that out cleanly. (Though I guess I'm not sure if it's worthwhile, given that the old defunct mir code would need to be updated)

sbillig:
@cburgdorf 
Initial fix for the bountiful panic is in, which is not lowering to MIR when there are errors in dependencies. The errors are due to the core abi trait changes that I unintentionally snuck in.

The deeper hardening to not panic when MIR lowering runs into this issue, but I'm going to defer that to a followup.

sbillig:
> @cburgdorf Initial fix for the bountiful panic is in, which is not lowering to MIR when there are errors in dependencies. The errors are due to the core abi trait changes that I unintentionally snuck in.
> 
> The deeper hardening to not panic when MIR lowering runs into this issue, but I'm going to defer that to a followup.

I'll push up a bountiful pr with the required changes.

Edit: https://github.com/fe-lang/bountiful/pull/8

micahscopes:
Ran rosetta-fe against this branch vs v26.0.1.

## Two compile failures

**`EffectRef<Evm>` on capability passthrough.** Any fn that declares `uses (evm: mut Evm)` and calls another fn with the same declaration fails with `` `evm` doesn't implement `EffectRef<Evm>` ``. Smallest case:

```fe
use std::evm::Evm
pub struct Buf { pub ptr: u256, pub len: u256 }
impl Copy for Buf {}
impl Buf {
    pub fn put(self, v: u256) -> Buf uses (evm: mut Evm) {
        evm.mstore(addr: self.ptr, value: v); self
    }
    pub fn wrapper(self) -> Buf uses (evm: mut Evm) { self.put(v: 1) }
}
```

Every call shape hits it, not just chaining:

```fe
// fails on the single call
pub fn one(self, a: u256) -> Buf uses (evm: mut Evm) { self.put(v: a) }

// fails at both links
pub fn chained(self, a: u256, b: u256) -> Buf uses (evm: mut Evm) {
    self.put(v: a).put(v: b)
}

// also fails when the intermediate goes through a let
pub fn seq(self, a: u256, b: u256) -> Buf uses (evm: mut Evm) {
    let x = self.put(v: a); x.put(v: b)
}
```

Trait-method dispatch is affected too: rosetta-fe's `Transcript::write<T: Writable>` calls `val.write_to(ptr: ...)` (a trait method declared `uses (evm: mut Evm)`) and hits the same error.

Suspected spot: `effect_provider_target_resolution` in `ty_check/expr.rs` around L2332 — the `as_capability` fast path doesn't seem to fire for forwarded `evm`, so resolution falls through to the EffectRef goal where `mut Evm` isn't covered by the reflexive impl. In rosetta-fe the verifier example hits this 182 times (plonk.fe, halo2.fe, transcript.fe).

**"runtime package has no root objects"** on packages whose contract lives in a submodule re-exported from `lib.fe`. Smallest case:

```
# src/lib.fe
pub use sub::{self, *}
```
```fe
// src/sub.fe
use std::abi::sol
msg M { #[selector = sol("ping()")] Ping -> u256 }
pub contract C {
    init() {}
    recv M { Ping -> u256 { 42 } }
}
```

Moving `contract C` (and its `msg`) inline into `lib.fe` works around it, so the shape that breaks is specifically "contract reachable only via re-export from the root module." Guard is at `crates/codegen/src/runtime_package.rs:9`; looks like `RuntimePackage::root_objects` isn't following the re-export. Hits math, merkle, and shared/ec in rosetta-fe (amm / erc20 / escrow / governance all declare the contract in `lib.fe` directly and build fine).

## Gas deltas on what still compiles

`testGas_fe_*` only; fuzz trimmed to 32 runs; percentages relative to v26.0.1.

```
amm        addLiquidity -1.9   getK +0.7   swapAForB -2.1
erc20      approve      +0.2   transfer +0.5
escrow     deposit      -0.1   getBalance -0.5   release +2.6
governance balanceOf    -0.5   delegate +1.2   execute +0.9   mint -0.3
           propose      +1.3   transfer +1.3   vote +1.7
```

amm's hot paths down ~2%, governance uniformly up ~1-1.7%, erc20/escrow flat.

Caveats: v26.0.1 baseline (master since is only docs work); had to exclude verifier/math/merkle/ec from the workspace to make `fe build ../..` succeed in setUp.

sbillig:
bountiful, rosetta, chz, and (local) zk-kit tests are all passing. chz takes about 3 minutes to build the test suite on my laptop; will work on improving that (but maybe for a later patch release).

All the findings reported by @g-r-a-n-t have been addressed. Some aren't related to this branch and I pushed up fixes as separate PRs and opened an issue for the parser panic.

cburgdorf:
Epic work 🎉

PR #1405, opened 2026-04-22
hir: materialize event encode receiver before reserving encoder head

cburgdorf:
When the emitted tuple of data fields is built inline after reserve_head(), its heap allocation lands inside the encoder's output range and corrupts log data. Bind the tuple (or single field) to `values` before constructing the encoder so its memory is allocated first.

### DIFF
--- a/crates/fe/tests/fixtures/fe_test/abi_dynamic_payload.fe
+++ b/crates/fe/tests/fixtures/fe_test/abi_dynamic_payload.fe
@@ -1,0 +1,13 @@
+use std::abi::Bytes
+use std::evm::encode_abi_payload
+
+#[test]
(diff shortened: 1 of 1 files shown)


cburgdorf:
@codex review

cburgdorf:
Something that surfaced when porting the ETH deposit contract to Fe.

cburgdorf:
This can wait until we've sorted out all the rest for the 26.1.0 release but I just want to make sure a fix for this bug will be included in the next release so that a implementation of the ETH deposit contract becomes viable.

sbillig:
This was working around a std abi implementation issue that has been fixed on master. I force-pushed to change this to a single commit that adds the test case.

PR #1419, opened 2026-04-30
Doc generation: msg variants as enum-parity children, contract polish, theming hooks

micahscopes:
## Summary

Restructures `msg` rendering so variants live as variant-kind children of the parent msg page (mirrors enum variants), adds first-class `init` and `recv_handler` children for contract pages, and parameterizes the doc viewer for downstream theming. Also fixes the doc-site deploy ref.

## Changes

- Model `msg` variants as `variant` children of their parent msg; drop the standalone msg_variant DocItem
- Add `Init` and `RecvHandler` DocChildKinds for contract pages
- Bump `SCHEMA_VERSION` 2 → 4; add v2→v3 and v3→v4 migrations in `feMigrate`
- Skip desugared msg artifacts (per-variant structs and their impl blocks) during extraction
- Resolve breadcrumb segments against ancestor kinds rather than hardcoding `/mod`
- Substitute ingot config name for `lib` on the root module page
- Add `fe doc --include-tests`; default off (test fns pollute the public API sidebar)
- SCIP: prefer specific descriptors when occurrences tie on byte range; exclude attribute-internal occurrences from rich-signature overlays
- Style scrollbars for dark mode, wrap long sidebar entries, expose font / weight / layout knobs as CSS variables
- Drop search results lacking `doc_url` (they rendered as unclickable links)
- **CI**: fix doc-site deploy ref `gh-pages` → `main` (matches the actual fe-docs branch GitHub Pages serves from)

## Notes

Existing 26.0.0 / 26.0.1 doc bundles will run the v2→v3 migration in-browser and lose msg-variant rows. Worth a follow-up regeneration pass for prior versions once this lands.

## Test plan

- [ ] `cargo test -p fe-web` — schema snapshot consistency
- [ ] Local `fe doc` against a contract with msgs — variants render under their parent msg page
- [ ] Local `fe doc` against a contract with `init` / receive handler — both surface as contract page children
- [ ] Tag a test release on a fork to confirm the docs job pushes to fe-docs main

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -219,7 +219,7 @@ jobs:
           repository: fe-lang/fe-docs
           token: ${{ secrets.DOC_DEPLOY_TOKEN }}
           path: fe-docs
-          ref: gh-pages
+          ref: main
 
       - name: Deploy docs
         run: |
--- a/crates/fe-web/assets/fe-doc-item.js
+++ b/crates/fe-web/assets/fe-doc-item.js
@@ -304,7 +304,8 @@ class FeDocItem extends HTMLElement {
 
   _renderBreadcrumbs(item) {
     var segments = item.path.split("::");
-    var base = this.getAttribute("base") || "";
+    var index = this._getIndex();
+    var items = (index && index.items) || [];
     var html = '<nav class="breadcrumb">';
     var accumulated = "";
     for (var i = 0; i < segments.length; i++) {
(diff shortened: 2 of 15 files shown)


micahscopes:
Pushed fixes for both Codex P2 findings:

- **extract.rs**: ingot-name rewrite now also requires `doc_item.path == cfg_name` so nested submodules like `foo::lib` are left alone
- **fe-scip-store.js**: v2→v3 migration drops legacy sub-anchors instead of concatenating them, so migrated links resolve to the variant row instead of producing a doubled-tilde hash that matches no element

Squashed into the relevant commits (`Flesh out contract pages and polish navigation` and `Bump doc schema for msg variant restructure`).

micahscopes:
@cburgdorf was hoping to get these updates into the bountiful docs as soon as possible

sbillig:
@micahscopes status on this? should we merge before 26.2?

micahscopes:
@sbillig yeah lemme get it fixed up today, if all goes well it will enable the doc site upgrade to just work

micahscopes:
Should be good

g-r-a-n-t (approved):

PR #1420, opened 2026-04-30
fe test: keep error output with status line

sbillig:
Log output from failed tests isn't (wasn't) emitted atomically with the FAIL status line, which can result in shuffled output.


sbillig:
@codex review

PR #1422, opened 2026-05-05
Solidity ABI optimizations

g-r-a-n-t:
| Commit | Area | Change |
| --- | --- | --- |
| [`5f093f79c`](https://github.com/argotorg/fe/commit/5f093f79c0f135f8481793f98be0f150c4540d4d) | Runtime selector dispatch | Replace linear selector dispatch with binary search. Reviewer feedback asks for this to be split into a separate configurable PR. |
| [`18df0e539`](https://github.com/argotorg/fe/commit/18df0e539996667ab58f71d499cca524611d25c1) | ABI passthrough runtime | Avoid malloc in trivial ABI passthrough recv arms by validating calldata and returning it directly. |
| [`7008838dc`](https://github.com/argotorg/fe/commit/7008838dcf43539733594b224018759a7bf2609a) | Test/bench/docs artifacts | Add ABI optimization benchmark artifacts, gas reports, docs, and updated snapshots. |


sbillig:
Could we pull this one out into a separate PR and make it configurable via an attribute and perhaps fe.toml?

aa158ec69 | Runtime selector dispatch | Replace linear selector dispatch with binary search.

sbillig:
9f8000eda | Calldata validation arithmetic | Avoid checked arithmetic in calldata view validation where bounds are established.

It looks like there are some cases where we should disable checked arithmetic here. Most of this seems to be replacing the built-in checked arithmetic with a manual checking, which suggests to me that the built-in checked arithmetic could/should be optimized.

PR #1425, opened 2026-05-06
Fix LSP CPU/memory blowup from SCIP regen on every edit

### LINKED ISSUE #1424: LSP: scip generation burns cpu and memory
A `fe lsp` process spawned by zed in the fe repo was burning 200% cpu and using a lot of ram.
Note that I was running a `cargo install`ed fe binary, which might be a bit outdated. Agent analysis:

  Root cause: LSP live doc reload is doing full documentation + SCIP regeneration for
  the whole workspace plus builtin core/std, and that work is extremely expensive and
  not cancellable. The process is still running at ~208% CPU with ~1.8 GB RSS; samples
  show a 2.3-2.4 GB physical footprint and 28 GB peak.

  The hot path is not the LSP request loop. Tokio/main threads are mostly parked; Rayon
  workers are busy in fe::scip_index:

  - handle_did_change / did_save emits DocReloadRequest: crates/language-server/src/
    functionality/handlers.rs:403, crates/language-server/src/functionality/
    handlers.rs:560
  - handle_doc_reload runs regenerate_doc_data_from_db on the worker pool and only
    discards stale results after completion; it does not cancel ongoing work: crates/
    language-server/src/functionality/handlers.rs:748
  - regenerate_doc_data_from_db rebuilds docs and SCIP for every ingot plus builtins:
    crates/fe/src/main.rs:888, crates/fe/src/main.rs:996
  - SCIP generation fans out through Rayon: crates/fe/src/scip_index.rs:263
  - span_to_scip_range rebuilds a full LineIndex for every span/occurrence: crates/fe/
    src/scip_index.rs:63
  - repeated File::url() calls hit Workspace::get_path, which clones the salsa paths
    input map before lookup: crates/common/src/file/workspace.rs:72

micahscopes:
## Summary

SCIP generation rebuilds the full reference index for all ingots (including core and std) on every edit. That's ~3s of work that produces the same output for builtins every single time.

Profiling shows 97% of that time is in `resolved_item_scope_targets` (building the reference index), not the SCIP formatting loop (30ms). Rayon parallelism over modules was speeding up the cheap part. This removes Rayon from SCIP gen and fixes the actual bottleneck with salsa caching.

## How it works

Three layers of salsa-tracked functions, each cached independently:

1. `module_references(db, module)` in hir. Per-module, only recomputes when that module's resolved targets change.
2. `ingot_references(db, ingot)` in semantic-indexing. Assembles per-module data into an ingot-level reference index. Builtins compute once and stay cached permanently.
3. `docs_for_ingot` / `generate_scip` read from cached layer 2 data. The formatting loop over modules takes 30ms.

`IngotContext::resolve` now calls `ingot_references` instead of `ReferenceIndex::build`, so every SCIP/LSIF generation path gets caching for free.

Cold start uses `salsa::par_map` to resolve all modules in parallel (1.8s vs 3.2s sequential for core). After that, edits only recompute the changed module.

## Other changes

- Moves SCIP/LSIF/doc extraction from the `fe` CLI binary into a new `fe-semantic-indexing` library crate so the CLI and language-server share one code path
- `ArcIndexMap` wraps `Workspace.paths` in an Arc so `File::url()` is O(1) instead of cloning the full IndexMap (showed up in Sean's flamegraph, 14/21 samples)
- `DocRegenerateFn` closure pattern replaced with direct function calls
- Criterion benchmarks for regression testing

Net +65 lines of library code. Rest of the diff is test fixtures, benchmarks, and file moves.

Closes #1424

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2049,6 +2049,7 @@ dependencies = [
  "fe-language-server",
  "fe-mir",
  "fe-resolver",
+ "fe-semantic-indexing",
  "fe-solc-runner",
  "fe-test-utils",
  "fe-web",
@@ -2081,7 +2082,10 @@ dependencies = [
  "criterion",
  "fe-common",
  "fe-driver",
+ "fe-hir",
  "fe-parser",
+ "fe-semantic-indexing",
+ "tempfile",
  "tracing",
  "url",
  "walkdir",
@@ -2231,6 +2235,7 @@ dependencies = [
  "camino",
  "clap",
  "codespan-reporting",
+ "criterion",
  "dir-test",
  "fe-codegen",
  "fe-common",
@@ -2240,6 +2245,7 @@ dependencies = [
  "fe-mir",
  "fe-parser",
  "fe-resolver",
+ "fe-semantic-indexing",
  "fe-test-utils",
  "fe-web",
  "futures",
@@ -2316,6 +2322,26 @@ dependencies = [
  "tracing",
  "url",
  "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "fe-semantic-indexing"
+version = "26.1.0"
+dependencies = [
+ "camino",
+ "dir-test",
+ "fe-common",
+ "fe-driver",
+ "fe-hir",
+ "fe-test-utils",
+ "fe-web",
+ "rustc-hash 2.1.1",
+ "salsa",
+ "scip",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "url",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ test-utils = { path = "crates/test-utils", package = "fe-test-utils" }
 resolver = { path = "crates/resolver", package = "fe-resolver" }
 codegen = { path = "crates/codegen", package = "fe-codegen" }
 fe-web = { path = "crates/fe-web" }
+semantic-indexing = { path = "crates/semantic-indexing", package = "fe-semantic-indexing" }
 language-server = { path = "crates/language-server", package = "fe-language-server" }
 mir = { path = "crates/mir", package = "fe-mir" }
 cranelift-entity = "0.130"
--- a/crates/bench/Cargo.toml
+++ b/crates/bench/Cargo.toml
@@ -6,15 +6,23 @@ edition.workspace = true
 [dev-dependencies]
 criterion = "0.5"
 walkdir = "2.4"
+tempfile = "3.23"
 
 [dependencies]
 driver.workspace = true
 parser.workspace = true
 common.workspace = true
+hir.workspace = true
+semantic-indexing.workspace = true
 camino.workspace = true
 url.workspace = true
 tracing.workspace = true
+tempfile = "3.23"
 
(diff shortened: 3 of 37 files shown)


sbillig:
@codex review

sbillig:
Agent feedback (will do a human lookover later today):

  - High: Normal stdio LSP now regenerates docs + SCIP on every edit/save.
    run_stdio_server always creates and passes Some(doc_reload_tx) to the backend, even
    when combined is None (crates/language-server/src/lib.rs:56, crates/language-server/
    src/lib.rs:86). The new guards check backend.doc_reload_tx.is_some() on save/change
    (crates/language-server/src/functionality/handlers.rs:408, crates/language-server/
    src/functionality/handlers.rs:559), so ordinary editor interactions emit
    DocReloadRequest; handle_doc_reload then calls semantic_indexing::doc::regenerate,
    which includes SCIP generation (crates/language-server/src/functionality/
    handlers.rs:763). This reintroduces the “SCIP regen during normal editor use”
    behavior the PR is meant to avoid. The enable bit needs to be separate from the
    broadcast channel, or the channel should only be Some when the docs server/live docs
    are active.
  - Medium: fe doc no longer preserves git-root-relative source paths for ingot/workspace
    docs. generate_docs still computes git_root, but the ingot/workspace paths now call
    extract_workspace / extract_ingot without it (crates/fe/src/doc.rs:216, crates/fe/
    src/doc.rs:228, crates/fe/src/doc.rs:528). Those paths use docs_for_ingot, which
    constructs DocExtractor::new(db) with no root_path (crates/semantic-indexing/src/
    tracked.rs:20); without root_path, get_source_location falls back to the bare
    filename (crates/semantic-indexing/src/extract.rs:706). For nested files this changes
    display_file from e.g. ingots/app/src/lib.fe to lib.fe, which breaks source block
    data-file matching against SCIP file keys and loses useful source context in
    generated docs.

micahscopes:
from my agent:

**Re: High (DocReloadRequest in stdio mode)**

Fixed by only creating `doc_reload_tx` when the combined server is active (`lib.rs:79`). But even without the gate, this is no longer a performance concern. With salsa caching at every layer, a repeat `doc::regenerate` call with no changes hits cached results throughout:

- `docs_for_ingot(core)` / `docs_for_ingot(std)` return cached instantly
- `ingot_references(core)` returns cached instantly
- `generate_scip(core)` rebuilds the ref index from cache (the 30ms formatting loop), but the output is identical so downstream JSON serialization produces the same string

The 3s full regen that was happening before is now microseconds of salsa cache lookups. The gate is cleaner but the safety net is the caching itself.

**Re: Medium (display_file paths)**

Fixed. `extract_ingot` / `extract_workspace` use tracked functions (no `root_path`), then `generate_docs` does a post-processing pass to rewrite `display_file` relative to `git_root`. One code path for extraction (tracked, shared between CLI and LSP), one post-processing step for the CLI-only concern. Added `display_file_is_relative_for_ingot` test.

**Re: Medium (enrich_signatures)**

Fixed. `doc::regenerate` now calls `enrich_signatures_with_base` per ingot, with correct `base_url` handling (None for file:// ingots, Some for builtins). Added `regenerate_produces_scip_with_enriched_signatures` test.

PR #1431, opened 2026-05-07
Add StorageKey impls for all primitive integer types and bool

cburgdorf:
Previously only `u256` and `Address` could be used as `StorageMap` keys. With these impls, `StorageMap<K, V>` accepts any of u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, i256, isize, and bool — matching Solidity's `mapping(intN/uintN/bool => V)` slot derivation, since each `write_key` left-pads to 32 bytes via `WordRepr::to_word()` (zero-extending unsigned, sign-extending signed, encoding `bool` as 0/1).

Notable consumer: containers like `DoubleEndedQueue` that want to pack their indices into a single storage slot via `u128` keys — previously forced to `u256` keys for lack of this impl.

The new test fixture `storage_map_key_types.fe` exercises each key type by storing and reading values back through a contract method.

Note: `test_signed_keys` constructs negative i8/i32/i256 keys *inside* the contract instead of decoding them from calldata, sidestepping an unrelated Sonatina calldata-decoder bug for negative `intN` values where N < 256.


cburgdorf:
@codex review

sbillig (approved):

PR #1433, opened 2026-05-07
Lower named aggregate consts as const refs

sbillig:
This changes Fe’s Sonatina lowering so aggregate constants and read-only aggregate parameters are represented by the cheapest available runtime source, instead of eagerly materializing object-backed values or emitting Sonatina data regions.

The main motivation is large const arrays: Fe should emit const refs/const operations for normal const use, and only force explicit code-region data when the program observes a code region through `code_region_offset` / `code_region_len`.

This results in huge gas and bytecode savings on some test cases; particularly the rosetta poseidon implementation.


sbillig:
@codex review

PR #1434, opened 2026-05-07
Fix signed scalar ABI widening for Sonatina

### LINKED ISSUE #1432: Sonatina backend: negative `intN` (N < 256) ABI-encode/decode reverts on call boundary
### Summary
                                                                                                                                                                                                                    
  On the Sonatina backend, any contract entrypoint whose Solidity ABI calldata or return data carries a negative `intN` value with N < 256 reverts with empty revert data. Both the **decode** path (negative `intN`
   argument) and the **encode** path (negative `intN` return value) are affected. The Yul backend handles the same code without issue.
                                                                                                                                                                                                                    
  `int256` is not affected — only the sub-256 widths (`i8`, `i16`, `i32`, `i64`, `i128`).                                                                                                                           
   
  ### Minimal reproduction                                                                                                                                                                                          
                                                         
  ```fe                                                                                                                                                                                                             
  // src/lib.fe                                          
  use std::abi::sol

  msg M {
      #[selector = 1]
      InputOnly { k: i8 } -> u256,
      #[selector = 2]                                                                                                                                                                                               
      OutputOnly { dummy: u256 } -> i8,
  }                                                                                                                                                                                                                 
                                                         
  pub contract C {
      recv M {
          InputOnly { k } -> u256 {
              if k == 0 { 0 } else { 1 }                                                                                                                                                                            
          }
          OutputOnly { dummy } -> i8 {                                                                                                                                                                              
              -1                                         
          }
      }
  }

  #[test]
  fn t_input_only_neg() uses (evm: mut Evm) {                                                                                                                                                                       
      let c = evm.create2<C>(value: 0, args: (), salt: 0)
      let k: i8 = -1                                                                                                                                                                                                
      let r: u256 = evm.call(addr: c, gas: 100000, value: 0, message: M::InputOnly { k: k })                                                                                                                        
      assert(r == 1)
  }                                                                                                                                                                                                                 
                                                         
  #[test]
  fn t_output_only_neg() uses (evm: mut Evm) {                                                                                                                                                                      
      let c = evm.create2<C>(value: 0, args: (), salt: 1)
      let r: i8 = evm.call(addr: c, gas: 100000, value: 0, message: M::OutputOnly { dummy: 0 })                                                                                                                     
      assert(r == -1)                                                                                                                                                                                               
  }
  ```                                                                                                                                                                                                               
                                                         
  ```toml                                                                                                                                                                                                           
  # fe.toml
  [ingot]                                                                                                                                                                                                           
  name = "repro"                                         
  version = "0.1.0"                                                                                                                                                                                                 
  ```
                                                                                                                                                                                                                    
  ### Expected                                           

  Both tests pass.

  ### Actual

  ```
  $ fe test
  FAIL  t_input_only_neg                                                                                                                                                                                            
      Test reverted: 0x
  FAIL  t_output_only_neg                                                                                                                                                                                           
      Test reverted: 0x                                                                                                                                                                                             
   
  test result: FAILED. 0 passed; 2 failed                                                                                                                                                                           
  ```                                                    

  ```
  $ fe test --backend yul
  PASS  t_input_only_neg
  PASS  t_output_only_neg                                                                                                                                                                                           
   
  test result: ok. 2 passed; 0 failed                                                                                                                                                                               
  ```                                                    

  The revert data is the empty byte string `0x` — no Solidity `Panic(uint256)` selector, no `Error(string)` selector, no custom-error selector. This indicates a low-level codegen failure (e.g. invalid opcode,    
  untrapped explicit `revert(0, 0)` in compiled bytecode) rather than an explicit user-level panic.
                                                                                                                                                                                                                    
  ### What does *not* trigger it                         

  | Pattern | Sonatina |                                                                                                                                                                                            
  |---|---|
  | Positive `intN` (N < 256) as calldata input | works |                                                                                                                                                           
  | Positive `intN` (N < 256) as return value | works |                                                                                                                                                             
  | **Negative `intN`, N < 256, as calldata input** | **reverts** |
  | **Negative `intN`, N < 256, as return value** | **reverts** |                                                                                                                                                   
  | Negative `i256` as calldata input or return value | works |
  | `intN` (negative) constructed *inside* a contract method, not crossing the ABI boundary | works |                                                                                                               
                                                                                                                                                                                                                    
  ABI-decoded `recv` argument or an ABI-encoded `recv` return slot.                                                                                                                                                 
                                                                                                                                                                                                                    
  ### Width sweep                                                                                                                                                                                                   
                                                         
  Repro extended to all signed widths confirms the cutoff:
                                                                                                                                                                                                                    
  ```fe
  #[test] fn t_i8_neg()   ...  // FAIL                                                                                                                                                                              
  #[test] fn t_i16_neg()  ...  // FAIL                   
  #[test] fn t_i32_neg()  ...  // FAIL                                                                                                                                                                              
  #[test] fn t_i64_neg()  ...  // FAIL
  #[test] fn t_i128_neg() ...  // FAIL                                                                                                                                                                              
  #[test] fn t_i256_neg() ...  // PASS                   
  ```                                                                                                                                                                                                               
                                                         
  So the bug is consistent across `i8`–`i128` and absent for `i256`. The natural-shaped fix probably lives wherever Sonatina special-cases the sub-256 path that needs explicit `signextend` after / before the     
  32-byte ABI slot conversion.                                                                                                                                                                                      
                              
  ### Hypothesis                                                                                                                                                                                                    
                                                         
  Solidity ABI encodes signed `intN` (N < 256) as a 32-byte slot with **sign extension** — for `int8(-1)` the slot is `0xFF...FF` (all bits set), not `0x00...00FF`. Decoding requires masking down to the low N
  bits and then re-sign-extending if the value is to be used as an `iN` again, or treating the slot as already in `iN` form via a `signextend` opcode at the appropriate byte index.                                
                                                                                                                                                                                    
  Yul backend appears to emit the right `signextend`. Sonatina apparently either:                                                                                                                                   
  - Treats the 32-byte slot as zero-extended on decode and then an internal range check fails for negative values, or                                                                                               
  - Re-encodes the return value without sign-extending, producing a non-Solidity-conformant calldata layout that the test harness rejects.
                                                                                                                                                                                                                    
  The empty revert data is consistent with either an explicit `revert(0, 0)` from a synthesised range check, or an INVALID opcode emitted by Sonatina's codegen.                                                    
                                                                                                                                                                                                                    
  ### Practical impact                                                                                                                                                                                              
                                                                                                                                                                                                                    
  This blocks any contract API that exposes `intN` (N < 256) at its public boundary — anything modelled on Solidity `int8`/`int32`/`int64`/`int128` parameters or return types. Workarounds available today:
                                                                                                                                                                                                                    
  1. Use `i256` everywhere at the ABI boundary, narrow internally.
  2. Carry `intN` values as `u256` carriers across the boundary and re-construct the signed type inside the contract.                                                                                               
  3. Use `--backend yul`.                                                                                            
                                                                                                                                                                                                                    
  ### Encountered while                                                                                                                                                                                             
                                                                                                                                                                                                                    
  Adding `StorageKey` impls for all primitive integer types in [`#1431`](https://github.com/ethereum/fe/pull/1431). The first version of the test fixture exposed this by accepting negative `i8` keys via calldata;
   the workaround there was option (2) above — construct the signed key inside the contract from a `u256` carrier. See `crates/fe/tests/fixtures/fe_test/storage_map_key_types.fe` in that PR for a working example 
  of the workaround.                                                                                                                                                                                               
                                                                                                                                                                                                                    
  ### Environment                                        
                                                                                                                                                                                                                    
  - `fe 26.1.0`
  - Linux x86_64                                                                                                                                                                                                    
  - Sonatina backend, default options

sbillig:
Fixes #1432

### DIFF
--- a/crates/codegen/src/sonatina/lower_runtime.rs
+++ b/crates/codegen/src/sonatina/lower_runtime.rs
@@ -1110,8 +1110,12 @@ impl<'ctx, 'db, 'a> FunctionLowerer<'ctx, 'db, 'a> {
                 self.lower_binary(*op, lhs, rhs, &lhs_class)?
             }
             RExpr::Cast { value, to } => {
+                let signed = self
+                    .body
+                    .value_class(*value)
+                    .is_some_and(RuntimeClass::is_signed_scalar);
                 let value = self.local_value(*value)?;
-                self.cast_scalar(value, scalar_ty(to))?
+                self.cast_scalar_with_signedness(value, scalar_ty(to), signed)?
             }
             RExpr::ConstRef { region, .. } => {
                 let gv = self.module.lower_const_region(*region)?;
@@ -1439,7 +1443,7 @@ impl<'ctx, 'db, 'a> FunctionLowerer<'ctx, 'db, 'a> {
                 let rhs = self.local_value(*rhs)?;
                 let lhs = self.cast_scalar(lhs, scalar_ty(class))?;
                 let rhs = self.cast_scalar(rhs, scalar_ty(class))?;
-                let signed = matches!(class.repr, ScalarRepr::Int { signed: true, .. });
+                let signed = class.is_signed_int();
                 match (op, signed) {
                     (SaturatingBinOp::Add, true) => self.fb.insert_saddsat(lhs, rhs),
                     (SaturatingBinOp::Add, false) => self.fb.insert_uaddsat(lhs, rhs),
@@ -1792,7 +1796,7 @@ impl<'ctx, 'db, 'a> FunctionLowerer<'ctx, 'db, 'a> {
             RuntimeIntrinsic::GenericSaturating { op, ty } => {
                 let prim = intrinsic_prim_from_ty(self.module.db, ty)?;
                 let op_ty = intrinsic_value_type(prim);
-                let signed = prim_is_signed(prim);
+                let signed = prim.is_signed_int();
                 let (lhs, rhs) = intrinsic_binary_args(self, args)?;
                 let lhs =
                     lower_intrinsic_operand(&mut self.fb, self.module.inst_set(), lhs, prim, op_ty);
@@ -1821,7 +1825,7 @@ impl<'ctx, 'db, 'a> FunctionLowerer<'ctx, 'db, 'a> {
         args: &[RLocalId],
     ) -> Result<ValueId, LowerError> {
         let op_ty = intrinsic_value_type(prim);
-        let signed = prim_is_signed(prim);
+        let signed = prim.is_signed_int();
         Ok(match op {
             NumericIntrinsicOp::Eq
             | NumericIntrinsicOp::Ne
@@ -3509,7 +3513,11 @@ impl<'ctx, 'db, 'a> FunctionLowerer<'ctx, 'db, 'a> {
         src: ValueId,
     ) -> Result<(), LowerError> {
         let value = match class {
-            RuntimeClass::Scalar(scalar) => self.cast_scalar(src, scalar_word_ty(scalar)),
+            RuntimeClass::Scalar(scalar) => self.cast_scalar_with_signedness(
+                src,
+                scalar_word_ty(scalar),
+                scalar.is_signed_int(),
+            ),
             RuntimeClass::Ref {
                 kind:
                     RefKind::Provider {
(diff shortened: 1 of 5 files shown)


cburgdorf:
Fix confirmed.

PR #1437, opened 2026-05-08
core: add fixed `String<N>` primitives

### LINKED ISSUE #1435: Make `String<N>` ergonomic for compile-time + runtime composition (concat + length + bytes round-trip)
Fe’s `String<N>` is a fixed-size string type (max length `N` bytes). Its length is fixed at compile time and shorter strings are padded; it’s intended for things like token names/symbols and error messages.

## Problem / What’s missing

Today it’s hard to build “nice” `String<N>` values from pieces in both `const` contexts and at runtime:
- No first-class concatenation (`a + b, a.concat(b)`, etc.) for `String<N>`.
- No supported way to construct a `String<N>` from computed bytes (so even if you can compute concatenated bytes, you can’t round-trip back into `String<OUT>`).
- No way to get the effective (unpadded) length of a `String<N>` at runtime, which makes `concat` semantics unclear given padding.
- (Nice-to-have) `String<N>` doesn’t support `==`, which makes basic testing/validation awkward.

## Suggested minimal API surface

A small, well-defined set of primitives would enable ergonomic const + runtime patterns without needing a full “string library”:

1. Bytes round-trip
  
 - `pub const fn String<N>::as_bytes(self) -> [u8; N]`
 - `pub const fn String<N>::from_bytes(bytes: [u8; N]) -> String<N>`

2. Effective length

- `pub const fn String<N>::len(self) -> usize` (define precisely how padding is interpreted)

3. Concatenation

- `pub const fn concat<const A: usize, const B: usize, const OUT: usize>(a: String<A>, b: String<B>) -> String<OUT>`
  - define behavior when `OUT != A + B` (error vs truncate) and whether concat uses “effective length” vs raw padded bytes
 
4. (Optional) Equality
  - Implement `core::ops::Eq for String<N>` so users can `assert(a == b)`.

## Acceptance criteria

- Works in `const` contexts (e.g., `const SYMBOL: String<8> = ...`).
- Works at runtime (in normal `fn` code).
- Padding/length semantics are explicitly specified so `concat` results are predictable.

g-r-a-n-t:
Added core `String<N>` utilities for fixed-width byte round-trips, effective byte length, concatenation, and equality in const and runtime code.

closes #1435

### DIFF
--- a/crates/hir/src/analysis/ty/method_table.rs
+++ b/crates/hir/src/analysis/ty/method_table.rs
@@ -6,7 +6,8 @@ use salsa::Update;
 use super::{
     binder::Binder,
     canonical::Canonical,
-    ty_def::{TyBase, TyId, strip_derived_adt_layout_args},
+    const_ty::ConstTyId,
+    ty_def::{InvalidCause, TyBase, TyId, strip_derived_adt_layout_args},
     unify::UnificationTable,
 };
 use crate::analysis::{HirAnalysisDb, ty::ty_def::TyData};
@@ -19,8 +20,10 @@ pub(crate) fn collect_methods<'db>(
 ) -> MethodTable<'db> {
     let mut collector = MethodCollector::new(db, ingot);
 
-    let impls = ingot.all_impls(db);
-    collector.collect_impls(impls);
+    for (_, external) in ingot.resolved_external_ingots(db).iter() {
+        collector.collect_impls(external.all_impls(db));
+    }
+    collector.collect_impls(ingot.all_impls(db));
     collector.finalize()
 }
 
@@ -128,6 +131,7 @@ impl<'db> MethodBucket<'db> {
     ) -> Vec<CallableDef<'db>> {
         let mut methods = vec![];
         let ty = strip_derived_adt_layout_args(table.db, ty);
+        let ty = saturate_ty_for_method_probe(table.db, ty);
         for (&cand_ty, funcs) in self.methods.iter() {
             let snapshot = table.snapshot();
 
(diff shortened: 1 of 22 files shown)


crates/hir/src/analysis/ty/method_table.rs
```diff
@@ -6,7 +6,8 @@ use salsa::Update;
 use super::{
```
cburgdorf, line 6: Might be worth to ask the agent to separate these changes into its own commit just to keep the history a bit cleaner.

cburgdorf (commented):
This seems to be a significant finding:

/tmp/fe-pr-1437.ntvyMt/ingots/core/src/string.fe:12: String<N>::as_bytes() exposes intrinsic::__as_bytes as a runtime-usable public method, but that intrinsic currently appears to have CTFE support only. A
    real runtime path with a String<N> parameter fails codegen with invalid runtime package: InvalidReturnClass. Since len() and concat() both call as_bytes(), this breaks the advertised runtime utilities for
    non-constant values. Minimal repro I used:

```
    fn fifth_byte(s: String<8>) -> u8 {
        let bytes: [u8; 8] = s.as_bytes()
        bytes[4]
    }
```

Running fe test on that helper failed in Sonatina emission before tests ran. The existing fixture passes because it only exercises literals/local constants, not a true runtime String<N> input. This needs
    runtime lowering for __as_bytes or an implementation based on runtime-supported word operations, plus a fixture with a String<N> parameter/storage value.

sbillig:
> String::as_bytes() exposes intrinsic::__as_bytes as a runtime-usable public method, but that intrinsic currently appears to have CTFE support only.

Yeah, that intrinsic was added to make `const fn sol` work, and it's currently CTFE-only. We should definitely have the ability to convert types to bytes at runtime; I'm not sure what the best way to do this is.

g-r-a-n-t:
@codex review

cburgdorf (approved):
Generally looks good but the issue with `as_bytes` still remains which can lead to an ICE.

```
use core::AsBytes
  fn via_trait<T: AsBytes>(_ t: T) -> [u8; T::N] { t.as_bytes() }
```

```
Failed to emit test Sonatina: invalid runtime package: InvalidReturnClass
  ERROR [2.7294s] Failed to emit test Sonatina: invalid runtime package: InvalidReturnClass
```

However, I just realized this is already the case in current master so this PR doesn't introduce the ICE. 

I think this can be merged and we just file the ICE as a separate issue.

PR #1438, opened 2026-05-08
Remove yul backend

sbillig:
Maintenance headache.


ingots/std/src/evm/effects.fe
```diff
@@ -750,8 +750,9 @@ impl ContractHost for Evm {
     fn create_contract<F>(mut self, _ runtime: F) -> ! {
         let len = self.code_region_len(runtime)
         let offset = self.code_region_offset(runtime)
-        self.codecopy(dest: 0, offset: offset, len: len)
-        self.return_data(offset: 0, len: len)
+        let ptr = mem::alloc(len)
+        self.codecopy(dest: ptr, offset: offset, len: len)
+        self.return_data(offset: ptr, len: len)
```
sbillig, line 755: Writing to 0 offset here isn't safe, as it can clobber memory that sonatina might reference after the codecopy op. This was revealed when a test case that was only running under yul was moved to sonatina.

g-r-a-n-t (approved):

PR #1440, opened 2026-05-12
abi: optimize dynamic payload copies and encode fast paths

g-r-a-n-t:
## Summary

This PR pulls in the first four ABI optimizations from #1422:

  - Optimizes dynamic ABI payload handling for strings, bytes, and dynamic arrays.
  - Optimizes Solidity custom integer canonical checks.
  - Lowers core `mcopy` as a MIR/codegen intrinsic and uses it for ABI memory payload copies.
  - Adds a `DIRECT_ENCODE` `encode_to_ptr` fast path for static ABI fields.


sbillig:
Issue with the first commit; from the agent:

The issue is in the new `dyn_array_payload_end` fast path.

For a dynamic array of dynamic elements, like `bytes[]`, the ABI payload shape is:

```text
array payload:
  length
  element head 0: offset to element 0 payload
  element head 1: offset to element 1 payload
  ...
  element 0 payload: length + bytes
  element 1 payload: length + bytes
```

The old code scanned every element and used the maximum element end. The new fast path says: if element offsets are word-aligned and monotonic, only inspect the last element, because the last element should end last.

That last sentence is only true for strict/canonical ABI. It is not guaranteed by monotonic offsets alone.

Example for `bytes[]` with two elements:

```text
length = 2
offset0 = 64
offset1 = 96
element0 length = 96
element1 length = 0
```

Offsets are aligned and monotonic, so the fast path accepts them. But element 0 ends after element 1. The new code computes the total copied payload size from element 1, so it copies too little. The decoded `Bytes` value for element 0 can then have `len = 96` while only part of its backing bytes were copied into memory. Later reads can observe zeros or unrelated memory instead of the input payload.

Can correct user code produce this? A normal ABI encoder should not. Canonical encoding places each dynamic tail after the previous dynamic tail, so an earlier element cannot extend beyond a later element. So this is not a bug for well-formed encoder output.

Is it still a decoding bug? Yes, unless Fe explicitly requires strict ABI input and rejects all non-strict encodings. Public calldata is adversarial, and this decoder already has checks for malformed inputs. If an input is non-strict, the decoder should either reject it or decode it consistently. This change does neither: it accepts a class of malformed-but-bounds-valid calldata and constructs a truncated in-memory value. So it’s not hypothetical in the sense of unreachable; it is reachable with crafted calldata. The right fix is either full scanning, or a stricter fast path that verifies each element’s computed end is `<=` the next element offset, reverting on overlap.

sbillig (approved):

PR #1443, opened 2026-05-14
sonatina: support gasprice builtin

### LINKED ISSUE #1442: gasprice
Slipped through the cracks:
```rust
            RuntimeBuiltin::GasPrice => {
                return Err(LowerError::Unsupported(
                    "gasprice is not supported by the Sonatina backend".to_string(),
                ));
            }
```
Someone should also verify that all other ops are supported.

g-r-a-n-t:
Fixes #1442 by adding Sonatina backend lowering for `gasprice` (`RuntimeBuiltin::GasPrice`), emitting Sonatina IR `evm_gas_price` instead of returning `unsupported`.

## Changes

- Lower `RuntimeBuiltin::GasPrice` via Sonatina `EvmGasPrice` (`crates/codegen/src/sonatina/lower_runtime.rs`).
- Add regression test asserting `evm_gas_price` appears in emitted Sonatina IR (`crates/codegen/tests/sonatina_gasprice.rs`).

## Notes

There are other `LowerError::Unsupported` cases in Sonatina lowering, but they fall into broader categories (const-global encoding gaps, provider/object-ref vs raw-pointer/address-space constraints, and general place/copy/store lowering limitations).

### DIFF
--- a/crates/codegen/tests/sonatina_gasprice.rs
+++ b/crates/codegen/tests/sonatina_gasprice.rs
@@ -1,0 +1,63 @@
+use common::{InputDb, ingot::IngotBaseUrl, stdlib::BUILTIN_STD_BASE_URL};
+use driver::DriverDataBase;
+use fe_codegen::emit_module_sonatina_ir;
+use url::Url;
+
+#[test]
+fn sonatina_lower_runtime_supports_gasprice() {
+    let mut db = DriverDataBase::default();
+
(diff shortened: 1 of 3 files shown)


crates/codegen/tests/sonatina_gasprice.rs
sbillig, line 1: I don't think we need these tests.

sbillig (approved):

PR #1448, opened 2026-05-16
abi decode gas optimization

sbillig:

PR #1451, opened 2026-05-17
Optimization: type-aware recv codegen, zero-copy views, #[abi] structs

micahscopes:
## Summary

Gas optimizations for the ABI decode path and contract entry points. The builder now uses CTFE to query trait constants and emit specialized code per msg type.

These optimizations will need a total refactor once compile-time reflection + builder rework land (stay tuned).

**Benchmark results (rosetta-fe, O2):**

| Contract | Function | Before | After | Delta | vs Solidity |
|---|---|---|---|---|---|
| Math | mulDiv_simple | 6483 | 5682 | -12.4% | -0.3% (beats) |
| AMM | getK | 9894 | 9658 | -2.4% | -0.7% (beats) |
| Merkle | computeRoot | 21602 | 10513 | -51.3% | +3.9% |
| Merkle | verify | 38393 | 16099 | -58.1% | +7.0% |
| Gov | execute | 32856 | 32350 | -1.5% | -4.2% (beats) |

## Changes

- **Type-aware recv wrappers**: builder queries `AbiSize::HEAD_SIZE` and `Encode::DIRECT_ENCODE` via CTFE. Static-only msg types get direct calldataload instead of malloc + decode
- **View types**: `ArrayView<T, N>`, `BytesView`, `StringView` with default type params for zero-copy calldata access
- **Lazy field loading**: per-field strategy (Direct / Decode / LazyDynamicView) based on mutability analysis
- **`#[abi]` attribute**: auto-derives `AbiSize`, `Encode<Sol>`, `Decode<Sol>` for user structs
- **Binary-search dispatch**: auto-selected for contracts with >3 selectors
- **Type checker fix**: method generics on default-type-param structs were clobbered by the struct's default

## Notes

- Overlaps with #1448 (ABI decode optimization) in the recv/synthetic builder area. Will need coordination on merge order
- Const predicates prototype (`where { const_expr }`) included but separate from the gas work
- This work led into an ongoing detour on effect-driven derive functionality (generating trait impls via `uses` capabilities), which is making good strides and will land separately


sbillig:
@micahscopes following up from your comment on #1448:
> I'm working on similar stuff (seemingly complementary) in https://github.com/argotorg/fe/pull/1451. I think we should try and get this one here merged first since I'm waiting on the Reflect effect derive stuff to rework that branch.

Would it be useful/feasible to merge some of the things on this branch before the Reflect derive stuff is ready? ABI encode/decode is the main gas issue right now, so it would be nice to land any improvements we can.

micahscopes:
@sbillig I'm not so sure if it makes sense to split it up. We could pull out some of the view types but the parts that make those most useful (automatically) for abi encode/decode are fairly wrapped up in the ctfe code. That said, it's possible that we could get this whole thing in now and overhaul as soon as the new reflection-based derive is done, but I'm not sure that's worth it, the derive stuff is coming along really nicely.

(I'm currently consolidating the new codegen to be based around the desugaring and hir builder stuff used in the contract lowering and this is going going very well)

micahscopes:
C.c. @cburgdorf if you feel up for it take a look here?

PR #1454, opened 2026-05-18
mir: lower DynString literals to a single CODECOPY

cburgdorf:
The previous lowering allocated memory and then wrote the length word plus each 32-byte chunk with its own MSTORE (and ADD per chunk). Now the literal payload is packed into a `[u256; N]` const region whose word-packed layout exactly matches the runtime `[len_word || padded_data]` shape, and a single CODECOPY copies it from the code segment.

Adds a `RuntimeBuiltin::ConstRegionAddr { region }` that lowers to a `SymAddr` of the const region's data — accepted by EvmCodeCopy as the `code_offset` operand without going through the `ConstRef` rewrite pass.

For a 38-byte revert message the old path emitted ~3 MSTOREs + 2 PUSH32s + 2 ADDs per chunk; the new path emits one CODECOPY of 64 bytes plus the length-word + size-word constants needed for the DynString header. Larger messages benefit linearly: no per-word PUSH32/MSTORE/ADD overhead.

### DIFF
--- a/crates/codegen/src/sonatina/lower_runtime.rs
+++ b/crates/codegen/src/sonatina/lower_runtime.rs
@@ -95,7 +95,7 @@ struct ModuleLowerer<'db, 'a> {
     layout_names: FxHashMap<LayoutId<'db>, String>,
     const_globals: FxHashMap<ConstRegionId<'db>, GlobalVariableRef>,
     const_names: FxHashMap<ConstRegionId<'db>, String>,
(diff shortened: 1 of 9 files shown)


cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@sbillig codex is happy, are you?

cburgdorf:
This one was superseded by the macro approach

PR #1455, opened 2026-05-19
Metaprogramming stuff

micahscopes:
messy first draft


micahscopes:
c.c. @sbillig

micahscopes:
@g-r-a-n-t @cburgdorf I'd be curious to get feedback from ya'll too

cburgdorf:
Had an agent walk me through the PR. Just glanced over it but general direction looks promising to me. End goal would be support for reflection and users being able to come up with their own `#[derive(MyTrait)]` macros etc?

micahscopes:
@cburgdorf exactly, and also to be able to do the same and similar in std/core, potentially including in desugared fe code

PR #1456, opened 2026-05-19
Allow local-bound blanket impls of foreign traits

sbillig:
Allows defining blanket impls of foreign traits when the impl is bound by a local trait.

For example:
```rust
use core::ops::WrappingAdd

struct W {
    value: u256,
}
trait Marker {}
impl Marker for W {}

impl<T: Marker> WrappingAdd for T {
    fn wrapping_add(self, _ other: T) -> Self {
        self
    }
}
```

### DIFF
--- a/crates/fe/tests/fixtures/fe_test/local_bound_blanket_foreign_trait_impl.fe
+++ b/crates/fe/tests/fixtures/fe_test/local_bound_blanket_foreign_trait_impl.fe
@@ -1,0 +1,25 @@
+use core::ops::WrappingAdd
+
+trait Marker {}
+
+struct W {
+    value: u256,
+}
+
+impl Copy for W {}
+impl Marker for W {}
+
+impl<T: Marker> WrappingAdd for T {
+    fn wrapping_add(self, _ other: T) -> Self {
+        self
+    }
+}
+
+#[test]
+fn local_bound_blanket_foreign_trait_impl() {
+    let a = W { value: 11 }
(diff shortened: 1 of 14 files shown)


cburgdorf:
The snapshot at `local_bound_blanket_foreign_trait_impl_diagnostics.snap` shows the conflict is caught when both impls live in the same ingot. I'd like to confirm what happens in a cross-ingot scenario:

```
 
  // ingot A
  pub trait MarkerA {} 
  impl<T: MarkerA> core::ops::Ord for T { ... }
 
  // ingot B
  pub trait MarkerB {}
  impl<T: MarkerB> core::ops::Ord for T { ... }
 
  // ingot C (depends on A and B)
  struct X { ... }
  impl A::MarkerA for X {}
  impl B::MarkerB for X {}
 
``` 

Both A and B are individually consistent: only when C combines them does X end up with two competing Ord impls. Does the coherence checker fire here at C's compile, or does it only consider impls within a single ingot?

Basically, I would love to allow this feature but I am wondering why it would be safe for us to do this when Rust doesn't allow it :sweat_smile:

sbillig:
@cburgdorf Right you are! We'd need two more restrictions for this feature to work without chance of conflicting blanket impls:
- marker trait must be non-pub
- no marker trait impls for non-local types

sbillig:
@codex review

sbillig:
@cburgdorf Suitably constrained now.

PR #1465, opened 2026-05-22
Mir dedupe

sbillig:
- **Structural layout dedup**
  Drop source_ty from struct/array/enum layouts and name from enum variants, so structurally-identical layouts collapse to a single LayoutId. Variant selection is by index, so the dropped names were cosmetic. This is the bulk of the snapshot churn (layouts/variants now render as struct/N, variant_N, etc.).
- **Canonical resolved-trait-method instances**
  When a trait method resolves to a concrete impl, build a canonical ImplEnv (impl scope, single resolving witness, no caller assumptions) and cap method args at the impl's generic-param count, so the same resolved method dedupes across callers regardless of call-site context.

Also includes a small refactor (shared ImplEnv::for_resolved_trait_method constructor, simplified const_lowering_ty) and a regression fixture confirming trait default bodies reached through a generic param receiver still resolve under the canonical env.

Practical impact is faster compile times for large projects:
(fe test total runtime)
bountiful: 13.38s -> 4.61s
zk-kit: 58.08s -> 45.40s
chz: 17.37s -> 16.05s

### DIFF
--- a/crates/mir/src/runtime/lower/classify.rs
+++ b/crates/mir/src/runtime/lower/classify.rs
@@ -3420,12 +3420,12 @@ uses (slot: Slot<u256>)
                     layout.data(&db)
                 );
             };
-            assert_eq!(layout_data.source_ty, binding_ty);
             // The wrapped value plus the zero-sized `TSlot<bool>` lock field.
             assert_eq!(
                 layout_data.fields.len(),
                 2,
-                "guarded_balances Mutex layout should expose the wrapped value and lock fields for arm {arm_idx}"
+                "guarded_balances Mutex layout should expose the wrapped value and lock fields for arm {arm_idx}; binding_ty={}",
+                binding_ty.pretty_print(&db),
             );
 
             assert!(
--- a/crates/mir/src/runtime/lower/classify.rs
+++ b/crates/mir/src/runtime/lower/classify.rs
@@ -16,7 +16,9 @@ use hir::analysis::{
         corelib::runtime_builtin_func_kind,
         normalize::normalize_ty,
         provider::registered_root_providers,
-        trait_def::{TraitInstId, resolve_trait_method_instance},
+        trait_def::{
+            TraitInstId, complete_resolved_trait_method_args, resolve_trait_method_instance,
+        },
         trait_resolution::{PredicateListId, TraitSolveCx},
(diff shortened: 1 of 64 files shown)


sbillig:
Will merge this after 26.2. I think this will also require some inliner tweaks in sonatina to avoid gas regression.

crates/mir/src/runtime/lower/classify.rs
```diff
@@ -1982,27 +1984,23 @@ pub(crate) fn resolve_runtime_call_key<'db>(
...
         hir::analysis::semantic::EffectProviderSubst::empty(db),
         ImplEnv::new(
             db,
-            caller_key.impl_env(db).normalization_scope(db),
-            assumptions,
-            witnesses.into_iter().collect::<Vec<_>>(),
+            BodyOwner::Func(impl_func).scope(),
+            PredicateListId::empty_list(db),
+            vec![concrete_inst],
         ),
```
micahscopes: agent:

> One case I couldn't convince myself about with the canonical env: a param
> receiver resolved via an assumption implementor that lands in a trait default
> body — caller assumptions are dropped here, so provability inside the default
> body rests entirely on the callee's own bounds + the single witness. A small
> test for that shape would settle it. Everything concrete checks out.

micahscopes (commented):
agent:

> Tiny one: const_ref.rs stores the raw call-site `callable.trait_inst()` as the
> canonical witness while classify.rs uses the normalized `concrete_inst`, so the
> two paths can produce spelling-dependent key splits — missed dedupe only, not a
> correctness issue.

sbillig:
> agent:
> 
> > Tiny one: const_ref.rs stores the raw call-site `callable.trait_inst()` as the
> > canonical witness while classify.rs uses the normalized `concrete_inst`, so the
> > two paths can produce spelling-dependent key splits — missed dedupe only, not a
> > correctness issue.

I think this is fine, and changing it would maybe be less correct.

sbillig:
> Will merge this after 26.2. I think this will also require some inliner tweaks in sonatina to avoid gas regression.

No inliner tweaks required. This branch provides a small gas benefit on the rosetta suite relative to master, with no regressions.
```
  master:      6,805,266
  mir-dedupe:  6,794,563
  delta:         -10,703 (-0.16%)
```

PR #1469, opened 2026-05-23
Ptr

sbillig:
This PR adds first-class typed memory pointers to Fe and redesigns the core, ABI, and EVM memory APIs around explicit pointer, view, ownership, and capacity types.

The previous APIs represented memory addresses primarily as `u256` values or effect-handle wrappers. This made it difficult for the type system and borrow analysis to distinguish memory pointers from integer-backed handles, track pointer provenance, prevent pointers from escaping into storage, and express bounded memory operations.

## Language changes

- Adds typed memory pointer syntax: `*T`.
- Supports dereference reads and writes with `*ptr`.
- Supports pointer field access, equality, typed pointer casts, and pointer arithmetic helpers.
- Allows borrowing through pointer dereferences.
- Implements `Index` and `IndexMut` for pointer-backed arrays and memory arrays.
- Routes indexed augmented assignments such as `items[i] += value` through `IndexMut`.
- Treats a bare `*` at the start of a line as a dereference. Multiline multiplication must keep `*` on the preceding line.
- Keeps nested dereferences such as `* *ptr` lexically separate when formatting because `**` is the exponentiation token.

## Memory APIs

Adds `core::ptr` with:

- `*T`: a typed address in linear memory.
- `MemArray<T>`: a dynamically sized typed memory allocation.
- `MemSlice<T>`: a copyable, read-only bounded view.
- `MemSpan`: an alias for `MemSlice<u8>`.
- `MemBuffer`: an owned byte allocation with separate logical length and writable capacity.
- `FixedMemBuffer<N>`: a fixed-size, non-escaping allocation that exposes constant allocation sizes to backend scratch-memory planning.
- Allocation, casting, offset, copying, zeroing, slicing, and checked indexing helpers.

Memory-bearing EVM APIs now accept `MemSpan`, `MemBuffer`, or typed pointers instead of loose integer address/length pairs. This applies to hashing, logging, copying, return/revert data, contract creation, ABI encoding, and low-level calls.

Low-level call results retain both the amount written into the supplied output buffer and the complete returndata length. Higher-level call outcomes can preserve owned returndata for decoding or bubbling reverts.

## ABI changes

- `Encode<A>` now writes directly to `*u8`.
- Allocation-based encoding helpers return `MemBuffer`.
- Removes the cursor-based `AbiEncoder` abstraction.
- ABI inputs and outputs use bounded memory regions.
- Static contract returns use fixed, non-escaping buffers so the backend can place them in scratch memory.

## Pointer safety

The compiler now tracks pointer provenance through aggregates, projections, calls, control flow, and returns.

Pointer-bearing values cannot be written into persistent or transient storage, including through generic functions that become storage-specialized later.

Memory pointers implement `EffectHandle` directly. User-defined effect handles must use `*Target` and the memory address space; integer-backed handles remain reserved for compiler-provided core and standard-library address spaces.

## Parser and tooling

- Aligns the compiler parser, formatter, and Tree-sitter grammar on pointer syntax and the line-start `*` policy.
- Fixes Tree-sitter parsing of chained `||` conditions.
- Regenerates the bundled Tree-sitter WASM artifact.
- Makes enum-tag sizing portable across host pointer widths, including WASM/WASI builds.

## Breaking changes and migration

| Previous API | Replacement |
| --- | --- |
| `MemPtr<T>` | `*T` |
| `core::abi::MemoryInput` | `MemSpan` or another `ByteInput` |
| `std::evm::MemoryBytes` | `MemSpan` |
| `std::evm::mem::alloc` | `core::ptr` allocation helpers or `MemBuffer` |
| integer pointer/length EVM arguments | `MemSpan`, `MemBuffer`, or `*T` |
| `AbiEncoder` | `Encode::encode(..., *u8)` and allocation helpers returning `MemBuffer` |
| `Into<T>` | explicit conversion APIs |


sbillig:
On hold till after 26.2 release.

crates/hir/src/analysis/semantic/borrowck/canon.rs
```diff
@@ -121,58 +422,491 @@ pub(super) struct BorrowCanonCx<'a, 'db> {
...
+        expr: &NExpr<'db>,
+        origin: SemOrigin<'db>,
+    ) -> Result<(), SemanticBorrowDiagnostic<'db>> {
+        let Some(dst_ty) = self.body.local(dst).map(|local| local.ty) else {
+            return Ok(());
         };
-        let loans = match expr {
+        if dst_ty.as_borrow(self.db).is_some() || !is_pointer_bearing_type(self.db, dst_ty) {
+            return Ok(());
+        }
```
micahscopes: agent:

> If I'm reading this right, a call whose result isn't pointer-bearing exits here,
> and since summaries only describe the return value, callee writes through mut
> args never invalidate caller facts — `fn set(h: mut Holder, p: *u256) { h.ptr = p }`
> leaves the caller holding a stale strong fact, and the later mut/mut conflict on
> `*h.ptr` goes undetected (`strong_lookup`'s exact hit suppresses the Unknown
> fallback). Is havoc on arg-reachable slots planned? Similar staleness shows up
> with >32-elem arrays: a whole-array overwrite at `[Any]` doesn't clear an earlier
> `arr[5]` fact.

micahscopes (commented):
agent:

> One borrowck question without a single line to hang it on: callee analysis
> seeds pointer params as pairwise-disjoint pointees, but I couldn't find
> anywhere that rejects `f(p, p)` — no check at the call site, no diagnostic in
> the callee. Is the no-alias contract meant to be checked, modeled as may-alias,
> or documented as a soundness precondition?

ingots/core/src/ptr.fe
```diff
@@ -0,0 +1,144 @@
...
+    pub len: u256,
+}
+
+impl<T> Copy for *T {}
+impl<T> Copy for MemArray<T> {}
+
+impl<T> MemArray<T> {
+    #[arithmetic(unchecked)]
+    pub fn new_uninit(_ len: u256) -> Self {
+        Self::from_raw_parts(ptr: cast<u8, T>(alloc_raw(len * size_of<T>())), len: len)
```
micahscopes: agent:

> `len * size_of<T>()` wraps here (fn is `#[arithmetic(unchecked)]`), so len ≥ 2^251
> gives a tiny allocation while the stored len keeps the Index bounds check happy —
> OOB writes through the safe API. `byte_len()` is checked; probably wants a checked
> multiply here too.

micahscopes:
agent:

> One more from the same family as the call-havoc question: the Store handler
> early-returns when the *source* type isn't pointer-bearing (canon.rs:847-850),
> deciding from the value type alone that pointer slots can't be affected. With
> ptr-to-ptr casts (`core::ptr::cast` does this generically) the written region
> can overlap a pointer slot of a different type:
> `let raw = core::ptr::cast<Holder, u256>(h); *raw = 123` physically overwrites
> h.ptr while its strong fact stays live — and slot keys are typed-path-based, so
> the punned view never collides with the original's key either. Punning is
> misuse territory, but if the hardening goal is that laundering degrades to
> Unknown, a store through a cast-derived view should probably havoc the pointer
> slots under the written roots rather than keep stale precise facts.

micahscopes:
agent:

> Possible ICE on the lowering side: when a Deref projection hits a class with no
> deref_target() — i.e. `RawAddr { target: None }`, every pointer-to-scalar —
> both body.rs:4428-4431 and classify.rs:653-668 fall back to the *root local's*
> pointee type. For `struct S { p: *u256 }`, the place `s` + `[Field(p), Deref]`
> has a non-pointer root, so the fallback returns None and classify's
> `expect("invalid deref projection class")` panics; for `pp: **u256`, `**pp`
> reuses the first-level pointee for the second deref (right mload chain by
> accident, wrong class label). Might be unreachable depending on how HIR
> normalizes these shapes — no fixture covers either.

micahscopes:
agent:

> Two policy-level things. `ty_is_noesc` returns false for `*T` (and for
> aggregates containing only pointers), so noesc's check_store permits storing
> raw pointers / MemArray into storage and transient — memory addresses that
> persist past the transaction and dangle on the next one. If pointer-in-storage
> is intended (transient scratch?), worth stating; otherwise pointers should
> probably be noesc. Separately, the parser change making infix `*` terminate at
> a line break (parser/expr.rs:378-385) silently reparses previously-valid
> multi-line multiplications — `a` newline `* b` was `a * b`, now `a` followed by
> a deref statement. Was a corpus check done for that pattern?

sbillig:
@codex review

crates/codegen/tests/fixtures/alloc_self_assign.fe
```diff
@@ -1,11 +1,11 @@
-use std::evm::alloc
+use core::ptr
```
micahscopes, line 1: 👀

crates/codegen/tests/fixtures/alloc_self_assign.fe
```diff
@@ -1,11 +1,11 @@
...
-    let mut x: u256 = size
-    x = alloc(x)
+pub fn bump_self_assign(size: u256) -> *u8 {
+    let mut x = ptr::alloc_bytes(size)
+    x = ptr::alloc_bytes(64)
     x
 }
 
-pub fn main() -> u256 {
+pub fn main() -> *u8 {
```
micahscopes, line 9: nice

crates/codegen/tests/runtime_handle_preservation.rs
```diff
@@ -430,112 +431,224 @@ fn view_receiver_with_storage_aggregate_keeps_receiver_runtime_visible() {
...
     );
 }
 
 #[test]
-fn mem_ptr_from_raw_helpers_build_ordinary_values_in_rmir() {
+fn default_view_pointer_params_use_direct_pointer_transport() {
     with_runtime_package!(
-        "mem_ptr_from_raw_helpers_build_ordinary_values_in_rmir.fe",
-        include_str!("fixtures/raw_log_emit.fe").to_string(),
+        "default_view_pointer_params_use_direct_pointer_transport.fe",
```
micahscopes, line 513: not sure if you intended to move these fixtures inline, I know the agents like writing them this way

crates/hir/src/analysis/semantic/borrowck/ir.rs
```diff
@@ -577,13 +599,113 @@ pub struct BorrowTransform<'db> {
...
+    },
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum MemorySummaryItem<'db> {
+    ReturnPointer {
+        output: NSProjectionPath<'db>,
+        targets: Vec<MemorySummaryTarget<'db>>,
+    },
+    Access {
```
micahscopes, line 676: I assume you made it this way for future memory access capability APIs, I'm adding one little protective integration test to help keep this intact

micahscopes (approved):
This seems solid!

PR #1471, opened 2026-05-24
Add assert! pseudo-macro builtin

sbillig:
`assert!(bool_expr)`
`assert!(bool_expr, "something is wrong")`

Hardcoded internally to output solidity-compatible Error(string) revert data, for now. This will need to be made target-dependent in the future.

Diff linecount is mostly treesitter.


cburgdorf (approved):
> 2. The `assert` fn still exists in parallel to the new `assert!` macro. Is this on purpose?

Will probably remove, just want to check gas/bytecode impact (function call vs "macro"-inlined branch)

sbillig:
@cburgdorf the gas regression from the second commit (replacing `assert` with `assert!`) is due to `compute_deposit_data_root` no longer being interned by sonatina (I guess because the fn call was replaced with a branch, but not sure). We could force inline that; I also think we should make sonatina's inlining more aggressive.

PR #1473, opened 2026-05-25
Instrumentation etc

micahscopes:


micahscopes:
Gonna rework the downstream parts of this so that jsonl is yet another artifact, not a representational bottleneck. It's important for the debugging engine that we get fast incremental updates to the graph when the code changes.

PR #1475, opened 2026-05-25
fix(parser): allow multiple items per line in impl/trait/extern blocks

cburgdorf:
The tree-sitter grammar permits `impl Foo for Bar { fn a() {}  fn b() {} }`, but the main parser rejected it with "expected newline, `}`, doc comment or doc comment". The post-item check in `parse_fn_item_block` and `parse_trait_item_block` forced `newline_as_trivia=false` and required a `Newline` separator. Dropping that check lets the next loop iteration parse the following item (or break on `}`) without changing the behavior for the newline-separated case.

### DIFF
--- a/crates/parser/src/parser/item.rs
+++ b/crates/parser/src/parser/item.rs
@@ -1006,17 +1006,6 @@ fn parse_fn_item_block<S: TokenStream>(
 
         if is_fn_head {
             parser.parse_cp(FuncScope::new(fn_def_scope), checkpoint)?;
-
-            parser.set_newline_as_trivia(false);
-            parser.expect(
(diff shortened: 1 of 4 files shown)


sbillig:
Why?

cburgdorf:
Feels natural to write very short impls as concise one liners.

sbillig:
`trait Pair { type Lhs type Rhs const N: u32 fn pair(self) -> u32 }`

sbillig (approved):

sbillig:
> Feels natural to write very short impls as concise one liners.

Fine, I'll allow it

PR #1482, opened 2026-05-28
Storage packed array benchmark

cburgdorf:
Add benchmark for StoragePackedArray vs Solady's LibMap


sbillig (approved):
Minor nit: it would be nice if differential_storage_packed_array.rs and the deposit test shared code.

cburgdorf:
> Minor nit: it would be nice if differential_storage_packed_array.rs and the deposit test shared code.

Good point. I moved some helpers around

PR #1489, opened 2026-06-09
Avoid scalar slots for read-only place reads

sbillig:
We were lowering some read-only scalars to memory alloc/store/load. Sonatina optimized many of these away, but not all.


crates/fe/tests/differential_deposit/gas_report.snap
sbillig, line 1: nice impact on deposit gas and bytecode

cburgdorf (approved):
Looks solid! Nice win.

PR #1492, opened 2026-06-20
Fix contract field layout of types with nested inferred slot const params

sbillig:
This is mostly a rework of the "layout hole" (the underscore in `struct StorageMap<K, V, const SLOT: u256 = _>`) implementation; this fixes a hole aliasing bug where two distinct holes that share the same type and structural position were given the same identity and thus slot, eg `pair: (StorageMap<..>, StorageMap<..>)`.

Also adds missing trait diagnostics.


sbillig:
@codex review

PR #1494, opened 2026-06-22
Support struct values in storage maps

g-r-a-n-t:
Adds typed pointers to `StorageMap` entries, allowing maps to store and access multi-word values such as structs without requiring those values to implement `WordRepr`.

Pointer-based access uses the normal storage effect model. Existing `get` and `set` APIs remain available for single-word `WordRepr` values.

## API

  - Adds `StorageMap::ptr`, returning a read-only `StoragePtr<V>`.
  - Adds `StorageMap::mut_ptr`, returning a mutable `StorageMutPtr<V>`.
  - Requires mutable access to the map before creating a mutable entry pointer.
  - Implements the standard effect-handle and effect-reference traits for these pointers.
  - Keeps `StorageMap::get` and `StorageMap::set` restricted to `V: WordRepr`.

## Storage layout

The map entry hash is treated as the base storage slot for the value. Aggregate fields are projected relative to that slot using the compiler's storage layout.

This PR also defines consistent packed storage layout for nominal structs:

  - Adjacent scalar fields smaller than one word may share a storage slot.
  - Writes to shared slots use read-modify-write operations so neighboring fields are preserved.
  - Nested structs use their recursively computed packed storage span.
  - Tuples, arrays, and other aggregate transitions remain storage boundaries.
  - Signed and narrow scalar values are stored in their canonical bit representation.

HIR layout reporting and MIR/codegen now use the same shared field-layout planner. The same rules therefore apply to both ordinary contract struct fields and structs reached through storage-map entry pointers. Memory and ABI layouts are unchanged.

## Example

```fe
struct Position {
  owner: u256,
  collateral: u256,
  debt: u256,
}

struct Store {
   positions: StorageMap<u256, Position>,
}

let position = store.positions.ptr(key: account)
let mut position = store.positions.mut_ptr(key: account)
```

The read-only pointer can inspect the entry, while the mutable pointer can replace the value or update individual fields through normal storage-provider semantics.

### DIFF
--- a/crates/codegen/src/sonatina/lower_runtime.rs
+++ b/crates/codegen/src/sonatina/lower_runtime.rs
@@ -386,7 +386,15 @@ impl<'db, 'a> ModuleLowerer<'db, 'a> {
         if let Some(&existing) = self.type_cache.get(&layout) {
             return Ok(existing);
         }
-
+        // Storage policy does not change the in-memory representation of a layout.
+        if let Some(existing) = self.type_cache.iter().find_map(|(candidate, ty)| {
+            layout
+                .shares_runtime_rep_with(self.db, *candidate)
+                .then_some(*ty)
+        }) {
+            self.type_cache.insert(layout, existing);
+            return Ok(existing);
+        }
         let ty = match layout.data(self.db) {
             Layout::Struct(data) => {
                 let fields = data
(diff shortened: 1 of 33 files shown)


g-r-a-n-t:
@codex review

g-r-a-n-t:
@codex review

g-r-a-n-t:
@codex review

PR #1495, opened 2026-06-22
hir: reject nested storage map fields

g-r-a-n-t:


cburgdorf:
This doesn't feel right to me. Why should the hir even know about `StorageMap`. Shouldn't this be caught by a bound on `V` already? It should be impossible to instantiate a `StorageMap` where `V` does not satisfy the `WordRepr` trait.

`impl<K: StorageKey, V: WordRepr, const SALT: u256> StorageMap<K, V, SALT>`

But maybe what we should do instead is moving the bound from the `impl` to the struct:

`pub struct StorageMap<K: StorageKey, V: WordRepr, const SALT: u256 = _> {}`

I believe that would then create a compile error one level earlier when a user attempts to declare an invalid `StorageMap` rather than at the time they try to instantiate it.

What does @sbillig think?

g-r-a-n-t:
@codex review

PR #1496, opened 2026-06-23
Add tracing subscriber to tests

### LINKED ISSUE #1153: add tracing subscriber to tests
Some tests use `error!` etc, but no output is generated when a test fails due to there being no `tracing` subscriber.

KiChjang:
Fixes #1153.

This PR properly adds a tracing subscriber to all tests that needs to emit log events during testing.

The default behaviour of the subscriber is to read from `RUST_LOG` to determine the appropriate log level, and, failing that, defaults to the INFO log level. Successful test cases will discard the log captured by the subscriber; only test failures would be printed out to stdout/stderr depending on log level. This is done to reduce the noisy output from successful tests, but can still be instructed to output all logs by running `cargo test -- --show-output`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,10 @@ tracing-tree = "0.4.0"
 wasm-bindgen-test = "0.3"
 semver = "1.0.26"
 petgraph = "0.8"
-sonatina-ir = { git = "https://github.com/fe-lang/sonatina", rev = "a74b939" }
-sonatina-triple = { git = "https://github.com/fe-lang/sonatina", rev = "a74b939" }
-sonatina-codegen = { git = "https://github.com/fe-lang/sonatina", rev = "a74b939" }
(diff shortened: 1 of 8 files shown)


KiChjang:
@codex review

Cargo.toml
```diff
@@ -50,10 +50,10 @@ tracing-tree = "0.4.0"
 wasm-bindgen-test = "0.3"
 semver = "1.0.26"
 petgraph = "0.8"
-sonatina-ir = { git = "https://github.com/fe-lang/sonatina", rev = "a74b939" }
```
cburgdorf, line 53: What's the reason for these changes? They seem unrelated or am I missing something?

g-r-a-n-t: https://github.com/fe-lang/sonatina/pull/297

cburgdorf: Ah, that's why. Ok, I'll merge it.

PR #1497, opened 2026-06-23
parser: Recover instead of panicking on a missing = in a LetExpr

### LINKED ISSUE #1411: parser: don't panic when `if let <pattern>` isn't followed by an Eq
crates/parser/src/parser/expr.rs:234:

```
  parser.bump_expected(SyntaxKind::LetKw);
  parser.set_newline_as_trivia(false);
  parse_pat(parser)?;
  parser.bump_expected(SyntaxKind::Eq); // panics on extraneous syntax
```

KiChjang:
Fixes #1411.

This PR removes a compiler panic and replaces it with a recoverable error when a `=` token is missing during the parsing of a let expr node.


KiChjang:
@codex review

cburgdorf (approved):
Thanks! 🎉

PR #1501, opened 2026-06-26
Replace sol-compat int trait impl boilerplate with blanket impls

sbillig:
Replaces the per-type `Encode` / `AbiSize` / `AbiSpan` / `TopicValue` impls
for solidity int compatibility int types with four blanket impls gated on a private `SolInt` marker
(cuts ~1k lines from ints.fe)

Plus resolver improvements that make this feasible:

- **Drop inapplicable trait candidates before visibility check**
  In method selection, for a fully-known receiver, drop candidates whose impl
  is provably inapplicable (self-type mismatch or unsatisfiable where-clause)
  before the single-candidate / visibility / ambiguity decisions. This fixes a blanket 
  `impl<T: Marker> Trait for T` shadowing a concrete impl and
  turning an unambiguous `concrete.method()` into a spurious "import the trait"
  error

- **Hide unsatisfied-bound hint for non-visible traits**
  Suppresses the `trait bound X: Y is not satisfied` sub-diagnostic when `Y`
  isn't visible at the error site, so the private `SolInt` marker doesn't 
  leak into error messages.

### DIFF
--- a/crates/codegen/tests/fixtures/sonatina_ir/effect_ptr_domains.snap
+++ b/crates/codegen/tests/fixtures/sonatina_ir/effect_ptr_domains.snap
@@ -92,7 +92,7 @@ func private %bump__gb2d8(v0.objref<@layout_0>) {
         return;
 }
 
-func private %impl_trait_StorPtr__from_raw__gb2d8(v0.i256) -> i256 {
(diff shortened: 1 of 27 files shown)


sbillig:
@codex review

cburgdorf (approved):
Looks solid 🚢

PR #1504, opened 2026-06-27
Untrack generated treesitter files

sbillig:
`parser.c`, `grammar.json`, and `node-types.json` are now gitignored and
regenerated from `grammar.js`. The tracked sources are `grammar.js`,
`scanner.c`, `queries/`, and the bindings.

gh ci lint and test jobs install the pinned CLI and run
`tree-sitter generate` before building; `make test` does the same locally. The
parser is still covered by `tree_sitter_parse.rs`.

`fe-lang/tree-sitter-fe` has been revived, but it now holds the generated artifacts,
and the editor plugins now point that that repo instead of argotorg/fe.
A scheduled workflow in tree-sitter-fe pulls the fe repo, regenerates
the parser, and commits it: `main` follows `fe@master`, and each fe release tag
gets a matching tag and GitHub release. zed-fe, nvim-fe, and emacs-fe pin a release 
tag on tree-sitter-fe.

nvim-fe and emacs-fe currently compile `parser.c` from `fe@master`, so their
PRs that repoint them at the mirror need to merge before this does. EDIT: merged


micahscopes (approved):

micahscopes:
Thanks for taking care of this

PR #1507, opened 2026-06-28
Concrete `where Eq<T>` constraints

micahscopes:


sbillig:
This syntax seems confusing to me, at least on its own, in this limited form. `Eq` is a generic trait with one param (aside from the hidden `Self` param), so when we write eg `impl Eq<Foo> for Bar`, `Foo` is the non-Self param. `T` in `Eq<T>` shouldn't be `Self` param in some contexts and the "other" param in other contexts.

We could write `impl Eq<Bar, Foo>` instead of `impl Eq<Foo> for Bar`, but it would probably be less confusing if we made it less rust-like.

I'm also not clear why being able to write `where Sometrait<T>` as an alternative to `where T: Sometrait` unlocks (presumably this syntax makes sense in the context of the first-class obligations).

micahscopes:
ah, yeah this whole PR here is a mistaken fluke

PR #1509, opened 2026-07-07
stdlib: Add isqrt (integer square root) to core::num

cburgdorf:
Returns floor(sqrt(x)) for u256, matching Vyper's isqrt builtin and OpenZeppelin's Math.sqrt: seed with 1 << (log2(x) / 2) via branchy bit-length estimate, run exactly seven Newton iterations, then round down with min(z, x / z). Deterministic gas, no input-dependent loops.

### DIFF
--- a/ingots/core/src/num.fe
+++ b/ingots/core/src/num.fe
@@ -919,6 +919,61 @@ pub const fn sign_extend(word: u256, _ mask: u256, _ sign_bit: u256) -> u256 {
     }
 }
 
+/// Returns floor(sqrt(x)), i.e. the largest `z` such that `z * z <= x`.
+#[arithmetic(unchecked)]
+pub fn isqrt(_ x: u256) -> u256 {
+    if x == 0 {
+        return 0
+    }
+
+    // Seed with `1 << (log2(x) / 2)`, a power of two within a factor of two
+    // of the true root, so a fixed number of Newton iterations suffices.
+    let mut y = x
+    let mut z: u256 = 1
+    if y >= 1 << 128 {
+        y >>= 128
+        z <<= 64
+    }
+    if y >= 1 << 64 {
+        y >>= 64
+        z <<= 32
+    }
+    if y >= 1 << 32 {
+        y >>= 32
+        z <<= 16
+    }
+    if y >= 1 << 16 {
+        y >>= 16
+        z <<= 8
+    }
+    if y >= 1 << 8 {
+        y >>= 8
+        z <<= 4
+    }
+    if y >= 1 << 4 {
+        y >>= 4
+        z <<= 2
+    }
+    if y >= 1 << 2 {
+        z <<= 1
+    }
+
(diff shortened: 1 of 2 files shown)


ingots/core/src/num.fe
```diff
@@ -919,6 +919,60 @@ pub const fn sign_extend(word: u256, _ mask: u256, _ sign_bit: u256) -> u256 {
...
+
+    // Newton's method doubles the number of correct bits per step, so seven
+    // iterations refine the seed past the 128 bits a root can occupy.
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
+    z = (z + x / z) / 2
```
sbillig: These should be `>> 1` instead of `/ 2` (saves 2 gas each). I think sonatina will make this transformation, but not on -O0. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol

ingots/core/src/num.fe
```diff
@@ -919,6 +919,60 @@ pub const fn sign_extend(word: u256, _ mask: u256, _ sign_bit: u256) -> u256 {
     }
 }
 
+/// Returns floor(sqrt(x)), i.e. the largest `z` such that `z * z <= x`.
+pub fn isqrt(_ x: u256) -> u256 {
```
sbillig, line 924: Should be `#[arithmetic(unchecked)]`; there's nothing unsafe here, and openzep's implementation is in an unchecked block.

ingots/core/src/num.fe
cburgdorf: Alright!

cburgdorf:
@sbillig resolved both issues

ingots/core/src/num.fe
```diff
@@ -919,6 +919,61 @@ pub const fn sign_extend(word: u256, _ mask: u256, _ sign_bit: u256) -> u256 {
...
+    z = (z + x / z) >> 1
+    z = (z + x / z) >> 1
+    z = (z + x / z) >> 1
+    z = (z + x / z) >> 1
+    z = (z + x / z) >> 1
+
+    // Newton converges from above and can settle on `floor + 1`; take the
+    // smaller of `z` and `x / z` to round down to the floor.
+    let q = x / z
+    if z < q { z } else { q }
```
sbillig: Sorry, didn't notice this before. Should be
    let q = x / z
    z - ((z > q) as u256)

(from oz: `return xn - SafeCast.toUint(xn > a / xn)`)

Saves 20some gas.

sbillig (approved):

ingots/core/src/num.fe
cburgdorf: aye aye sir!

PR #1511, opened 2026-07-07
std: expose encode_msg_calldata and add Address::static convenience

cburgdorf:
Two ergonomic additions to std::evm for code that drops down to raw_call/raw_staticcall (e.g. tolerant ERC20 transfers):

- encode_msg_calldata<M>: public helper that ABI-encodes a msg variant (4-byte selector + payload) into fresh memory without performing a call, returning (offset, length) for Call::raw_call/raw_staticcall. Delegates to the existing private encode_calldata.
- Address::static<M>: STATICCALL sibling of Address::call, forwarding all gas and bubbling the callee's revert data on failure.

Fixture tests cover encoding equivalence against the typed call path (static args and dynamic Bytes, via a keccak-of-calldata echo callee), Address::static happy paths returning u256 and Address, and revert-data bubbling observed through a relay contract. These also add the first fixture coverage for the typed Call::static path.

### DIFF
--- a/crates/fe/tests/fixtures/fe_test/address_static_method.fe
+++ b/crates/fe/tests/fixtures/fe_test/address_static_method.fe
@@ -1,0 +1,80 @@
+/// Test the `Address::static` convenience method: typed STATICCALLs that
+/// forward all gas, plus revert-data bubbling on failure.
+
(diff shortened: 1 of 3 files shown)


sbillig (approved):
👍

Perhaps we should add a `try_call` method, as an alternative to dropping into  `raw_call`.

PR #1512, opened 2026-07-08
std: add evm::packed — tightly packed byte encoding (abi.encodePacked…

cburgdorf:
std: add evm::packed — tightly packed byte encoding (abi.encodePacked parity)
Adds std::evm::packed with two layers:

- A declarative tuple API: an EncodePacked trait with encode_packed /
  keccak_packed entry points, implemented for u8-u256, i8-i256 (two's
  complement), bool, Address, String<N> (effective len() bytes),
  [u8; N], FixedBytes<N>, Bytes, BytesView<MemoryInput>, and tuples of
  arity 1..=16, so `keccak_packed((token0, token1))` covers CREATE2
  salts and EIP-712 digests with zero size bookkeeping.
- An imperative Packed builder for runtime-shaped payloads, with
  byte-precise writes (full-word mstore fast path only when a whole
  word fits below the buffer limit, mstore8 fallback otherwise) and
  copy-on-grow doubling, plus keccak256/sha256/as_bytes terminals.

Also adds crypto::eip712_digest built on the tuple API.

Fixes a codegen bug this exposed: a tuple/struct const region holding
a String<N> literal of 16 bytes or fewer crashed Sonatina emission
("global initializer immediate type does not match global type") —
the aggregate field type is always I256 while the immediate was typed
from the literal's byte length. gv_initializer_for_const now threads
the layout's per-field RuntimeClass down to immediate_for_const, which
types FixedBytes immediates from the declared class width. As a side
effect, string constants emit as direct .i256 immediates instead of a
small immediate plus zext (two sonatina_ir snapshots updated).

The packed_encoding fixture's expected values are all precomputed with
Foundry (cast abi-encode --packed / cast keccak), so the tests are not
their own oracle. string_in_tuple is a minimal regression fixture for
the codegen fix.

### DIFF
--- a/crates/codegen/src/sonatina/lower_runtime.rs
+++ b/crates/codegen/src/sonatina/lower_runtime.rs
@@ -225,7 +225,7 @@ impl<'db, 'a> ModuleLowerer<'db, 'a> {
         }
 
         let ty = self.ty_for_layout(region.layout(self.db))?;
-        let init = self.gv_initializer_for_const(region.value(self.db).clone())?;
+        let init = self.gv_initializer_for_const(region.value(self.db).clone(), None)?;
         let name = self.const_name(region);
         let gv = self.builder.declare_gv(GlobalVariableData::constant(
             name,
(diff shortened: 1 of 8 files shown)


cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@sbillig ready to review now. There's a codegen fix in a separate commit

sbillig (approved):

PR #1513, opened 2026-07-08
Add fe build --from-metadata to rebuild from a metadata.json input

cburgdorf:
`fe build --emit metadata` produces a complete, deterministic recompilation input, but the reverse step existed only as a test helper. External toolchains (foundry-compilers, Sourcify-style verifiers) need it as a CLI feature: one invocation, JSON in, artifacts out.

`fe build --from-metadata <PATH|->` (`-` reads stdin) materializes the recorded project (regenerated fe.toml per settings.ingots[] entry, sources routed by namespace) into a temporary directory and runs the normal ingot build on it:

- default target is the settings.compilationTarget contract; --contract overrides
- optimizer level comes from settings.optimizer.level; an explicit -O wins with a stderr warning that the bytecode deviates
- compiler.version mismatch warns on stderr without aborting
- artifacts follow --emit into --out-dir, defaulting to ./out
- invalid input (bad JSON, language != "Fe", missing sources/ingots) exits 1; combining with [path], --ingot, --standalone, or --report is a clap conflict (exit 2)

The reconstruction lives in crates/fe/src/metadata_input.rs and validates untrusted input (no path traversal out of the temp dir). The round-trip tests now exercise the CLI feature directly; the test helper is removed. New tests cover stdin, target selection, the -O and version warnings, and the error cases.


cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
@codex review

cburgdorf:
This should have shipped with Fe 26.2 already but wasn't ready in time. This will be needed for deeper Sourcify integration as well as for foundry.

sbillig:
5.6 feedback:

  crates/fe/src/build.rs:1082 silently treats an unresolved
  settings.compilationTarget path as though no path was requested, then falls back to ingot-wide
  contract-name resolution.

  Confirmed repro: metadata specifies src/missing.fe: Foo, while Foo exists only in src/
  actual.fe; rebuilding succeeds and emits Foo.runtime.bin from the wrong source.

  Required fix:

  - Validate that compilationTarget contains exactly one string path/name pair.
  - When using the recorded target, error unless the path exists and that module defines the
    named contract.

  - Never fall back when a recorded path was supplied.
  - Add failure tests for both a missing path and a path whose module lacks the recorded
    contract.

cburgdorf:
@codex review

cburgdorf:
@sbillig Thanks! Feedback is addressed.

sbillig (approved):

PR #1514, opened 2026-07-08
std: add call_with_default and std::evm::erc20 safe-transfer helpers

cburgdorf:
Typed external calls strictly decode returndata, so a `-> bool` message reverts when a token (USDT-style) executes but returns no data. Add two opt-in layers; the strict `call<M>` path is untouched:

- `Call::call_with_default<M>` (+ `Address::call_with_default`): identical to `call<M>` except that a successful call with EMPTY returndata yields a caller-supplied default instead of a decode revert. Reverts still bubble verbatim and non-empty returndata is still decoded strictly. Caveat (as in Vyper's default_return_value): a call to a codeless address yields the default.

- `std::evm::erc20` with `safe_transfer` / `safe_transfer_from` / `safe_approve` (OpenZeppelin SafeERC20 semantics): tolerates empty returndata, reverts on a `false` return, and closes the no-code hole with an extcodesize check performed only on the empty-returndata path.

Fixture tests cover the full mock matrix for both layers: compliant, reverting (revert data asserted verbatim), no-return, false-returning, malformed (<32 bytes) returndata, and no-code targets, plus a non-bool return type exercising both the empty and non-empty paths.

### DIFF
--- a/crates/fe/tests/fixtures/fe_test/call_with_default.fe
+++ b/crates/fe/tests/fixtures/fe_test/call_with_default.fe
@@ -1,0 +1,265 @@
+/// Tests for `call_with_default`: like `call<M>`/`Address::call`, but a call
+/// that succeeds with EMPTY returndata yields a caller-supplied default
+/// instead of a decode revert. Reverts still bubble and non-empty returndata
+/// is still decoded strictly.
+
+use std::evm::{RawMem, RawOps, last_returndata, mem}
+
+/// Strict caller-side view of an ERC20-style `transfer`.
+msg TokenMsg {
+    #[selector = sol("transfer(address,uint256)")]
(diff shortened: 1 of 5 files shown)


cburgdorf:
I'm not quite sure if we really want the erc20 stuff in the std lib or if  that should be in an external ingot. But I think the `call_with_default` is a good API to deal with the different ways that a call might return.

sbillig:
> I'm not quite sure if we really want the erc20 stuff in the std lib or if that should be in an external ingot

For now I think std should have rich support for all common interfaces. We can optionally kick it out to an external ingot later.

cburgdorf:
@sbillig sounds good. Then, this should be ready for human review.

sbillig (approved):

PR #1515, opened 2026-07-09
Layout root graph

sbillig:
Replaces layout-hole handling type/provenance-based layout hole (aka const hole, eg `const SALT: usize = _` ) inference to an explicit layout-root graph and layout-evidence ABI.

Each source layout parameter now has a stable identity, and each structural occurrence receives its own
slot assignment. Evidence is carried through HIR and MIR instead of being reconstructed later from types or provenance.

## Bug fixes

- Fixed issue where repeated instances of a hole-bearing type within a single contract field would be assigned the same slot number. Eg `maps: (StorageMap<u256, u256>, StorageMap<u256, u256>)`.
- Reserves explicit layout parameters before assigning inferred ones. Mixed layouts such as an explicit `SALT = 1` followed by inferred maps can no longer collide, while intentionally duplicated explicit parameters still
share.
- Correctly lays out static arrays of hole-bearing types, including nested and multidimensional arrays, using
checked indexed root families.
- Preserves runtime bounds checks for dynamically indexed layout-only or zero-sized array elements.
- Rejects unresolved, ambiguous, overflowing, cyclic, or otherwise invalid layouts atomically instead of
producing provisional or aliased layouts.

## Storage layout LSP hover

Contract-name hover now shows the complete contract layout, grouped under Storage, Transient Storage, Immutable (code). Example:
```rust
struct CounterStore { global: u256, counts: StorageMap<Address, u256> }

// The contract itself: declares its storage and handles messages.
pub contract Counter uses (ctx: Ctx) {
    mut store: CounterStore,
    mut table: [[u256; 2]; 4],
    mut baz: Mutex<StorageMap<u256, u256, 0>,
```
shows:
```
Storage
  0: baz.value.SALT (explicit parameter, u256)
  1: store.global (inline field, u256)
  2: store.counts.SALT (inferred parameter, u256)
  3 + i0*2 + i1 (i0: 0..4, i1: 0..2): table[i0][i1] (inline field, u256)
  
Transient Storage
  0: baz.lock.SLOT (inferred parameter, u256)
```

Field hover shows the entries belonging to that field. Array layouts include their index formula and
dimensions, and invalid or blocked layouts explain why no assignment is available.


sbillig:
@codex review

PR #1520, opened 2026-07-17
Report msg selector/field ABI type mismatches at declaration time

cburgdorf:
Previously, a msg variant whose field type was wire-compatible but semantically different from the type named in its sol("...") selector literal (e.g. a Bytes field under a string selector arg) compiled to full bytecode and only failed during ABI JSON emission, as a contract-level error with no source span. A plain `fe build` without `--emit abi` produced the mismatched artifact silently, and the error didn't point at a fix.

The mismatch is fully known when the msg declaration is checked, so validate it there:

- MsgSelectorAnalysisPass now compares the parsed selector signature argument types against the fields' semantic ABI types and reports a spanned error on the offending field (9-0006), plus an arity check against the field count (9-0007). Variants with integer-literal selectors have no recoverable signature and are skipped.
- When the selector-declared type has a canonical stdlib representation, the diagnostic suggests it, e.g.:

      error[9-0006]: selector declares `string` but field
      `name_override` has ABI type `bytes`
        = `std::abi::DynString` decodes and encodes Solidity `string`

  The mapping (string -> DynString, bytes -> Bytes, address -> Address, non-native uintN/intN -> std::abi wrappers, native widths -> uN/iN, bytesN -> BytesN) lives in one function so future wrapper types extend it in one place.
- The semantic-type-to-ABI-type computation moves from crates/fe's abi.rs into the new shared hir module analysis/ty/abi_ty.rs; ABI emission delegates to it, so the declaration-time check and the emission-time backstop cannot drift. The backstop remains and its message now also carries the suggestion.
- Fix a latent panic ("cannot lower absent expression") when recovering the selector signature of a variant whose selector attribute failed to lower.

Return types are not checked: selector strings are strictly `name(type,...)` (the Solidity selector hash never includes returns), so there is no declared return type to compare `-> T` against. Event TOPIC0 and error SELECTOR are derived from the fields' own SOL_TYPEs and cannot mismatch by construction.


cburgdorf:
@codex review

sbillig:
from gpt5.6:

# Message Selector ABI Validation Stack-Overflows On Recursive ADTs

## Problem

Branch `improved_diagnostic` moves semantic-to-Solidity ABI type expansion into
HIR and invokes it while checking every recoverable `sol("...")` message
selector signature. Struct expansion recursively visits field types without a
nominal recursion guard, so invalid recursive source exhausts the process stack
before the existing recursive-type diagnostic can be returned.

## Evidence

At `54ab4fdb4`, this standalone source aborts `fe check` with exit code 134 and
`thread 'main' has overflowed its stack`:

```fe
use std::abi::sol

struct Node { next: Node }

msg ReviewMsg {
    #[selector = sol("set(uint256)")]
    Set { node: Node },
}
```

The exact master baseline at `745e56358` terminates normally and emits the
existing `recursive type definition` and unsatisfied ABI-trait diagnostics.
The regression is introduced by `check_variant_signature_types` calling
`semantic_ty_to_abi_desc`, whose struct branch recursively expands every field
through the same function.

## Impact

Invalid user input can abort the compiler, including editor and `fe check`
analysis paths that do not request JSON ABI emission. The shared ABI mapper can
also overflow when called directly by ABI generation.

## Suggested Fix

Reject nominally recursive ADTs before expanding their ABI components. Use the
existing `AdtDef::recursive_cycle` result rather than guarding only repeated
`TyId`s, because generically expanding recursion can produce a fresh
instantiation on every descent. Return an ABI-description error and let the
existing recursive-type diagnostic remain authoritative.

Cover direct, mutually recursive, and generically expanding ADTs nested under
message fields, tuples, fixed arrays, and dynamic arrays. Each case must
terminate under ordinary analysis and bytecode-only builds.

sbillig:
@cburgdorf I had 5.6 fix that issue. Rebasing now, then I'll merge.

sbillig (approved):

PR #1521, opened 2026-07-17
Fix trait method selection cycle for projected impl bounds (#1505)

### LINKED ISSUE #1505: ICE: dependency-graph cycle when resolving a method whose `impl` has an associated-type projection in its `where`-clause
### Summary

  Calling a method on a type whose `impl` block carries a `where`-clause bound on an **associated-type projection** (e.g. `where <Holder as Assoc>::Out: Tagged`) panics the compiler with a salsa dependency-graph
  cycle instead of either type-checking or producing a diagnostic.

  Notably, the **same bound on a free function type-checks fine** — the cycle is specific to the bound living on an `impl` that then gets consulted during method selection.

  ### Steps to reproduce

  ```fe
  trait Tagged {}

  struct Holder {}
  impl Tagged for Holder {}

  trait Assoc {
      type Out
  }

  impl Assoc for Holder {
      type Out = Holder
  }

  trait Consume {
      fn consume(self)
  }

  impl Consume for Holder
  where <Holder as Assoc>::Out: Tagged
  {
      fn consume(self) {}
  }

  fn caller() {
      let h = Holder {}
      h.consume()
  }
  ```

  Type-checking this file panics.

  ### Actual behavior

  ```
  thread '...' panicked at salsa-0.20.0/src/function/fetch.rs:138:25:
  dependency graph cycle when querying is_query_satisfiable(Id(14408)), set cycle_fn/cycle_initial to fixpoint iterate.
  Query stack:
  [
      check_func_body,
      is_query_satisfiable,
      is_query_satisfiable,
      impls_for_trait_in_ingots,
      impls_for_trait_def,
      lower_hir_ty,
      collect_generic_params,
      collect_methods,
      lower_hir_ty,
      kind_,
      resolve_query,
      lower_hir_ty,
      resolve_query,
      containing_ingot_config_,
      containing_ingot_config_,
  ]
  ```

  ### Expected behavior

  The bound `<Holder as Assoc>::Out: Tagged` is satisfiable (`Holder::Out = Holder`, and `Holder: Tagged`), so the call should type-check. At minimum it should not ICE.

  ### Notes on the trigger

  The cycle requires all three of:

  1. the bound is on an **`impl`** (the same projection bound on a free function does **not** ICE — it type-checks cleanly), and
  2. the bound is on an **associated-type projection** (`<T as Assoc>::Out` / `T::Out`), not a plain type, and
  3. a **method call** drives selection of that impl. Without the call site (`h.consume()`), the file type-checks without panicking.

  `impls_for_trait_def` → `lower_hir_ty` (lowering the impl's projection-bearing constraint) re-enters `is_query_satisfiable`, which closes the cycle.

cburgdorf:
Make the trait satisfiability query participate in Salsa fixpoint iteration when impl collection re-enters trait solving through an associated-type projection. Keep candidates ambiguous during the incomplete pass, then resolve them once impl collection converges.

Add positive and unsatisfied regression coverage for projection bounds on trait impls.

Fixes #1505

### DIFF
--- a/crates/hir/src/analysis/ty/trait_resolution/mod.rs
+++ b/crates/hir/src/analysis/ty/trait_resolution/mod.rs
@@ -244,7 +244,11 @@ pub(crate) fn normalize_trait_inst_preserving_validity<'db>(
     }
 }
 
-#[salsa::tracked(return_ref)]
+#[salsa::tracked(
+    return_ref,
+    cycle_fn=is_query_satisfiable_cycle_recover,
+    cycle_initial=is_query_satisfiable_cycle_initial
+)]
 fn is_query_satisfiable<'db>(
     db: &'db dyn HirAnalysisDb,
     origin_ingot: Ingot<'db>,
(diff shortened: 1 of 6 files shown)


cburgdorf:
@codex review

sbillig (approved):

PR #1524, opened 2026-07-17
std: add DynArray access and MemVec builder

cburgdorf:
Add ABI-decoded element access for static single-word DynArray elements, including canonical decoding and Solidity Panic(0x32) bounds failures.

Introduce a fixed-length mutable MemVec with zeroed and copy construction, indexed reads and writes, and conversion back to an ABI-encodable DynArray. Seal the supported element marker so dynamic and multi-word types fail at compile time.

Cover calldata reads, address canonicalization, bounds checks, copy isolation, returndata and event-payload encoding, and dynamic-element compile failures.


cburgdorf:
@codex review

cburgdorf:
Any opinions about this one? It's something that I came up with while working on porting various Solidity projects to Fe.

cburgdorf:
Merged based on our discussion from the meeting

PR #1531, opened 2026-07-28
Resolve trait obligations eagerly in `Callable::process_constraints`

### LINKED ISSUE #1525: Option::map can't infer result type U without annotation
## Overview 
**tldr: map<U, F: Fn<T, U>> can't infer U at the call site — needs let m: Option\<u256\> = n.map().** 

### Reproduction 

```rust
use core::option
use core::functional::Fn
struct Doubler {}
impl Fn<u256, u256> for Doubler { fn call(self, _ x: own u256) -> u256 { x * 2 } }

#[test]
fn map_no_annotation() {
    let n: Option<u256> = Option::Some(42)
    let m = n.map(Doubler {})        // error: type annotation is needed
    assert!(m.unwrap() == 84)
}
```
Output: 
```bash
error[8-0014]: type must be known
  └─ assert!(m.unwrap() == 84)
              ^^^^^^^^^^ type must be known here
error[8-0031]: type annotation is needed
  └─ consider giving `: Option<_>` here
```

KiChjang:
Fixes #1525.

Adopts the solution to eagerly resolve trait obligations during `enqueue_constraints`, which happens when type-checking calls and was unconditionally deferred till finalization.


KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

sbillig:
@KiChjang thanks for taking this on. Current state is definitely an improvement, but I think it's not quite right yet. 5.6sol dump below, of two concrete failure cases:

```rust
trait Pick<T> {}

trait HasOutput {
    type Output
}

struct A {}
struct B {}
struct Subject {}

impl HasOutput for A {
    type Output = u256
}

impl HasOutput for B {
    type Output = bool
}

impl Pick<A> for Subject {}

extern {
    fn todo() -> !
}

fn pick<T: HasOutput, U: Pick<T>>(_ value: U) -> T::Output {
    todo()
}

// ERROR CASE 1: a direct call reports a false type mismatch.
//
// `Subject: Pick<A>` determines `T = A`, and `A::Output = u256`, so this
// assignment is valid. The PR checks the unresolved `T::Output` against the
// expected `u256` before processing `U: Pick<T>`, then reports:
//
//     expected `u256`, but `_::Output` is given
//
// Processing the bound afterward is too late to retract that diagnostic.
fn direct_call_reports_false_error() {
    let value: u256 = pick(Subject {})
}

struct S {}

trait Foo<M> {
    fn foo<P: HasOutput, Q: Pick<P>>(
        self,
        _ marker: M,
        _ value: Q,
    ) -> P::Output
}

impl Foo<u8> for S {
    fn foo<P: HasOutput, Q: Pick<P>>(
        self,
        _ marker: u8,
        _ value: Q,
    ) -> P::Output {
        todo()
    }
}

impl Foo<bool> for S {
    fn foo<P: HasOutput, Q: Pick<P>>(
        self,
        _ marker: bool,
        _ value: Q,
    ) -> P::Output {
        todo()
    }
}

// ERROR CASE 2: a deferred method call is silently left unresolved.
//
// The `bool` marker selects `Foo<bool>`, while `Subject: Pick<A>` determines
// `P = A` and therefore `P::Output = u256`. This assignment is also valid.
//
// Unlike the direct-call case, `fe check` emits no diagnostic here. During
// deferred method selection, the PR compares the expected `u256` with the
// still-unresolved `P::Output` before processing `Q: Pick<P>`. It rejects the
// valid candidate and leaves the typed method-call expression without a
// recorded callable. A HIR-level assertion that `callable_expr(call).is_some()`
// fails even though source analysis reports no error.
fn deferred_method_is_silently_unresolved() {
    let s = S {}
    let value: u256 = s.foo(true, Subject {})
}
```

Both issues come from the same architectural problem: the checker makes observable decisions before the call’s local constraints have stabilized.

## 1. Direct calls

Current behavior:

```text
type arguments
→ obtain unresolved T::Output
→ compare it with expected u256 and emit an error
→ solve Subject: Pick<T>
→ discover T::Output = u256
→ attempt to recover
```

The ideal fix is to finalize the call before checking its result against the surrounding context:

```text
type callee and arguments
→ process call bounds to a fixed point
→ normalize the return type
→ coerce/unify with the expected type
→ record the final expression type
```

Concretely, `Callable::process_constraints` should run before the current `try_coerce...` and `unify_ty(..., expected)` block. Once it establishes `T = A`, normalization turns `T::Output` into `u256`, and the expected-type comparison succeeds normally.

The current post-solve retyping block should then disappear. It is inherently unable to recover from an error already added to the diagnostic list.

A robust version would treat these as one call-local inference phase:

1. Argument equations.
2. Trait obligations.
3. Return/expected-type equation.
4. Repeat while any step makes progress.
5. Only emit diagnostics after reaching the fixed point.

That also handles cases where the expected type makes a previously pending bound solvable.

## 2. Deferred method selection

Each candidate should have its own speculative call-local inference phase.

Current behavior:

```text
instantiate candidate
→ unify receiver and arguments
→ compare expected u256 with unresolved P::Output
→ reject candidate
→ never process Q: Pick<P>
```

Ideal behavior:

```text
instantiate candidate in a snapshot
→ unify receiver, explicit generics, and arguments
→ process the candidate’s bounds to a fixed point
→ normalize P::Output
→ compare the normalized result with the expected type
→ accept, reject, or keep the candidate pending
→ roll back the snapshot
```

For this example, processing `Q: Pick<P>` establishes `P = A`, so the return type becomes `u256` before viability is checked.

Candidate viability should ideally be tri-state rather than Boolean:

- `Viable`: all currently available constraints are compatible.
- `Incompatible`: a concrete contradiction exists.
- `Pending`: viability still depends on unresolved inference.

An unresolved projection must not be treated as a concrete mismatch. Otherwise valid candidates can be silently discarded.

When the unique candidate is committed, use the same ordering:

1. Unify receiver and arguments.
2. Process its bounds.
3. Normalize the return type.
4. Unify/retype the method expression.
5. Resolve effects/layout specialization.
6. Record the callable.

KiChjang:
I haven't touched the `is_viable` deferred method processing bit yet, but locally testing case 2 still passes for some reason.

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

KiChjang:
@codex review

sbillig (approved):
Looks good, thanks!

> locally testing case 2 still passes for some reason

I'll look into this and see if there's actually a problem here.

sbillig:
> > locally testing case 2 still passes for some reason
> 
> I'll look into this and see if there's actually a problem here.

Seems to be fine.

PR #1540, opened 2026-08-23
core/std: require exact-length returndata when decoding bool returns

### LINKED ISSUE #1539: [ABI/stdlib] External bool decoder accepts non-canonical returndata: any >=32-byte buffer with leading true-word decodes as true, tail ignored (v26.1.0, v26.2.0)
## Summary

The msg-decode helper emitted by fe/Sonatina for external calls returning `bool` accepts **any returndata buffer of at least 32 bytes whose first 32-byte word equals the canonical true word** — everything past byte 32 is silently ignored.

A contract author's mental model ("the external call returned exactly one ABI-encoded bool") does not match emitted behavior ("first word is 1, tail is irrelevant"). A counterparty returning `(true, <attacker-controlled data>)` tuples, padded buffers, or stale-memory tails decodes as a clean `true`.

This diverges from Solidity's stricter handling and widens the attack surface for cross-language contract composition.

## Reproduction

Setup: real fe BountyRegistry build deployed locally (self-admin); a Solidity mock challenge whose `isSolved()` returns attacker-chosen raw returndata via inline assembly `return(ptr, len)`; register → fund → lock → claim pipeline per shape.

### Accept/reject matrix (12 shapes)

| Shape | Length | Payload | Result |
|---|---|---|---|
| canonical true (control) | 32 | `...01` | ACCEPT — prize paid |
| true + garbage byte | 33 | `...01 ff` | **ACCEPT — prize paid** |
| true + zero tail | 33 | `...01 00` | **ACCEPT — prize paid** |
| true + mixed tail | 47 | `...01 + 7x00 + deadbeefdeadbeef` | **ACCEPT — prize paid** |
| true + long tail | 64 | `...01 + deadbeef...` | **ACCEPT — prize paid** |
| bare byte | 1 | `0x01` | reject — empty-revert panic |
| short prefix | 31 | first 31 B of true word | reject — empty-revert panic |
| first word = 0 (+tail) | >=32 | `...00 ...` | decoded as clean false -> caller reverts InvalidClaim(5) |
| word = 2 | 32 | `...02` | reject — empty-revert panic |

Decoder model fully determined:

```
if returndatasize < 32   -> panic-revert with EMPTY returndata
w = first 32 bytes
if w == 0x...01          -> true   (tail ignored entirely)
else if w == 0x...00     -> false  (tail ignored)
else                     -> panic-revert
```

Two distinct rejection fingerprints:
- `<32 B` or non-01/00 first word -> revert with empty returndata (panic inside the stdlib decode, before the caller's own error handling)
- first-word-zero -> clean `false` -> the caller's normal failure branch (its own error code)

## Versions affected

- **fe v26.1.0** — confirmed; this exact build is deployed on Ethereum mainnet (Bountiful bug-bounty registry)
- **fe v26.2.0** (commit 1fffb9e official release binary, macOS arm64) — confirmed byte-identical behavior across all shapes; not fixed

## Impact

Any fe contract that treats an external `bool` return as authoritative proof of counterparty state can be satisfied by a counterparty emitting padded or tuple-shaped returndata with a leading true-word. In the audited target this would let a registered challenge satisfy `claim()` without an ABI-canonical response. On mainnet it is currently unreachable only because the seven deployed challenges happen to emit exactly canonical words (verified pre- and post-solve on real builds) — a property of those programs, not a compiler guarantee. Any future encoder regression (e.g. an msize-derived buffer or multi-value return path) would silently pair with this decoder.

Found during an independent audit of the Bountiful bug-bounty contracts.

Full differential harness (dual-registry, shape-parameterized mock) available on request.

exstttt:
Fixes #1539

## Problem

The msg-decode helper for external calls returning `bool` accepted any returndata buffer of at least 32 bytes whose first word was the canonical true word, silently ignoring everything past byte 32. A counterparty returning `(true, <attacker-controlled data>)` tuples or padded buffers decoded as a clean `true`, diverging from Solidity semantics and widening the cross-language composition hazard described in the issue.

## Fix

- Add `remaining()` to the `AbiDecoder` trait and implement it for the Sol decoder (`input.len() - pos`).
- Enforce exactly 32 bytes in `bool::decode_payload` before accepting the word, any other length reverts via `decode_error`.

Fixed-width siblings (`u8`, `u64`, ...) have the same structural gap and can adopt the guard as follow-ups; this patch deliberately keeps scope to the reported bool case.

## Testing

- Full `cargo test -p fe` suite passes (454 tests across all targets, including contract-level ABI fixtures).
- The 12-shape accept/reject matrix from the issue is covered: canonical true accepted; 33/47/64-byte tails and dirty-tail variants now revert through the new guard.
- `cargo build -p fe` clean; `cargo fmt` state unchanged for touched files.

### DIFF
--- a/ingots/core/src/abi.fe
+++ b/ingots/core/src/abi.fe
@@ -1835,13 +1835,6 @@ impl<A: Abi> Encode<A> for DynString {
 impl<A: Abi> Decode<A> for bool {
     #[inline(always)]
     fn decode_payload<D: AbiDecoder<A>>(_ d: mut D) -> Self {
-        // A canonical ABI bool return is exactly 32 bytes. Reject buffers with
-        // trailing data instead of silently decoding the first word — a
-        // counterparty returning `(true, <attacker-controlled tail>)` must not
-        // decode as a clean `true`.
-        if d.remaining() != 32 {
-            A::decode_error()
-        }
         A::decode_bool_word(d.read_word())
     }
 
--- a/ingots/core/src/abi.fe
+++ b/ingots/core/src/abi.fe
@@ -156,6 +156,9 @@ pub trait AbiDecoder<A: Abi> {
     fn pos(self) -> u256
     fn set_pos(mut self, _ new_pos: u256)
 
+    /// Number of bytes remaining in the input after the current cursor.
+    fn remaining(self) -> u256
+
     fn read_u8(mut self) -> u8
     fn read_word(mut self) -> u256
     fn peek_word(self, _ at: u256) -> u256
@@ -1832,6 +1835,13 @@ impl<A: Abi> Encode<A> for DynString {
 impl<A: Abi> Decode<A> for bool {
     #[inline(always)]
     fn decode_payload<D: AbiDecoder<A>>(_ d: mut D) -> Self {
+        // A canonical ABI bool return is exactly 32 bytes. Reject buffers with
+        // trailing data instead of silently decoding the first word — a
+        // counterparty returning `(true, <attacker-controlled tail>)` must not
+        // decode as a clean `true`.
+        if d.remaining() != 32 {
+            A::decode_error()
+        }
         A::decode_bool_word(d.read_word())
     }
 
--- a/ingots/std/src/abi/sol.fe
+++ b/ingots/std/src/abi/sol.fe
@@ -344,6 +344,14 @@ impl<I> AbiDecoder<Sol> for SolDecoder<I>
         self.cur.pos = new_pos
     }
 
+    fn remaining(self) -> u256 {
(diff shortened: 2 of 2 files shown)


exstttt:
Both P1 points addressed in the new push (a7e3835):

1. **Wrong path**, correct: returns route through `decode_output_at` → `decode_from`, so `decode_payload` never saw the reported case. Canonicality now enforced at the boundary that knows the full payload: `decode_output_at` rejects a non-dynamic root whose input does not occupy exactly its head size. The 64-byte `(true, tail)` buffer now reverts.
2. **Composite breakage**, payload-level guard reverted; leaf decoders again accept whatever the frame check pre-approves.

Full suite green after the change (454 tests).

Issue #1538, opened 2026-08-23
[compiler] Sentinel-initialized scan + StorageMap default: out-of-bounds write at key 16 in compiled 15-puzzle (v26.1.0, v26.2.0)


janeeric111:
## Summary

Two behaviors combine to produce a silent **out-of-bounds storage write that the Fe source does not authorize**:

1. A sentinel-initialized empty-cell scan (`let mut j: usize = 16`) falls through when no element matches — execution continues with the out-of-range sentinel `16`.
2. `StorageMap.get()` returns the default value for ANY key. Using the sentinel as a key into an adjacency table whose default is `0`, combined with an adjacency predicate where "movable to cell 0" evaluates true for the all-zero encoding, accepts a move **with no empty cell on the board** and then writes `board.set(key: 16, val)` — outside the modeled 0..15 domain.

In a packed-u256 board representation this pollutes bits ≥64 of the word; with other layouts it could hit neighboring logical fields.

## Minimal reproduction

Fe source pattern (abridged from a real 15-puzzle contract):

```fe
fn find_empty(self) -> usize {
    let mut j: usize = 16
    let mut i: usize = 0
    while i < 16 {
        if self.board[i] == 0 { j = i }
        i += 1
    }
    j                       // can be 16 if no zero exists
}

pub fn move_field(mut self, index: u256) {
    // bounds check on `index` only; NOT on find_empty result
    let empty_idx = self.find_empty()
    let encoded = ADJACENCY[empty_idx]          // OOB read when empty_idx == 16 → default 0
    if !is_movable_to(encoded, index as u256) { revert(Error::NotMovable) }
    let val = self.board[index]
    self.board.set(index, 0)
    self.board.set(empty_idx, val)              // OOB WRITE at key 16
}
```

`is_movable_to(0, target)` decodes slot words of the default-0 encoding as "movable to target 0" → true for `index == 0`.

### Steps

1. Deploy the compiled GameBitboard (v26.2.0 build) with constructor `(validator: <addr>, board: 0x1FEDCBA987654321)` — note: **no zero nibble**.
2. Call selector `0x8bf02f32` (`moveField(uint256)`) with arg `uint256(0)`.
3. Call SUCCEEDS (source semantics say it must revert `NotMovable`).
4. Storage slot 1 changes from `0x…1FEDCBA987654321` to `0x…11fedcba987654320` — nibble 0 zeroed AND new nibble written at bits 64–67 ("cell 16" = shift 64).

## Versions affected

[...]

PR #1541, opened 2026-08-23
std: add checked accessors to StoragePackedArray

Nirbhay007:
get/set on StoragePackedArray are deliberately unbounded: indices map arithmetically onto packed slot words, so an index past the caller's modeled domain silently spills into the next storage region instead of failing. A 16-cell board backed by StoragePackedArray<4> happily accepts set(16, v) and writes a lane into bits 64-67 of the second slot word — the footgun reported in #1538.

Add get_checked/set_checked which take a const-generic modeled length and revert with Solidity's standard Panic(uint256) payload (code 0x32, IndexOutOfBounds) when the index is out of domain — the same signal the compiler emits for failed fixed-array bounds checks.

Note: #1538's headline claim (fixed-array ADJACENCY[16] reading a default instead of trapping) does not reproduce: bounds-checked array indexing with an empty revert has been in place well before v26.2.0 (see checked_index_value in codegen). The silent-write behavior in that report comes from StorageMap/StoragePackedArray default-and-spill semantics, i.e. a caller-side missing guard, which these accessors now let callers express.


cburgdorf:
Hey! Thanks for the PR, and sorry it took me so long to review it. I'm not entirely convinced by the proposed API: callers still have to supply `LEN` at each call site, so different callers can specify different bounds for the same array, or accidentally use the unchecked methods.

Would it make more sense to keep `StoragePackedArray` as it is and introduce a wrapper such as `BoundedStoragePackedArray<4, 16>`? That would make the length part of the type, allowing regular get/set calls and range-based operations such as search to enforce the same bound consistently.


What do you think?

PR #1547, opened 2026-09-11
Add equality comparisons for arrays and tuples

### LINKED ISSUE #528: Need a simple way to compare two arrays
### What is wrong?

We currently can't compare two arrays via the `==` operators:

>error: can't move value onto stack

This seems like pretty fundamental functionality to have.

### How can it be fixed

Need to investigate.

### LINKED ISSUE #748: Allow tuples to be compared with == operator
### What is wrong?

We currently can't easily compare tuples with the `==` operator. E.g. the following doesn't work.

```
if get_tuple() == (1, 2) {
...
}
```

Instead what we have to do is:

```
if get_tuple() == (1, 2) {
...
}
let (val1, val2): (u8, u8) = get_tuple()
if val1 == 1 and val2 == 2 {
}
```

### How can it be fixed

~~We can probably tackle that at MIR lowering~~

Make `PartialEq` and `Eq` traits possible and use them for this.

cburgdorf:
Adds == and != for fixed-size arrays and tuples of up to six elements, including (). Comparisons invoke each element’s Eq implementation and short-circuit at the first mismatch. No Copy bound is required;
  existing ownership semantics are preserved.

  The compiler changes ensure aggregate operator methods execute their bodies, correctly lower aggregate literals passed through implicit views, and handle empty-array bounds checks without attempting to
  materialize nonexistent elements.

  Includes regression tests and newsfragments for both issues.

  Validation

  - 128 EVM tests, 233 diagnostic tests, and 20 core/std tests passed before rebasing.
  - CLI and formatter roundtrip tests passed.
  - All 11 new EVM tests and both diagnostic fixtures passed again on the new branch base.

  Fixes #528
  Fixes #748

### DIFF
--- a/crates/fe/tests/fixtures/fe_test/aggregate_equality.fe
+++ b/crates/fe/tests/fixtures/fe_test/aggregate_equality.fe
@@ -1,0 +1,176 @@
+use core::ops::Eq
+
+fn equal<T: Eq>(_ a: T, _ b: T) -> bool {
+    a == b
+}
+
+fn pair() -> (u256, bool) {
+    (7, true)
+}
+
+#[test]
+fn tuple_arities_and_function_result() {
+    assert!(() == ())
+    assert!(!(() != ()))
+    let t1: (u256,) = (1,)
+    let t2: (u256, bool) = (7, true)
+    let t3: (u256, bool, u8) = (1, true, 3)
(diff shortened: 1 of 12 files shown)


sbillig (approved):

cburgdorf:
@codex review

PR #1549, opened 2026-09-13
Add experimental native executable builds

sbillig:
Adds an optional Cranelift backend for standalone Fe programs. With the `cranelift` feature enabled, `fe build --backend native program.fe` compiles `pub fn main() -> i32` and its reachable callees into a host executable. Native executables currently support x86-64 Linux and AArch64 macOS; `--emit ir` exposes the target-neutral Sonatina IR.

The implementation preserves unchecked division/remainder behavior, lowers native wrapping power and checked power with bounded exponentiation by squaring, and preserves the entry function's identity and exported `main` symbol when helpers share its name. Metadata rebuilds retain their default artifacts and use the shared backend/emit validation.

Pins Sonatina to merged revision `88ff0922763016ea1d43a2f59c68018d12f2ef14`, including the narrow EVM exponentiation fix in https://github.com/fe-lang/sonatina/pull/312. The prerequisite #1548 is merged.


sbillig:
@codex review

PR #1552, opened 2026-09-16
fix: ICE when a recv block names a top-level file module

basedpeasant:
## About

`fe build` panicks with an internal compiler error when a contract's `recv` block named a top-level file module (e.g. `recv test_mod { ... }`) instead of a `msg` module:

```
thread 'main' (1903860) panicked at crates/hir/src/analysis/ty/ty_check/contract.rs:913:13:
internal error: entered unreachable code
```

This fix makes the compiler report a clear diagnostic instead.

## Reproducing the error
I've made a very simple example case below in-case anyone wants to reproduce the error.  
```
// ./src/test_mod.fe
use std::abi::sol

pub msg test_msg {
    #[selector = sol("balanceOf(address)")]
    BalanceOf { account: Address } -> u256,

    #[selector = sol("allowance(address,address)")]
    Allowance { owner: Address, spender: Address } -> u256,

    #[selector = sol("totalSupply()")]
    TotalSupply {} -> u256,
}

// ./src/lib.fe
use self::test_mod::test_msg

pub contract Token {
    recv test_mod {

    }
}
```

Running `fe build` will result in the error:
```
thread 'main' (1903860) panicked at crates/hir/src/analysis/ty/ty_check/contract.rs:913:13:
internal error: entered unreachable code
```

## Fix

Match the module item directly in the pattern instead of using `if let` + `unreachable!()`:

```rust
Ok(PathRes::Mod(ScopeId::Item(ItemKind::Mod(mod_)))) => Some(mod_),
```

Any other module-like scope (`ItemKind::TopMod`) now falls through to the `Ok(other)` arm. The message now actually describes what a recv block name expects. Using the example from above, the error should now show the following instead of the ICE that was present before:

```
error[8-0090]: recv block expects a msg module
   ┌─ src/lib.fe:12:10
   │
12 │     recv test_mod {
   │          ^^^^^^^^ expected `msg` module, but file module `test_mod` is given
```


cburgdorf:
Thank you for your contribution @basedpeasant

PR #1556, opened 2026-09-18
Fix CTFE evaluation of symbolic array repeats

sbillig:
Generic constants such as `const VALUES: [u8; N] = [7; N]` were rejected while `N` was still symbolic. Preserve repeat elements and lengths as abstract constants, then materialize them after specialization. Deferred indexing keeps its bounds check, and empty arrays still evaluate their element expressions, including deferred faults.

Add HIR coverage for generic records, nested and empty arrays, trait-associated lengths/elements, runtime reification, and invalid expressions. Add an execution fixture covering specialized arrays and independent mutation of repeated nested arrays.


sbillig:
@codex review

sbillig:
@codex review

sbillig:
@codex review

PR #1557, opened 2026-09-20
Replace semantic borrow checking with structural ownership analysis

sbillig:
Replace semantic borrow checking with a shared ownership and capability analysis over verified normalized semantic IR. Moves, loans, aggregate and array contents, raw-memory effects, and interprocedural requirements now use the same operation contracts and structural storage model. Runtime lowering consumes that verified boundary and its explicit layout evidence.

This closes gaps where moving a scalar-only non-Copy pointee, forwarding effects through helpers, overwriting native references through raw memory, or reading through a temporary pointer could escape ownership checks.

- Track access footprints, ownership requirements, native validity, and caller-visible contents in structural summaries. Unknown raw addresses conservatively account for possible aliases of moved storage. Byte writes cannot establish valid native references or definite typed initialization; typed stores can restore initialized contents.
- Defer generic calls that depend on unresolved implementations through explicit pending validation, then resolve and check them at concrete specialization. Executable calls that remain unresolved are rejected. External-call effects cover persistent and transient state, including effects from callbacks.
- Share fresh-allocation birth events between provenance and ownership availability, including allocations returned by calls and dynamic string literals. Loop generations distinguish separate allocation occurrences while preserving the moved state of earlier objects.
- Preserve native references' addresses, layouts, and aliasing when they pass through slots, fields, and returns. Native references use a distinct runtime representation while ordinary implicit views retain their static representation.
- Fix frontend place typing and move marking for temporary-pointer dereferences and projected fields and array elements. Fix native-reference slot reads, assignments, and compound assignments, contextual Copy reads from reference-returning calls, and diagnostics for mutable field borrows through immutable owned receivers.
- Reserve StorageMap's complete key-plus-salt hashing buffer before invoking custom key encoders, preventing allocations during encoding from overlapping the buffer. **Custom `StorageKey` implementations must now provide `encoded_len`.**
- Make internal lowering errors fail codegen snapshots instead of being accepted as snapshot output.

Raw operations retain a valid-whole-range precondition; this analysis does not prove arbitrary allocation bounds. Boolean-selected allocating factories can still lose path correlation, and recursive functions returning fresh allocations can still fail summary convergence. These conservative limitations remain tracked separately.

The design and invariants are described in [the borrow checker architecture](https://github.com/argotorg/fe/blob/semantic-borrow-restart/docs/architecture/borrow-checker.md). Sonatina is pinned to merged main `54492147373b8bf45f381bcef39477f0fb194648`, including the fixes for [aggregate scalarization ordering](https://github.com/fe-lang/sonatina/pull/318) and [callee heap publication](https://github.com/fe-lang/sonatina/pull/319).

The separate symbolic array-repeat CTFE fix in #1556 depends on this PR.


sbillig:
@codex review

PR #1560, opened 2026-09-21
Optimize packed string ABI encoding

sbillig:
Constant ERC20 `name()` and `symbol()` values retain generic packed-string length loops across ABI encoding boundaries. Compute packed-string length with five fixed binary-search steps, size its payload directly from whether the masked word is empty, and inline the string encoder and single-root allocation wrapper so constant metadata folds during optimization.

Add independent ABI checks across effective lengths 0–31, capacities 0/4/31, dirty high bits, and interior/trailing zeros. Add constant-getter and ERC20 size regressions, with ERC20 getters, transfers, allowances, self-transfers, and revert-payload checks. Refresh the affected IR snapshots.

Measured with unchanged ERC20 source, Fe O2, against master `9c9840ce4`:

| Metric | Master | This PR |
| --- | ---: | ---: |
| Initcode (without arguments) | 1,675 bytes | 1,343 bytes |
| Runtime bytecode | 1,551 bytes | 1,219 bytes |
| Internal CREATE gas | 388,913 | 322,429 |
| `name()` internal-call gas | 3,861 | 857 |
| `symbol()` internal-call gas | 2,549 | 985 |


sbillig:
@codex review

cburgdorf (approved):

PR #1561, opened 2026-09-21
Fix verification of payable value-returning handlers

cburgdorf:
ABI wrappers for value-returning recv arms terminate through a call to ContractHost::return_value. Accept that terminal call in the package verifier; the runtime body verifier already checks its non-returning callee and argument signature. Payable arms have no nonpayable revert guard, so they previously failed verification with InvalidReturnClass.

Cover scalar, tuple and dynamic returns with ETH at every optimization level, and verify that a payable deposit retains its uint256 return in the ABI.


sbillig:
@codex review

PR #1563, opened 2026-09-22
Decode string literal escapes before semantic lowering

sbillig:
String literals currently reach HIR with escape backslashes intact. This gives escaped strings incorrect byte lengths and makes native JSON output invalid.

Decode quote, backslash, newline, carriage-return and tab escapes before interning semantic strings. Preserve original syntax spelling, diagnose unknown escapes at their source spans, and carry invalid literals through error recovery without treating their raw spelling as a valid string. Literal typing, constant evaluation, EVM and native emission now share the decoded UTF-8 value.

Depends on #1568; stacked on `fevm-pr-native-modular`.


sbillig:
@codex review

PR #1571, opened 2026-09-22
Lower unchecked signed negation to EVM-supported subtraction

cburgdorf:
Unchecked signed negation emitted a Sonatina `Neg` instruction that the EVM backend does not support. Lower it to width-preserving `0 - value` instead, retaining wrapping behavior for minimum signed values.

Validation: runtime regression tests cover i8 and i256 values, including their signed minima, at O0, O1, O2 and Os.


sbillig (approved):
This might be better solved as a sonatina improvement, but I'll allow it.

PR #1575, opened 2026-09-22
Preserve ordinary mutable argument memory contracts in borrow summaries

cburgdorf:
Composed ABI decoders could report a borrow conflict when calldata reads were treated as potentially aliasing an ordinary mutable decoder argument. Preserve the argument's memory transport contract in input provenance and summary verification; receivers and effects retain their provider address spaces.

Validation: all 294 semantic borrow tests pass. Three runtime regressions pass at O1 and O2, covering two-address round trips and noncanonical words in either position.


sbillig:
I need to look at this in the context of my borrowcheck follow up branch; there might be a more general fix.

